Issue #6, opened 2019-02-14
uninitialised field in const object


stefantalpalaru:
The Nim VM fails to zero an object field in a `const` assignment involving a proc call.

Nim code:
```nim
proc initNibbleRange*(bytes: ByteRange): NibblesRange =
  result.bytes = bytes
  result.ibegin = 0
  result.iend = bytes.len * 2> Does this sound familiar to anyone? The Nim VM fails to zero an object field in a `const` assignment involving a proc call.

Nim code:
```nim
proc initNibbleRange*(bytes: ByteRange): NibblesRange =
  result.bytes = bytes
  result.ibegin = 0
  result.iend = bytes.len * 2

const zeroNibblesRange* = initNibbleRange(zeroBytesRange)
```
Now this constant is imported and used in another file. This is how the generated C looks for the latter:
```C
NIM_CONST tyObject_NibblesRange_NTijGP6lMbjliuMIEehBLw TM_9aIpn29c4nFUFIFQM9b2UMu5w_4 = {{NIM_NIL, NIM_NIL, ((NI) 0)}, ((NI) 0), ((NI) IL64(279856062208720))};
```
We suddenly have uninitialised memory in the `iend` field instead of 0.

This is a hard to replicate bug. I tried copying the relevant Nim code in the same file or in separate files (in case it's an import issue) to make a stand-alone test case, without any luck.

The best test case I can come up with needs nim-eth:
```nim
import eth/trie/[trie_defs,nibbles]

echo zeroBytesRange.len
echo zeroNibblesRange.len
let nr = initNibbleRange(ByteRange())
echo nr.len
```
The obvious workaround is to replace the `const` with a `let`, but that won't solve the underlying VM bug.

const zeroNibblesRange* = initNibbleRange(zeroBytesRange)
```
Now this constant is imported and used in another file. This is how the generated C looks for the latter:
```C
NIM_CONST tyObject_NibblesRange_NTijGP6lMbjliuMIEehBLw TM_9aIpn29c4nFUFIFQM9b2UMu5w_4 = {{NIM_NIL, NIM_NIL, ((NI) 0)}, ((NI) 0), ((NI) IL64(279856062208720))};
```
We suddenly have uninitialised memory in the `iend` field instead of 0.

This is a hard to replicate bug. I tried copying the relevant Nim code in the same file or in separate files (in case it's an import issue) to make a stand-alone test case, without any luck.

The best test case I can come up with needs nim-eth:
```nim
[...]

PR #7, opened 2019-02-14
convert zeroBytesRange and zeroNibblesRange into threadvars

stefantalpalaru:
It's ugly, but it fixes https://github.com/status-im/nim-eth/issues/6


zah:
Why didn't you use a thread-local let variable? The required patch will be smaller and once the upstream issue is fixed, it will be easy to get back to `const`.

stefantalpalaru:
Converted to threadvars.

PR #8, opened 2019-02-14
clean up the test suite

stefantalpalaru:
by removing test wrappers that are no longer used and renaming a helper file that had a name starting with the "test_" prefix


zah:
Is there still a way to run all Trie tests separately?

stefantalpalaru:
There is now, for all included modules, with custom test tasks: `nimble test_trie`.

PR #10, opened 2019-02-18
more robust dbTransaction dispose

jangko:


jangko:
this PR is needed for upcoming nimbus `defer: computation.dispose()` but I have no idea why appveyor `test_flow_control` failed

eth/trie/db.nim
```diff
@@ -34,11 +34,15 @@ type
     containsProc: ContainsProc
     mostInnerTransaction: DbTransaction
 
+  TransactionFlags = enum
```
zah: Since you are modelling a state machine here, it would be better if you have a single `TransactionState` enum with values such as `Pending`, `Committed`, `RolledBack`. This will make all of the checks below slightly simpler.

cheatfate: Also by default `TransactionFlags` value without initialization will have `Commited` value, this can cause problems in future, so its better to add `None` or `Pending` to this enum.

eth/trie/db.nim
```diff
@@ -137,24 +141,32 @@ proc rollback*(t: DbTransaction) =
   # Transactions should be handled in a strictly nested fashion.
   # Any child transaction must be committed or rolled-back before
   # its parent transactions:
-  doAssert t.db.mostInnerTransaction == t and not t.committed
+  doAssert t.db.mostInnerTransaction == t and
+    Committed notin t.flags and
+    RolledBack notin t.flags
```
cheatfate: This can be replaced with
```doAssert (t.db.mostInnerTransaction == t) and (t.flags * {Commited, RolledBack} == {})```

File: tests/trie/testutils.nim [hunk 3]
```diff
@@ -27,6 +27,11 @@ proc randString*(len: int): string =
   for i in 0..<len:
     result[i] = rand(255).char
 
+proc randBytes*(len: int): Bytes =
```
cheatfate: `nimcrypto` is already dependency for `nim-eth`, so it would be better to use `nimcrypto.sysrand`.

eth/trie/db.nim
```diff
@@ -137,24 +141,32 @@ proc rollback*(t: DbTransaction) =
...
   t.db.mostInnerTransaction = t.parentTransaction
   t.modifications.commit(t.db)
-  t.committed = true
+  t.flags.incl Committed
 
 proc dispose*(t: DbTransaction) {.inline.} =
-  if not t.committed:
+  if Committed notin t.flags and
+     RolledBack notin t.flags:
```
cheatfate: Can be replaced with
```if t.flags * {Commited, RolledBack} == {}:```

File: eth/trie/db.nim [hunk 5]
```diff
@@ -137,24 +141,32 @@ proc rollback*(t: DbTransaction) =
...
     t.rollback()
 
 proc safeDispose*(t: DbTransaction) {.inline.} =
-  if t != nil and not t.committed:
+  if t != nil and
+     Committed notin t.flags and
+     RolledBack notin t.flags:
```
cheatfate: You must not use explicit comparison with `nil`, please use `isNil(value)`
Can be replaced with
```if (not isNil(t)) and (t.flags * {Commited, RolledBack} == {}):```

File: eth/trie/db.nim [hunk 6]
```diff
@@ -137,24 +141,32 @@ proc rollback*(t: DbTransaction) =
...
+  t.flags.incl RolledBack
 
 proc commit*(t: DbTransaction) =
   # Transactions should be handled in a strictly nested fashion.
   # Any child transaction must be committed or rolled-back before
   # its parent transactions:
-  doAssert t.db.mostInnerTransaction == t and not t.committed
+  doAssert t.db.mostInnerTransaction == t and
+    Committed notin t.flags and
+    RolledBack notin t.flags
```
cheatfate: This can be replaced with
```doAssert (t.db.mostInnerTransaction == t) and (t.flags * {Commited, RolledBack} == {})```

jangko:
All comments have been addressed.

tests/trie/testutils.nim
zah: It's probably better to keep the tests deterministic though, so a pseudo-random generator is better here.

cheatfate: Could not agree with deterministic generator, just because if the problem will happens at least once you will need to review whole the algorithm, but deterministic one can hide the problem because algorithm will work all the time with same inputs.

PR #11, opened 2019-02-19
don't swallow random exceptions

arnetheduck:


tersec:
I'm in favor of this PR.

arnetheduck:
this is a follow-up on https://github.com/status-im/nim-eth-p2p/pull/56

funny enough, looks like tests fail because of a previously swallowed exception ;)

arnetheduck:
randomly, on some platforms, at that!

yglukhov:
I've seen this test failing before. cc @kdeme

kdeme:
@yglukhov I thought the failure you saw before was with the mocked version of the test.

I think this failure at test_shh_connect is related to this PR. That is, the issue is probably always/often there, but the error only shows up now. I will investigate.

yglukhov:
I could be wrong, but here's a recent build from nim-eth-p2p, master branch. The failure is not exactly the same, but feels somewhat similar to me. https://travis-ci.org/status-im/nim-eth-p2p/jobs/485331507

Issue #3, opened 2019-02-06
[RLPx] Test is passing even though there are errors


mratsim:
Is that normal that the following is passing even though there are errors?

https://travis-ci.org/status-im/nim-eth/jobs/489580420#L3422-L3431

cc @zah 

```bash
DBG 2019-02-06 15:54:11+00:00 Received neighbours                        topics="discovery" thread=15118 neighbours="@[Node[127.0.0.1:30303], Node[127.0.0.1:30304], Node[127.0.0.1:30305]]" remote=Node[127.0.0.1:30301]
DBG 2019-02-06 15:54:11+00:00 onPeerConnected Whisper                    thread=15118
DBG 2019-02-06 15:54:11+00:00 onPeerConnected Whisper                    thread=15118
DBG 2019-02-06 15:54:11+00:00 Whisper peer                               thread=15118 peer=Node[127.0.0.1:51484] whisperVersion=6
DBG 2019-02-06 15:54:11+00:00 Sending envelopes                          thread=15118 amount=0
DBG 2019-02-06 15:54:11+00:00 Whisper peer initialized                   thread=15118
DBG 2019-02-06 15:54:11+00:00 Whisper peer                               thread=15118 peer=Node[127.0.0.1:30304] whisperVersion=6
DBG 2019-02-06 15:54:11+00:00 Whisper peer initialized                   thread=15118
DBG 2019-02-06 15:54:11+00:00 Not enough nodes                           topics="discovery" thread=15118 present=3 requested=10
DBG 2019-02-06 15:54:11+00:00 Not enough nodes                           topics="discovery" thread=15118 present=3 requested=10
DBG 2019-02-06 15:54:11+00:00 Received neighbours                        topics="discovery" thread=15118 neighbours="@[Node[127.0.0.1:30303], Node[127.0.0.1:30301], Node[127.0.0.1:30305]]" remote=Node[127.0.0.1:30304]
DBG 2019-02-06 15:54:11+00:00 Received neighbours                        topics="discovery" thread=15118 neighbours="@[Node[127.0.0.1:30303], Node[127.0.0.1:30301], Node[127.0.0.1:30304]]" remote=Node[127.0.0.1:30305]
DBG 2019-02-06 15:54:11+00:00 Received neighbours                        topics="discovery" thread=15118 neighbours="@[Node[127.0.0.1:30303], Node[127.0.0.1:30301], Node[127.0.0.1:30304]]" remote=Node[127.0.0.1:30305]
[...]

PR #16, opened 2019-03-07
Fix unhandled exceptions #3

kdeme:
Fixes the two issues mentioned in #3 .

The reason as to why the message `dispatchMessages failed` would not always occur is because it happens in the last test and sometimes that code simply did not run before the test finished. Adding an `asyncSleep` of several ms to the test will make it happen always.

I'm also thinking of reworking the `rlpxConnect` and `rlpxAccept` to make them both behave the same in terms of returning a "failing" Future. But I didn't want to do those changes in this PR as impact will be a little bit bigger.

### DIFF
--- a/eth/p2p/rlpx.nim
+++ b/eth/p2p/rlpx.nim
@@ -505,7 +505,16 @@ proc nextMsg*(peer: Peer, MsgType: type): Future[MsgType] =
 
 proc dispatchMessages*(peer: Peer) {.async.} =
   while true:
-    var (msgId, msgData) = await peer.recvMsg()
+    var msgId: int
+    var msgData: Rlp
+    try:
+      (msgId, msgData) = await peer.recvMsg()
+    except TransportIncompleteError:
+      trace "Connection dropped in dispatchMessages"
+      # This can happen during the rlpx connection setup or at any point after.
+      # Because this code does not know, a disconnect needs to be done.
+      asyncDiscard peer.disconnect(ClientQuitting)
+      return
 
     if msgId == 1: # p2p.disconnect
       await peer.transport.closeWait()
@@ -1337,7 +1346,7 @@ proc rlpxConnect*(node: EthereumNode, remote: Node): Future[Peer] {.async.} =
   except TransportIncompleteError:
     trace "Connection dropped in rlpxConnect", remote
   except UselessPeerError:
-    trace "Useless peer ", peer = remote
+    trace "Disconnecting useless peer", peer = remote
   except RlpTypeMismatch:
     # Some peers report capabilities with names longer than 3 chars. We ignore
     # those for now. Maybe we should allow this though.
@@ -1345,9 +1354,11 @@ proc rlpxConnect*(node: EthereumNode, remote: Node): Future[Peer] {.async.} =
   except TransportOsError:
     trace "TransportOsError", err = getCurrentExceptionMsg()
   except:
-    debug "Exception in rlpxConnect", remote,
+    error "Exception in rlpxConnect", remote,
           exc = getCurrentException().name,
           err = getCurrentExceptionMsg()
+    result = nil
+    raise
 
   if not ok:
     if not isNil(result.transport):
@@ -1365,6 +1376,7 @@ proc rlpxAccept*(node: EthereumNode,
   var handshake = newHandshake({Responder})
   handshake.host = node.keys
 
+  var ok = false
   try:
     let initialSize = handshake.expectedLength
     var authMsg = newSeqOfCap[byte](1024)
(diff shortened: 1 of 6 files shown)


eth/p2p/rlpx.nim
```diff
@@ -505,7 +505,16 @@ proc nextMsg*(peer: Peer, MsgType: type): Future[MsgType] =
...
-    var (msgId, msgData) = await peer.recvMsg()
+    var msgId: int
+    var msgData: Rlp
+    try:
+      (msgId, msgData) = await peer.recvMsg()
+    except TransportIncompleteError:
+      trace "Connection dropped in dispatchMessages"
+      # This can happen during the rlpx connection setup or at any point after.
+      # Because this code does not know, a disconnect needs to be done.
+      asyncCheck peer.disconnect(ClientQuitting)
```
yglukhov: If `disconnect` raises, we're gonna fail in the middle of nowhere. I guess the `disconnect` error should be ignored.

kdeme: Do you mean that the `peer.disconnect` should be in a `try` clause?

Currently it is also an issue here then: https://github.com/status-im/nim-eth/blob/master/eth/p2p/rlpx.nim#L1237 . And perhaps at other disconnect locations...

How about we make sure disconnect does not raise? I think errors currently can only come from the disconnect handlers?

kdeme: If we don't want to alter the "raise" behaviour of `disconnect()` proc, then I would 
add a proc that does the `disconnect` and a `debug` log of possible exceptions. 
Only a discard would leave errors completely unnoticed.

Question about the current `disconnect()` proc though: if any of the disconnect handlers raises, then `removePeer` will not be executed.
This looks like problem to me as the peer will remain in the `peerPool.connectedNodes`, no?

zah: I think each of the user supplied disconnect handlers has to be guarded in a `try: ... except CatchableError: ...` block. If an error is detected, a warning can be printed and we try to execute the rest of the handlers. In a finally block, we clean up the resources associated with the peer.

arnetheduck: can we enforce this with a `{.raises.}` clause of some sort?

arnetheduck:
in general, one of the reasons why it's so hard to write tests is that networking tests like these are fundamentally non-deterministic. 

unfortunately, the architecture of rlpx does not take testing into account making it far more difficult to test than it would have to be.. thus, while this branch can be merged, we also need to make architectural strides to ensure that either rlpx gains testing capabilities or significantly more code is moved out of it - it's completely unreasonable to require a running client to test that messages are set up and encoded appropriately, and that protocols go through the various stages correctly, including handling unexpected scenarios.

kdeme:
Turns out that thanks to [this](https://github.com/status-im/nim-chronos/commit/ea3fb9629a90a5ccda9ab2e74f17dc8ce86ef39c) change all handlers will still complete even when one fails. We need the `except` though still to catch it and not crash. 

Other option would be to change `return all(futures)` in `callDisconnectHandlers` to:
```
var f = all(futures)
yield f
if f.failed:
  warn "Error while running disconnectHandler", peer,
    err = f.readError.msg
```
I can't however seem to make the raises pragma work on this (async) proc, so I opted for the simple `except`

eth/p2p/rlpx.nim
```diff
@@ -505,7 +505,16 @@ proc nextMsg*(peer: Peer, MsgType: type): Future[MsgType] =
 
 proc dispatchMessages*(peer: Peer) {.async.} =
   while true:
-    var (msgId, msgData) = await peer.recvMsg()
+    var msgId: int
+    var msgData: Rlp
+    try:
+      (msgId, msgData) = await peer.recvMsg()
+    except TransportIncompleteError:
```
zah: This seems slightly redundant, because the code that uses `dispatchMessages` is already handling the errors from it.

I see it a slight problem, because it changes the contract/interface that `dispatchMessages` provides. Before the change, the function was exiting without an error only in the case where the connection was properly terminated with a `Quit` message. All the other expected ways the connection can be interrupted were communicated by raising a `CatchableError`.

Now, after the change, there are situations where the function exists without an error even when one actually occurred.

tests/p2p/test_failing_handler.nim
```diff
@@ -0,0 +1,93 @@
...
+  let port = Port(port)
+  result = Address(udpPort: port, tcpPort: port, ip: parseIpAddress("127.0.0.1"))
+
+proc startDiscoveryNode(privKey: PrivateKey, address: Address,
+                        bootnodes: seq[ENode]): Future[DiscoveryProtocol] {.async.} =
+  result = newDiscoveryProtocol(privKey, address, bootnodes)
+  result.open()
+  await result.bootstrap()
+
+proc setupBootNode(): Future[ENode] {.async.} =
```
zah: Are these taken from the Whisper tests? We can create a helper module for these common setup steps.

kdeme: Yes, very similar from Whisper tests, I'll put this in a module.

eth/p2p/rlpx.nim
```diff
@@ -1161,6 +1170,9 @@ proc disconnect*(peer: Peer, reason: DisconnectionReason, notifyOtherPeer = fals
     try:
       if not peer.dispatcher.isNil:
         await callDisconnectHandlers(peer, reason)
```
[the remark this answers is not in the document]
kdeme: OK, did not know of that proc, will adjust it.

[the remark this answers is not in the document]
kdeme: is it actually OK to use `traceAsyncErrors` considering that now an `await` is done?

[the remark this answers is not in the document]
kdeme: Created an `traceAwaitErrors` to keep the same waiting behaviour

eth/p2p/rlpx.nim
kdeme: As `RlpError` was also violating this contract/interface, and as there is also an unneeded "reraise" which would also create a double log message (one in `dispatchMessages` and one in its callback), I've decided to handle all "known" fatal messages inside `dispatchMessages`. See more information in code comment.

kdeme:
As the code touches RLPx, I've also tested this last commit with nim-beacon-chain sims and with testnet0. Did not notice anything out of the ordinary. 
Actually, a (harmless) `Data incomplete!` exception message that was sometimes visible on a disconnect has gotten cleaned up.

PR #18, opened 2019-03-12
AbstractChainDb -> AbstractChainDB

stefantalpalaru:
Trivial letter case change for the sake of consistency (and ctags).


arnetheduck:
Unless you're after the travis build, it's generally ok to go straight to master with renames, typos etc (of course, it makes sense to communicate with the team through some other means instead)

PR #20, opened 2019-03-19
bind to any address for listening

arnetheduck:


cheatfate:
@arnetheduck i think the proper place to change is
https://github.com/status-im/nim-beacon-chain/blob/master/beacon_chain/eth2_network.nim#L39-L41

arnetheduck:
Not quite - at least not in a nat setup - there are two ip's involved then:

* your local computer IP - typically something like 192.168.x.x, as given by the router
* the router / external IP

The enode is meant for anyone to connect to, while your local interface needs to be bound to the right interface that goes to the router. Since it's tricky to bind to the right interface, most p2p apps bind to anything, then have options to limit that, for more unusual, multihomed setups. 

The enode needs to reflect the IP that people should be connecting through. Technically, it should contain all the IP's of the machine, or possibly send different enodes to local and remote machines. We're far from there.

See geth:
```
[arnetheduck@tempus nim-beacon-chain]$ geth --nat=extip:$(curl -s ifconfig.me)
INFO [03-19|06:50:25.931] Started P2P networking                   self=enode://fe7fad4977abb624011e167ea1a9c1996273493874a660d97973b4e245dbf47f16855aac7717419f781bfdf842dad9daab88d6c23305dae96a3f05439aadb188@201.207.254.214:30303
...
[arnetheduck@tempus nim-beacon-chain]$ netstat -lpn
...
tcp6       0      0 :::30303                :::*                    LISTEN      23860/geth          
udp6       0      0 :::30303                :::*                                23860/geth          
```
(yes, the tcp6 all addresses work for tcp4 as well)

PR #23, opened 2019-03-23
Add powTarget and other changes for Whisper RPC implementation

kdeme:
PR with some changes for upcoming Whisper RPC PR.

Biggest change is adding the option of `powTarget` so that you don't always need to wait for `powTime`. This is an option in the `shh_post` RPC call.

Annoyingly, this does violate the Whisper EIP in its current implementation (so does geth by the way, I understand now why). I think in general it would be better if Whisper is not RLP depended. However the PoW hash is still RLP depended so this does not really resolve that...

Nice side effect is that testing time of test_shh_connect.nim dropped ~30 seconds.

By the way, I would perhaps move `generateRandomID ` to somewhere more "common", e.g.  `nim-eth/eth/common/utils.nim`?

### DIFF
--- a/eth/p2p/rlpx_protocols/whisper_protocol.nim
+++ b/eth/p2p/rlpx_protocols/whisper_protocol.nim
@@ -23,6 +23,7 @@ const
   defaultQueueCapacity = 256
   defaultFilterQueueCapacity = 64
   whisperVersion* = 6
+  whisperVersionStr* = "6.0"
   defaultMinPow* = 0.001'f64
   defaultMaxMsgSize* = 1024'u32 * 1024'u32 # * 10 # should be no higher than max RLPx size
   messageInterval* = 300 ## Interval at which messages are send to peers, in ms
@@ -79,6 +80,7 @@ type
     topic*: Topic
     pow*: float64
     hash*: Hash
+    dst*: Option[PublicKey]
 
   Queue* = object
     ## Bounded message repository
@@ -103,12 +105,12 @@ type
   FilterMsgHandler* = proc(msg: ReceivedMessage) {.gcsafe, closure.}
 
   Filter* = object
-    src: Option[PublicKey]
-    privateKey: Option[PrivateKey]
-    symKey: Option[SymKey]
-    topics: seq[Topic]
-    powReq: float64
-    allowP2P: bool
+    src*: Option[PublicKey]
+    privateKey*: Option[PrivateKey]
+    symKey*: Option[SymKey]
+    topics*: seq[Topic]
+    powReq*: float64
+    allowP2P*: bool
 
     bloom: Bloom # cached bloom filter of all topics of filter
     handler: FilterMsgHandler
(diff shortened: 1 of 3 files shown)


kdeme:
hmm, more than the usual failure on Mac, probably fixed by adding more margin to the `sleepAsync(messageInterval)` .

eth/p2p/rlpx_protocols/whisper_protocol.nim
```diff
@@ -868,14 +877,43 @@ proc sendP2PMessage*(node: EthereumNode, peerId: NodeId, env: Envelope): bool =
...
+proc sealEnvelope*(msg: var Message, powTime: float, powTarget: float): bool =
+  let size = msg.env.len
+  if powTarget > 0:
+    let x = powTarget * size.float * msg.env.ttl.float
+    var bestBitTarget: int
+    if x <= 1: # log() would return negative numbers or 0
+      bestBitTarget = 1
+    else:
+      bestBitTarget = ceil(log(x, 2)).int
+    (msg.env.nonce, msg.hash) = msg.env.minePow(powTime, bestBitTarget)
```
zah: How much time does the proof-of-work take in practice with typical settings? Should we do something to offload it from the main thread? (it will block all network communications at the moment)

zah: It's also possible to make it run on the main thread in small incremental steps that `yield` control back to the I/O loop after a fixed interval of time.

kdeme: I thought I was using the same default minimum PoW as in geth but apparently not.
Anyway, setting it to 0.2 (default of geth) gives me numbers of around 20-30 ms up to 80 ms with TTL=30 and message of 256 bytes (very quick test).

I do like the `yield` idea. I think I'll have a go on this later.

kdeme:
Changed the default minimum PoW to the one that geth has.
Added some marge to the `messageInterval` sleep to be sure that the messages have arrived at the receiving peer before checking. (this gave many errors on test_shh_connect on macOS).

eth/p2p/rlpx_protocols/whisper_protocol.nim
kdeme: Would the use of `poll()` be accepted in this scenario? I think it is the easiest and cleanest (no need to make proc async). However, perhaps we don't want to promote the usage of `poll()` (and even remove it?) due to issues it can create when used in async proc (recursive poll)?

zah: @cheatfate has opposed recurrent usage of `poll` or `waitFor` in the past. I'll let him comment on this and I'll also mention that we could benefit from an `AsyncChannel` facility provided by Chronos is situations like this where the work can be sent to a background thread.

Is the recently added support for pipes enough to implement `AsyncChannel` quickly?

kdeme: OK, to be clear, the `poll` would not be in an async call and thus not re-entrant (at least that is how I understand it). It is just that it is a very direct usage of the call and perhaps now what we want.

kdeme: Actually, it's a bad idea. Giving it some thought, you don't know how the API will be used, and thus you don't know if it will ever end up in a `async` call (unless you can somehow enforce not to allow this?).

In fact, it will already do in the Whisper RPC implementation (and a bit hidden due to the macro, thus extra dangerous). While it did not create any issue while I was testing this, it probably will bite back at the time you least expect it.

Can we merge this PR as is? I will create an issue to improve this.

PR #25, opened 2019-03-25
use Moment for tracking timeouts

arnetheduck:


eth/p2p/rlpx.nim
```diff
@@ -305,7 +305,9 @@ proc registerRequest*(peer: Peer,
   inc peer.lastReqId
   result = peer.lastReqId
 
-  let timeoutAt = fastEpochTime() + uint64(timeout)
+  # TODO turn timeout into Duration, but that requires messing with obnoxious
+  #      macro code
+  let timeoutAt = Moment.fromNow(milliseconds(timeout))
```
zah: So, what's the TODO here? How is `Duration` different than `timeoutAt` semantically? I don't think you need to change anything in the macro code. All timing-related code is located here. You may need to change the definition of `OutstandingRequest` which is located in `p2p_types.nim`.

arnetheduck: timeout should be of the `Duration` type, as should default timeout etc, so that refactoring etc doesn't accidentally confuse time unit.

arnetheduck: `timeoutAt` is a moment in time - a concrete time point. `timeout` is the interval that you add to the current time, to establish that point in time.

zah:
I've finished the TODO items and rebased the branch.

PR #27, opened 2019-03-25
*.nimble: remove imports

stefantalpalaru:
- also speed up the tests with release builds

### DIFF
--- a/eth.nimble
+++ b/eth.nimble
@@ -15,34 +15,86 @@ requires "nim >= 0.19.0",
          "chronicles",
          "std_shims"
 
-import strutils
-import oswalkdir, ospaths # In newer nim these are merged to os
+proc runTest(path: string) =
+  echo "\nRunning: ", path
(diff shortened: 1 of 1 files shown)


arnetheduck:
we tend to have an `all_tests.nim` that collects all tests in a single file, like here: https://github.com/status-im/nim-beacon-chain/blob/master/tests/all_tests.nim

this is nice because it allows to run individual tests from command line but requires only a single compile for the big test.. maybe something similar here would be good

stefantalpalaru:
This is a repo obtained from merging 5 different repos with separate test suites. There was a requirement to still be able to run those test suites individually. If we move to an "all_tests.nim", we also need 5 more "all_foo_tests.nim" besides it.

Some tests use "\*.json" files and I think the path to them is relative to the current directory- i.e.: the binary needs to be run in "tests/rlp" or "tests/trie".

arnetheduck:
> If we move to an "all_tests.nim", we also need 5 more "all_foo_tests.nim" beside it.

that's probably not a bad way to organize things

arnetheduck:
in particular, it seems better to keep it in nim files than in nimble

PR #29, opened 2019-03-26
Improved error-handling; Timeouts in all handshakes

zah:


zah:
For reviewers, here is what the hidden code from macros expands to:

Sending a message:
https://gist.github.com/zah/dbfafeed64b773d67a34627c1f20cc27#file-gossip_sub-nim-L54-L66

Handling incoming message:
https://gist.github.com/zah/dbfafeed64b773d67a34627c1f20cc27#file-gossip_sub-nim-L142-L149
called from
https://github.com/status-im/nim-eth/blob/12154af8f5cfa9ad35942338e1abddcff4058432/eth/p2p/rlpx.nim#L517

eth/p2p/rlpx.nim
```diff
@@ -1127,16 +1150,18 @@ macro handshake*(peer: Peer, timeout: untyped, sendCall: untyped): untyped =
...
 
-proc disconnect*(peer: Peer, reason: DisconnectionReason, notifyOtherPeer = true) {.async.} =
+proc disconnect*(peer: Peer, reason: DisconnectionReason, notifyOtherPeer = false) {.async.} =
   if peer.connectionState notin {Disconnecting, Disconnected}:
     peer.connectionState = Disconnecting
+    if notifyOtherPeer and not peer.transport.closed:
+      var fut = peer.sendDisconnectMsg(reason)
+      yield fut
+      if fut.failed:
+        warn "Failed to delived disconnect message", peer
```
arnetheduck: doesn't seem like this should be a warning - it's normal behavior that peers disconnect (at any point in time).. warn generally means action is needed by whoever is reading the log.

arnetheduck:
in general, it seems to me that it would be easier to control for timeouts generally for any response-inducing request - if a client doesn't respond on time, we could just drop them and let the normal client-disconnected procedures kick in.

once you start returning errors from sends, you need to introduce error handling there as well, which tends to clutter the code - unless you intend to handle errors differently depending on the send type, but that's rare. 

it's even more difficult when you start conflating errors from different layers - consider the case where you request a block and it doesn't exist - do you raise an exception? If yes, it means you're pushing both transport exceptions and application level exceptions up the same pipe, violating layering abstractions.

arnetheduck:
the last comment is more about the structural issues, not a blocker for this branch

zah:
All requests have a `timeout` parameter set to 10 seconds by default (this was another existing constant with a reasonable value). The request returns `Option[ResponseType]` and in case of a timeout, it will be `none`. 

It's a bit inconvenient at the moment that besides this `none` value, you can also get exceptions such as `PeerDisconnected` os `TransportOsError` (e.g. Broken Pipe) and you must handle them in the sending code (both of these will automatically disconnect the peer and perform the necessary clean-up in the PeerPool). I've considered changing the timeout to an exception or collapsing every possible error into the `none` value, but both have some drawbacks.

eth/p2p/rlpx.nim
```diff
@@ -1198,9 +1223,8 @@ proc postHelloSteps(peer: Peer, h: devp2p.hello) {.async.} =
 
   messageProcessingLoop.callback = proc(p: pointer) {.gcsafe.} =
     if messageProcessingLoop.failed:
-      error "dispatchMessages failed",
-            err = messageProcessingLoop.error.msg
-      asyncCheck peer.disconnect(ClientQuitting)
+      error "dispatchMessages failed", err = messageProcessingLoop.error.msg
+      asyncDiscard peer.disconnect(ClientQuitting)
```
arnetheduck: I thought for a bit about `asyncDiscard` - and like it currently works, it will swallow overflows and other.. er, `Defect`'s silently in addition to the intended behavior of ignoring "normal" errors

zah: `asyncDiscard` doesn't do anything at the moment. It's an "insurance" syntax for possible problems discovered in the future

arnetheduck: yeah. meaning that the general async transformation exception handler will stuff the exception in the future and never re-raise it.. right?

zah: But you are right we can have a third option that will panic on defects.

PR #30, opened 2019-03-27
fix nimbus/premix appveyor build

jangko:
this PR fix recent nimbus appveyor failed build

### DIFF
--- a/eth/p2p/rlpx.nim
+++ b/eth/p2p/rlpx.nim
@@ -1121,7 +1121,7 @@ proc handshakeImpl[T](peer: Peer,
                       sendFut: Future[void],
                       responseFut: Future[T],
                       timeout: Duration): Future[T] {.async.} =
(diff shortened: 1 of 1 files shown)


stefantalpalaru:
If this fix works, Chronicles is off the hook here: https://github.com/status-im/nim-chronicles/issues/50

jangko:
it works on my windows machine.
technically, it is not gcsafe, due to chronicles `debug` presence.
but we can pretend it is gcsafe, at least for release build

stefantalpalaru:
If we pretend now, the root cause will come back to bite us later.

jangko:
I see some of the other callbacks also marked `{.gcsafe.}` (not only me have the same idea here)
it will bite us back at many places. unless we can make chronicles evade nim compiler sensor detection like `debugEcho`.

PR #31, opened 2019-03-28
Add a new asyncDiscard replacement that traces recoverable errors and aborts on defects

zah:


yglukhov:
How about (a) adding this to chronos, and (b) provide only an explicit way of specifying error handler in case of CatchableError, eg.
```
myAsyncProc().asyncCheck do(e: CatchableError):
  # my handler
```

zah:
We can do this, but let's revisit it after the testnet launch.

yglukhov:
> We can do this, but let's revisit it after the testnet launch.

Why not now? It's like a 5 minutes change.

zah:
@cheatfate can add the helper in Chronos when he has the time, but even with it, we'll could still use the chronicles-enable wrapper in nim-eth, anyway.

PR #34, opened 2019-04-08
Retry bootstrapping in case of failure

yglukhov:


eth/p2p/kademlia.nim
```diff
@@ -416,11 +416,21 @@ proc resolve*(k: KademliaProtocol, id: NodeId): Future[Node] {.async.} =
...
+  ## Bond with bootstrap nodes and do initial lookup. Retry `retries` times
+  ## in case of failure, or indefinitely if `retries` is 0.
+  var numTries = 0
+  while true:
+    let bonded = await all(bootstrapNodes.mapIt(k.bond(it)))
+    if true notin bonded:
+      info "Failed to bond with bootstrap nodes"
+      inc numTries
+      if retries == 0 or numTries < retries:
+        info "Retrying"
```
arnetheduck: can we be a bit more.. informative here? ie at info, try to run the app and look at the output - envision that you're reading the log, and this suddenly this pops up - hard to form a story around what it is?

eth/p2p/kademlia.nim
```diff
@@ -416,11 +416,21 @@ proc resolve*(k: KademliaProtocol, id: NodeId): Future[Node] {.async.} =
...
-  let bonded = await all(bootstrapNodes.mapIt(k.bond(it)))
-  if true notin bonded:
-    info "Failed to bond with bootstrap nodes "
-    return
+proc bootstrap*(k: KademliaProtocol, bootstrapNodes: seq[Node], retries = 0) {.async.} =
+  ## Bond with bootstrap nodes and do initial lookup. Retry `retries` times
+  ## in case of failure, or indefinitely if `retries` is 0.
+  var numTries = 0
+  while true:
+    let bonded = await all(bootstrapNodes.mapIt(k.bond(it)))
```
arnetheduck: from the discussions about `all`, this will raise if _any_ bonding raises.. is that the intended semantic?

yglukhov: Yes, `bond` should not raise.

arnetheduck: but it does, from what I can tell - ie socket send will raise, etc

arnetheduck: or is that swallowed somewhere? hard to tell

eth/p2p/kademlia.nim
yglukhov: Addressed in #35

eth/p2p/kademlia.nim
yglukhov: Is is swallowed, here: https://github.com/status-im/nim-eth/blob/master/eth/p2p/discovery.nim#L109-L112

zah: This could use the new fancier `traceAsyncErrors` helper btw. The goal would be to abort on defects.

arnetheduck: this will come back to bite us - as a design it's very brittle - the swallowing hides errors at a different abstraction level than bond - it's very easy to accidentally change this and not notice until we deploy this to users - specially because it sits at the very core of the networking

PR #35, opened 2019-04-09
Fixed discovery busyloop when no bootnodes provided

yglukhov:
Also made logging more informative

PR #38, opened 2019-04-10
Workaround compilation error caused by chronos newFuture changes

yglukhov:


yglukhov:
I've moved this fix to #35, because the fixes work only together :)

zah:
While this change is relatively harmless, I've pushed a more proper fix in Chronos itself:
https://github.com/status-im/nim-chronos/pull/35

Without the fix there, the error might get back to bite us if another overload of `newFuture` is introduced.

cheatfate:
@zah i'm curious why you need your own overload of newFuture[T]?

zah:
@cheatfate, my overload can be added to Chronos. It's slightly nicer in the sense that it doesn't require you to repeat the type of the future in the code.

```nim
type
  Foo = object
    fut: Future[int]

proc init(T: type Foo) =
  newFuture res.fut # no need to specify `int` here.
```

Issue #37, opened 2019-04-10
Transports not closed on disconnect() or when connection is dropped


kdeme:
As discussed on gitter, there are some transport leaks currently. Creating this issue so that there is some traceability and that you know that I'm looking into it.

Basically on every `disconnect()`, unless it is followed specifically with a `peer.transport.close()` (which is never?), we leak transports.
And also on a dropped connection (https://github.com/status-im/nim-eth/blob/master/eth/p2p/rlpx.nim#L518) it will not be closed currently.

Transport does get closed when a disconnect message is received: https://github.com/status-im/nim-eth/blob/master/eth/p2p/rlpx.nim#L526

PR #39, opened 2019-04-11
Bug/fix transport leaks #37

kdeme:
Proposal PR for fixing #37 

Transport close is now done either in:
- `disconnect()`: in case in Connected state
- `rlpxConnect` or `rlpxAccept`: in case of failure before Connected state 

`dispatchMessages` loop needed some logic to differentiate these situations. Including the one where a connection is dropped from the other side.

I decided not to use an `AsycEvent` in the end (was one of the suggestions) because:
1. It is not necessary, we can check `connectionState`
2. I feel that then we should use more Events to be consitent. E.g. also when `Connecting` fails, or when sending of messages could fail due to a `disconnect` or dropped connect.

It does have as drawback that now the `await peer.recvMsg()` will have to wait until `TransportIncompleteError` in case it happens to be at that point in the loop when a `disconnect` is done. 

It could be that we now see "broken pipe" or "Transport is already closed" errors again, e.g. when a send would be done after transport closure (this was before also possible, it is just that transport were not getting closed :) ). To be monitored...

### DIFF
--- a/eth/p2p/rlpx.nim
+++ b/eth/p2p/rlpx.nim
@@ -511,22 +511,34 @@ proc nextMsg*(peer: Peer, MsgType: type): Future[MsgType] =
 # handled a layer lower for clarity (and consistency), as also the actual
 # message handler code as the TODO mentions already.
 proc dispatchMessages*(peer: Peer) {.async.} =
-  while true:
+  while peer.connectionState notin {Disconnecting, Disconnected}:
     var msgId: int
     var msgData: Rlp
     try:
       (msgId, msgData) = await peer.recvMsg()
     except TransportIncompleteError:
-      trace "Connection dropped, ending dispatchMessages loop", peer
-      # This can happen during the rlpx connection setup or at any point after.
-      # Because this code does not know, a disconnect needs to be done.
-      await peer.disconnect(ClientQuitting)
+      # Note: Could also "Transport is already closed!" error occur? Might have
+      # to change here to the general TransportError.
+      case peer.connectionState
+      of Connected:
+        # Dropped connection, still need to cleanup the peer.
+        # This could be seen as bad behaving peer.
+        trace "Dropped connection", peer
+        await peer.disconnect(ClientQuitting, false)
+        return
+      of Disconnecting, Disconnected:
+        # Graceful disconnect, can still cause TransportIncompleteError as it
+        # could be that this loop was waiting at recvMsg().
+        return
+      else:
+        # Connection dropped while `Connecting` (in rlpxConnect/rlpxAccept).
+        return
+    except PeerDisconnected:
       return
 
     if msgId == 1: # p2p.disconnect
-      await peer.transport.closeWait()
       let reason = msgData.listElem(0).toInt(uint32).DisconnectionReason
-      await peer.disconnect(reason)
+      await peer.disconnect(reason, false)
       break
 
     try:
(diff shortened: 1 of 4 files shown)


eth/p2p/rlpx.nim
```diff
@@ -1171,18 +1183,29 @@ macro handshake*(peer: Peer, timeout: untyped, sendCall: untyped): untyped =
...
+      yield fut
+      if fut.failed:
+        debug "Failed to deliver disconnect message", peer
+
+      proc waitAndClose(peer: Peer, time: Duration) {.async.} =
+        await sleepAsync(time)
+        await peer.transport.closeWait()
+
+      # Give the peer a chance to disconnect
+      await peer.waitAndClose(2.seconds)
```
yglukhov: I'm not a fan of the fact that `peer.disconnect(myReason)` called from high level code in normal conditions will always hang for 2 seconds.

yglukhov: I'm not a fan of the fact that `peer.disconnect(myReason)` called from high level code in normal conditions will always hang for 2 seconds. Can we do better?

kdeme: True, should be async or called somewhere else I'll think about this. This PR in general needs some rethinking around the `disconnect` call.

yglukhov:
I still find the peer lifetime pretty convoluted. My idea was running the `dispatchMessages` early in the lifetime, probably even before the devp2p handshake, and do the disconnection from there. I understand it might be not as straightforward, but I'll try it if you gave up :)

kdeme:
Have to be careful with starting the `dispatchMessages` (in its current form) to soon. It was already moved a step later some time ago (https://github.com/status-im/nim-eth-p2p/commit/58d6c9c208402144965e158cd7519948f402cd1b) because of issues with handshake procedures not started yet versus the first incoming messages being processed already. Of course `dispatchMessages` could check for example the `Connected` state or so.
Anyway, didn't give up (just) yet, just paused it and let it linger in my mind a bit :)

cheatfate:
Using variables as flags for the state `disconnecting`, `disconnected` is very bad idea here, because you can stuck inside of `recvMessage` while other task can disconnect `peer`. And this will become race condition as we have it now.

kdeme:
Made the 2 seconds wait + close now no longer blocking.

I believe that in its current implementation `recvMessage` could only hang if it does not throw an exception when the connection gets closed. This does happen, but if this behaviour changes or if this behaviour is not the same for different OSes than this might be a problem. 

It is not so error prone though and mistakes can be easily made if the `transport.close` is moved. As @cheatfate  suggests, using AsyncEvent might be a better way.

eth/p2p/rlpx.nim
```diff
@@ -1176,18 +1188,29 @@ macro handshake*(peer: Peer, timeout: untyped, sendCall: untyped): untyped =
...
+      yield fut
+      if fut.failed:
+        debug "Failed to deliver disconnect message", peer
+
+      proc waitAndClose(peer: Peer, time: Duration) {.async.} =
+        await sleepAsync(time)
+        await peer.transport.closeWait()
+
+      # Give the peer a chance to disconnect
+      asyncCheck peer.waitAndClose(2.seconds)
```
yglukhov: I'd keep away from `asyncCheck`. Use `traceAsyncErrors` instead.

kdeme: right, done.

PR #43, opened 2019-04-17
More whisper changes

kdeme:
We should not disconnect form a peer when we don't support a particular packet code, see also here: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-627.md#specification

Currently `ValueError` was raised, which is apparently a`Defect`. I've added a specific error that inherits from `CatchableError`. In the future this problem would have probably resolved 'itself' as I noticed that `ValueError` is changed to inherit from `CatchableError` (see https://github.com/nim-lang/Nim/commit/87232a38e4fc40a3acb154a78bf32f8ea5734d13). Anyhow, lets not rely on that.

Also added the actual unimplemented procs for the specific Status packet codes for clarity, this is not necessary though.

### DIFF
--- a/eth/p2p/private/p2p_types.nim
+++ b/eth/p2p/private/p2p_types.nim
@@ -70,6 +70,7 @@ type
     # protocol to a peer that doesn't support the protocol.
 
   MalformedMessageError* = object of CatchableError
+  UnsupportedMessageError* = object of CatchableError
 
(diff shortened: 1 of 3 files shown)


kdeme:
Can this be reviewed/merged please? 

Except for the `CatchableError` change, its only cosmetics

PR #45, opened 2019-04-18
Fix to only allow sync for peer with eth support

kdeme:
This is to fix a crash that occurs when connecting to a node that does not support the eth sub-protocol. E.g. a node that only supports whisper.

Following stacktrace would occur:
```
Traceback (most recent call last)
nimbus.nim(147)          nimbus
nimbus.nim(126)          process
asyncloop.nim(266)       poll
asyncmacro2.nim(36)      connectToNode_continue
peer_pool.nim(119)       connectToNodeIter
peer_pool.nim(111)       addPeer
blockchain_sync.nim(334) :anonymous
blockchain_sync.nim(317) onPeerConnected
asyncmacro2.nim(306)     startSyncWithPeer
asyncmacro2.nim(36)      startSyncWithPeer_continue
blockchain_sync.nim(293) startSyncWithPeerIter
asyncmacro2.nim(306)     peersAgreeOnChain
asyncmacro2.nim(36)      peersAgreeOnChain_continue
p2p_backends_helpers.nim(9) peersAgreeOnChainIter
uint_public.nim(56)      <
uint_comparison.nim(20)  <
```
This is because the `eth.state` is being accessed when doing the sync.

### DIFF
--- a/eth/p2p/blockchain_sync.nim
+++ b/eth/p2p/blockchain_sync.nim
@@ -331,10 +331,12 @@ proc onPeerDisconnected(ctx: SyncContext, p: Peer) =
 proc startSync(ctx: SyncContext) =
   var po: PeerObserver
   po.onPeerConnected = proc(p: Peer) {.gcsafe.} =
-    ctx.onPeerConnected(p)
+    if p.supports(eth):
+      ctx.onPeerConnected(p)
 
   po.onPeerDisconnected = proc(p: Peer) {.gcsafe.} =
-    ctx.onPeerDisconnected(p)
+    if p.supports(eth):
+      ctx.onPeerDisconnected(p)
 
   ctx.peerPool.addObserver(ctx, po)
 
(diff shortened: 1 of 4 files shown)


kdeme:
Come to think of it, it probably would be nicer to give the observer an option to only run for peers with a certain protocol. I'll have a go at that.

yglukhov:
Maybe it would be better to not even let non-eth peers into the peerpool? Otherwise the peerpool might be populated with useless peers.

kdeme:
@yglukhov If the node has a focus on Whisper enabled peers, then why should a whisper only peer be seen as useless?

Sure, nodes with eth and whisper would be best, but you might not have that luxury

kdeme:
Both ways trigger another underlying problem (https://github.com/status-im/nim-eth/issues/46). Last commit resolves that.

yglukhov:
> If the node has a focus on Whisper enabled peers, then why should a whisper only peer be seen as useless?

Well my idea was to allow a user customizable predicate of what is a useful peer, but you're right that it's not sufficient on its own.

Which leads me to an opinion that the changes in this PR are useful regardless. While a more sophisticated interface for peerpool observers to subscribe for "interesting" peers only would be unnecessarily complex.

PR #52, opened 2019-05-09
Added randomPeers iterator with protocol filter

yglukhov:


eth/p2p.nim
```diff
@@ -154,8 +154,15 @@ proc randomPeer*(node: EthereumNode): Peer =
     if i == peerIdx: return peer
     inc i
 
+iterator randomPeers*(node: EthereumNode, maxPeers: int, Protocol: type): Peer =
```
zah: If functions like this become too widely used, it will be worth maintaining per-protocol Peer collections in the library.

yglukhov: Yeah, still I hope we won't have so many peers in the pool for this to have any visible difference :)

PR #55, opened 2019-05-23
Add several checks for currently unhandled errors

kdeme:
Checks needed against possible `AssertionError`s being raised in discovery protocol:
- need check if there is packet-data (rlp payload)
- need check if packet-data is rlp list
- depending on packet id, need to check if minimum size of list.
- specifically for FindNode Packet, need to check if the size of target public key

Need also to check on packet id range, else `RangeError` is possible (the code would actually work correct in `-d:release`, but that is to tricky)

And need to catch all types of `RlpError`.

Don't merge yet, will still add the error cases in a test. And might rework the code somewhat also.

### DIFF
--- a/eth/rlp.nim
+++ b/eth/rlp.nim
@@ -275,7 +275,7 @@ proc listElem*(self: Rlp, i: int): Rlp =
 
 proc listLen*(self: Rlp): int =
   if not isList():
-    raise newException(RlpTypeMismatch, "List expected, but the source RLP is not a list.")
+    return 0
 
   var rlp = self
   for elem in rlp:
--- a/eth/rlp.nim
+++ b/eth/rlp.nim
@@ -26,7 +26,7 @@ type
     of rlpList:
       elems*: seq[RlpNode]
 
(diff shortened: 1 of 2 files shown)


eth/rlp.nim
```diff
@@ -275,7 +275,7 @@ proc listElem*(self: Rlp, i: int): Rlp =
 
 proc listLen*(self: Rlp): int =
   if not isList():
-    return 0
+    raise newException(RlpTypeMismatch, "List expected, but the source RLP is not a list.")
```
zah: While it's very good that you're changing RLP to use `CatchableError`, we have to be very careful when changing APIs like this because this may have unexpected consequences for `nim/eth/trie` and Nimbus in general. I'd suggest running the change through @jangko who can verify that it's not causing any regressions.

jangko: @kdeme, please revert this one(listLen), hexary-trie actually depends on this behavior for it's internal algorithm. If you really need `listLen` with `raise`, then you should implement a new one rather than change the current one.

PR #60, opened 2019-06-05
[WIP] Some improvements to RLPx.

cheatfate:


eth/p2p/rlpx.nim
```diff
@@ -255,6 +255,16 @@ proc supports*(peer: Peer, Protocol: type): bool {.inline.} =
 template perPeerMsgId(peer: Peer, MsgType: type): int =
   perPeerMsgIdImpl(peer, MsgType.msgProtocol.protocolInfo, MsgType.msgId)
 
+proc join*(peer: Peer): Future[void] =
+  ## Wait until `peer` is alive.
```
kdeme: Its actually waiting till the peer is dead

zah: I understand why the name `join` was used, but it seems a bit overloaded with possible meanings. Perhaps we can name it `waitDisconnect`?

arnetheduck:
obsolete

PR #64, opened 2019-06-13
Clean-up some exceptions

kdeme:
clean up some exception logging in blockchain sync due to transport close + getting rid of Asserts I notice being thrown higher up

### DIFF
--- a/eth/common/eth_types.nim
+++ b/eth/common/eth_types.nim
@@ -230,7 +230,11 @@ proc read*(rlp: var Rlp, T: typedesc[StUint]): T {.inline.} =
   if rlp.isBlob:
     let bytes = rlp.toBytes
     if bytes.len > 0:
-      result.initFromBytesBE(bytes.toOpenArray)
+      # be sure the amount of bytes matches the size of the stint
+      if bytes.len <= sizeof(result):
+        result.initFromBytesBE(bytes.toOpenArray)
+      else:
+        raise newException(RlpTypeMismatch, "Unsigned integer expected, but the source RLP has the wrong length")
     else:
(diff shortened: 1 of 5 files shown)


eth/common/eth_types.nim
```diff
@@ -310,7 +312,7 @@ proc rlpHash*[T](v: T): Hash256 =
 func blockHash*(h: BlockHeader): KeccakHash {.inline.} = rlpHash(h)
 
 proc notImplemented =
```
zah: Where are we using this? Perhaps the time has come to implement all such missing methods.

kdeme: Yes, I realize it is arguably better to have the "Not implemented" as being a real `AssertionError`.
I mainly did this so we wouldn't disconnect with the peer, without having to remove some code in `eth_protocol`. 

Better would be to implement the missing parts. If necessary I can revert this part of the change

kdeme: I've decided to change it to keep the assertion and just comment the non implemented parts for now.
Created an issue for the implementation: https://github.com/status-im/nim-eth/issues/65

kdeme:
I looked into the case triggering the issue resolved by last commit:
It was occuring at an `eth_protocol` `status` message (handshake) with a `DifficultyInt` of 33 bytes.
I've also gone through the full Rlp data that was send, our decoding looks fine.
Offending peer had `peerId` of `Geth/v1.5.0-unstable/linux/go1.9.2`, looks old and unstable ;)

The second commit is for a case where there was no list in the rlp data for `hello_thunk` (p2p hello message). I've only seen it once, so I could not further investigate.

PR #69, opened 2019-06-17
replace waitFor with await

kdeme:


stefantalpalaru:
You might need to merge https://github.com/status-im/nim-eth/pull/68 first, and then rebase, so you get the right Nim version.

PR #71, opened 2019-06-21
Fix for hanging protocol_handlers test

kdeme:
Do not merge this yet, even if tests succeed.

I will still add another commit to make the connections happening in the test more reliable/predictable.
However, I want to see test results with this change to see if it is the same underlying issue that I spotted. But, even if it is, the test might still hang (for a while?) due to incoming & outgoing connection races between the two nodes.

### DIFF
--- a/eth/p2p/rlpx.nim
+++ b/eth/p2p/rlpx.nim
@@ -1291,6 +1291,13 @@ proc postHelloSteps(peer: Peer, h: devp2p.hello) {.async.} =
   #
   await all(subProtocolsHandshakes)
 
+  # This is needed as a peer might have already disconnected. In this case
+  # we need to raise so that rlpxConnect/rlpxAccept fails.
+  # Disconnect is done only to run the disconnect handlers. TODO: improve this
+  # also TODO: Should we discern the type of error?
+  if messageProcessingLoop.finished:
+    await peer.disconnectAndRaise(ClientQuitting,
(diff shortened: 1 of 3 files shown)


kdeme:
OK, looks good now, ready to review/merge.

Basically the underlying issue that I spotted is that it is possible that a peer drops connection before the end in `postHelloSteps` in  `rlpxConnect`/`rlpxAccept` is reached. In this case no error is raised, and this results in the peer being added to the peerPool.
This would specifically create issues in `test_protocol_handlers.nim` as there we would wait for both nodes to connect to the network. When one node would have this issue, it would think it is connected to the peer, but actually isn't, hence it is not trying to reconnect, hence the other node would wait forever.
The first commit resolves this (although improvements could be made, created issue for this: https://github.com/status-im/nim-eth/issues/72).

But next to this, there would be races in general between incoming and outgoing connections. This is why I originally set 1 of the 2 nodes not to listen. However, that does not stop them from trying to connect (to the not listening one). You can see this by the builds that do pass now, but where it takes a quite long and the time it takes is very different between builds.

To solve this, the second and third commit makes the connection setup more regulated. Notice also that build times have decreased a lot.

As to why this all pops up now: this must be due to the changes in https://github.com/status-im/nim-eth/commit/f8bdec88c9afa1559208f67a5ccb9cfebeece1da . That commit made disconnects in case of duplicate connections happen sooner in the process.

PR #73, opened 2019-06-26
Fix leading zeroes bug + add tests from geth and parity

kdeme:
Triggered by https://github.com/ethereum/go-ethereum/pull/19330 and https://github.com/ethereum/go-ethereum/pull/19753, I tested and noticed a difference in the best bits calculation of our implementation.

Not sure where the `+ 1` came from, but with this change that part of the PoW calculation should be the same for geth, parity and this implementation (and also in line with EIP-627).

The size calculation is still different though (and not in line with EIP-627), see comments in code & test.
We can decide here to follow spec or not, currently geth does not either (albeit in a different way than us, we have the old version of geth basically).

Should be noted also that the last changes in geth will have an impact in old versus new whisper nodes compatibility. More specifically, in possible many dropped whisper messages due to a too low PoW (as the hash will be different between old and new clients).

### DIFF
--- a/eth/p2p/rlpx_protocols/whisper_protocol.nim
+++ b/eth/p2p/rlpx_protocols/whisper_protocol.nim
@@ -155,12 +155,12 @@ proc leadingZeroBits(hash: MDigest): int =
       result += countLeadingZeroBits(h)
       break
 
-proc calcPow(size, ttl: uint64, hash: Hash): float64 =
+proc calcPow*(size, ttl: uint64, hash: Hash): float64 =
   ## Whisper proof-of-work is defined as the best bit of a hash divided by
   ## encoded size and time-to-live, such that large and long-lived messages get
   ## penalized
 
-  let bits = leadingZeroBits(hash) + 1
+  let bits = leadingZeroBits(hash)
   return pow(2.0, bits.float64) / (size.float64 * ttl.float64)
 
 proc topicBloom*(topic: Topic): Bloom =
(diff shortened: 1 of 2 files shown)


stefantalpalaru:
> possible many dropped whisper messages due to a too low PoW

I'm seeing that already:

```text
WRN 2019-06-26 14:42:02+02:00 Message PoW too low                        topics="whisper" tid=6754 file=../vendor/nim-eth/eth/p2p/rlpx_protocols/whisper_protocol.nim:509 minPow=0.2 pow=0.0003427592116538132
WRN 2019-06-26 14:42:02+02:00 Message PoW too low                        topics="whisper" tid=6754 file=../vendor/nim-eth/eth/p2p/rlpx_protocols/whisper_protocol.nim:509 minPow=0.2 pow=0.0006578947368421052
WRN 2019-06-26 14:42:02+02:00 Message PoW too low                        topics="whisper" tid=6754 file=../vendor/nim-eth/eth/p2p/rlpx_protocols/whisper_protocol.nim:509 minPow=0.2 pow=0.01052631578947368
WRN 2019-06-26 14:42:02+02:00 Message PoW too low                        topics="whisper" tid=6754 file=../vendor/nim-eth/eth/p2p/rlpx_protocols/whisper_protocol.nim:509 minPow=0.2 pow=0.0003427592116538132
WRN 2019-06-26 14:42:02+02:00 Message PoW too low                        topics="whisper" tid=6754 file=../vendor/nim-eth/eth/p2p/rlpx_protocols/whisper_protocol.nim:509 minPow=0.2 pow=0.0006578947368421052
WRN 2019-06-26 14:42:02+02:00 Message PoW too low                        topics="whisper" tid=6754 file=../vendor/nim-eth/eth/p2p/rlpx_protocols/whisper_protocol.nim:509 minPow=0.2 pow=0.01052631578947368
```

kdeme:
Yes, exactly, because our implementation already uses the same hashing algo as in the EIP-627 (and as Parity). So these messages are the proof of the incompatibility of geth whisper versus nimbus whisper.

Now, geth will get this same issue with old versus new nodes. This will also impact the Status client I'd guess.

PR #80, opened 2019-07-10
use nim-metrics

stefantalpalaru:


eth/common/utils.nim
```diff
@@ -8,5 +10,5 @@ proc parseAddress*(hexString: string): EthAddress =
 proc `$`*(a: EthAddress): string =
   a.toHex()
 
-var nimbusStats*: NimbusStats
+var peerGauge* = newGauge("connected peers", "number of peers in the pool")
```
arnetheduck: eventually would be nice to have a less demanding api where you don't need to register first - something like https://github.com/Shopify/statsd-instrument#usage - this is important so that it's "cheap" to instrument stuff

zah: I'd prefer if the global metric declarations are hidden behind a template, so it will be easier to eliminate them with a compile-time switch in the future. For example:

```nim
gauge peerGauge, "connected peers", ...
```

zah: Also, since the current metrics are `ref` objects, this will create problems for `gcsafe`-ty. How are you addressing this problem?

eth.nimble
```diff
@@ -13,7 +13,8 @@ requires "nim >= 0.19.0",
          "chronicles",
          "stew",
          "result",
-         "nat_traversal"
+         "nat_traversal",
+         "https://github.com/status-im/nim-metrics"
```
zah: The Nimble https dependencies are notoriously unreliable at the moment. Please change to dependency to just "metrics" once this is merged:
https://github.com/nim-lang/packages/pull/1113

eth/p2p/peer_pool.nim
```diff
@@ -107,10 +107,11 @@ proc getRandomBootnode(p: PeerPool): Option[Node] =
   if p.discovery.bootstrapNodes.len != 0:
     result = option(p.discovery.bootstrapNodes.rand())
 
-proc addPeer*(pool: PeerPool, peer: Peer) =
+proc addPeer*(pool: PeerPool, peer: Peer) {.gcsafe.} =
   doAssert(peer.remote notin pool.connectedNodes)
   pool.connectedNodes[peer.remote] = peer
-  inc(nimbusStats.num_peers)
+  {.gcsafe.}:
```
kdeme: So will we need to add this `.gcsafe.` for each location we put a metric ? Can't we hide this away somehow?

eth.nimble
stefantalpalaru: Will do.

eth/p2p/peer_pool.nim
stefantalpalaru: How? I'm open to ideas.

eth/common/utils.nim
zah: I'm talking about a more surgical metrics removal targeting individual metrics and modules. 

The gcsafety issues can be hidden by the user if the user-facing APIs are also one-liner templates that put each operation in a `gcsafe:` block.

stefantalpalaru: You'd need to add another symbol in there, so instead of
```nim
{.gcsafe.}:
  myCounter.inc()
```
you have something like
```nim
metric myCounter.inc()
```
Not much of a gain.

Can you come up with a scenario in which you need to disable individual metrics? You usually want instrumentation to be all on or all off.

zah: This was discussed at length in Gitter. When we lack support for removing metrics from a particular library, this creates a trade-off for the library creator. She can either include metrics and increase the cost of the library for the users, or she can avoid them and thus fail to bring additional usefulness to the library. This trade-offs resides on a spectrum where the number of metrics is also relevant (a library may feature essential and obscure metrics at the same time).

zah: I think the usage can safely be just `myCounter.inc()`. I can take a look at the problems myself if you wish.

stefantalpalaru: Keep in mind that different collector types have different procedures modifying them.

arnetheduck: re gitter discussion - we also mentioned that as soon as you add logging or, to a lesser degree, metrics, this information disappears from the public api (for API convenience).. thus compile-time disabling really means removing functionality that is not replaced by something else, meaning basically that the library becomes crippled.

about declaration+use, it leads to more resistance to add new metrics, and leaves cruft behind when the use is removed but declaration remains - it also creates tricky situation where the same metric can be reported from multiple places, a design that's probably worth avoiding, again due to maintenance issues (for example, when investigating a metric, you now have to consider more code paths that could have led to it being changed)

stefantalpalaru: > tricky situation where the same metric can be reported from multiple places, a design that's probably worth avoiding

Not possible. It was necessary to increment and decrement the same metric in different places in this very PR, because that's where the peers were added and removed.

eth/p2p/peer_pool.nim
```diff
@@ -107,10 +107,11 @@ proc getRandomBootnode(p: PeerPool): Option[Node] =
...
+    peerGauge.inc()
```
arnetheduck: since we know how many nodes there are here, better use an absolute value

stefantalpalaru: If peer addition and removal are always done in these procedures, there should be no difference.

arnetheduck: it's an unnecessary assumption to make that only makes the code more brittle. 

also, when you do statsd the server may lose the current value (for example in a restart) - using absolutes is more robust in cases like this

stefantalpalaru: We store that value on the client side, so it only disappears when the instrumented program is restarted.

arnetheduck: ok, so I've mentioned a few advantages, such as there being less state and order dependence by supplying absolute values. 

can you think of any advantages of repeating the inc/dec calculation a third time here (the second inc/dec is hidden in connectednodes.add/remove)?

stefantalpalaru: > there being less state

A collector's state is always there, whether you change it with inc()/dec() or set().

> can you think of any advantages of repeating the inc/dec calculation a third time here (the second inc/dec is hidden in connectednodes.add/remove)?

No, but I don't see any advantage to set() either.

Wait, there is an advantage to increasing and decreasing a gauge instead of storing `pool.connectedNodes.len` into it: that's how I noticed it became negative at some point, due to a bug in `removePeer()` where the same peer was removed multiple times.

PR #85, opened 2019-07-16
Add max open files limit of 512 to rocksdb init

kdeme:
Requires https://github.com/status-im/nim-rocksdb/pull/20

Change rocksdb default of -1 to 512, as we reach the 1024 usual (?) open files limit in Linux when running Nimbus at around block ~1.300.000.
I could not immediately find what the default limit is in Windows, so perhaps I just have to test it.

Setting of 512 is somewhat based on value used in parity client: https://github.com/paritytech/parity-common/blob/master/kvdb-rocksdb/src/lib.rs#L197
Should be quite safe, and if needed can probably be increased safely to somewhere between 700-800. Will need incoming connection limitations for that, but that is another issue.

Could also forward this setting to Nimbus + Nimbus config but decided to not yet do this until it is proven to have an practical impact. 
I'll create an issue to benchmark + (if needed) tune rocksdb at later moment.

### DIFF
--- a/eth/trie/backends/rocksdb_backend.nim
+++ b/eth/trie/backends/rocksdb_backend.nim
@@ -7,6 +7,10 @@ type
 
   ChainDB* = RocksChainDB
 
+# Maximum open files for rocksdb, set to 512 to be safe for usual 1024 Linux
+# limit per application
+const maxOpenFiles = 512
+
 proc get*(db: ChainDB, key: openarray[byte]): seq[byte] =
   let s = db.store.getBytes(key)
   if s.ok:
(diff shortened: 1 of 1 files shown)


arnetheduck:
one thing to consider is tcp connection spamming, and how that affects open files - good to have some margin in general and not push it too far if there's no clear advantage.. ie between 500 and 800  open files on the rocksdb side, what changes? does it have a bigger impact than having 200 instead of 500 spare files for other purposes?

kdeme:
Yes, that's what I mean with "Will need incoming connection limitations for that, but that is another issue." in my original comment.

I was going to make an issue regarding that. To investigate having some better mechanisms regarding connection limiting. 
However, without this change, things are much worse actually, we have no margin at all.

PR #92, opened 2019-08-17
Allow longer rlpx protocol names (for compatability with swarm)

nolash:
One of the protocols of the [Swarm](https://github.com/ethersphere/swarm) suite has a four-letter protocol name (`hive`)

### DIFF
--- a/eth/p2p/rlpx.nim
+++ b/eth/p2p/rlpx.nim
@@ -581,7 +581,7 @@ proc p2pProtocolBackendImpl*(protocol: P2PProtocol): Backend =
 
   if protocol.rlpxName.len == 0: protocol.rlpxName = protocol.name
   # By convention, all Ethereum protocol names must be abbreviated to 3 letters
-  doAssert protocol.rlpxName.len == 3
+  doAssert protocol.rlpxName.len > 2 
 
   new result
 


eth/p2p/rlpx.nim
```diff
@@ -581,7 +581,7 @@ proc p2pProtocolBackendImpl*(protocol: P2PProtocol): Backend =
 
   if protocol.rlpxName.len == 0: protocol.rlpxName = protocol.name
   # By convention, all Ethereum protocol names must be abbreviated to 3 letters
-  doAssert protocol.rlpxName.len == 3
+  doAssert protocol.rlpxName.len > 2
```
cheatfate: I think it would be better to use
```nim
doAssert protocol.rlpxName.len > 2 and protocol.rlpxName.len < 5
```

arnetheduck:
Is it documented anywhere how long the names "should" be? ie an official upstream guideline?

kdeme:
I could be looking over it but I can't immediately find it in the devp2p specs: https://github.com/ethereum/devp2p/blob/master/rlpx.md#capability-messaging
I just read there `A capability is identified by a short ASCII name and version number.` 

go-ethereum code mentions this: https://github.com/ethereum/go-ethereum/blob/master/p2p/protocol.go#L29 
But I don't think it is actually enforced in the code.

Parity does seem to limit it to 3 chars: 
https://github.com/paritytech/parity-ethereum/blob/bf55db4c7e3b4298597aec868713428a18d53e4d/util/network/src/lib.rs#L64
https://github.com/paritytech/parity-ethereum/blob/66e4410be72da38dd001b7ad6c9e99b18cfae9ad/ethcore/sync/src/api.rs#L63

If we want to increase it to allow for `hive` then I agree with @cheatfate and we should limit it to 4 (or 5 chars).

nolash:
5 = 3 + 1 each for "short" and "often" :dango:

zah:
The 3-letter limit used to be specified in the ETH1-related wikis, but I've seen other protocols in the wild that violate this informal spec, so we can indeed remove the restriction.

arnetheduck:
generally, this is the kind of background / domain knowledge / rationale that ideally would would be added as a comment in the code

PR #94, opened 2019-08-20
Small clarification to protocol name length doc

Swader:
As per conversation in PR.

Also testing how the docgen will react...


arnetheduck:
this is good, but even better if we also provide (authoritative) sources - I believe @kdeme tracked down a few and this way if there's doubt in the future, there's no need to redo the work

PR #96, opened 2019-09-03
[WIP] Fuzzing additions

kdeme:
Don't merge yet, playing around with libFuzzer and will do some more changes.


kdeme:
This is sorta ready to merge or review. That is, it is still very PoCish, however I'm undecided on how to make this convenient to use for a specific module that needs to be fuzzed. 

The current code for compiling and starting the fuzzing is currently done in nimscript, because that way it can be conveniently called from e.g. nimble:
```
import tests/fuzzing/fuzz_helpers

task fuzz_afl, "afl fuzz a target":
  if paramCount() < 2:
    echo "Need a fuzzing target"
  runFuzzer(paramStr(2), afl)
```
However, this will now only work in nim-eth, as that is where the code currently resides. 
The idea would be to move this code so that other modules can use it, and then this will no longer be possible.

Another option would be to make this fuzz script in a properly packaged application, with much better cmd line handling, functionality, etc. That application could then be a dependency in the nimble file of the module. However, this would probably be a bit less flexible (need to provide all the necessary options @ cmd line), and perhaps overkill.

tests/fuzzing/fuzztest.nim
```diff
@@ -0,0 +1,59 @@
...
+
+template test*(body: untyped): untyped =
+  when defined(standalone):
+    var payload {.inject.} = readStdin()
+
+    fuzz: `body`
+  else:
+    proc fuzzerCall(data: ptr byte, len: csize):
+        cint {.exportc: "LLVMFuzzerTestOneInput".} =
+      var payload {.inject.} : seq[byte]
```
zah: You can make `payload` a template that will use `stew/ranges/ptr_arith/makeOpenArray` to expose the input bytes to Nim as `openarray[byte]` without allocating memory for a sequence. This will buy you a little bit of fuzzing performance.

kdeme: oh thanks! I had briefly tried to access directly via an `openarray` but couldn't get it to work (hence the TODO). I'll give this a go.

tests/fuzzing/fuzz_helpers.nim
```diff
@@ -0,0 +1,101 @@
...
+    cleanStart = false) =
+  if not dirExists(inputDir):
+    # create a input dir with one 0 file for afl
+    mkDir(inputDir)
+    withDir inputDir: exec "echo '0' > test"
+
+  var fuzzCmd: string
+  # if there is an output dir already, continue fuzzing from previous run
+  if (not dirExists(resultsDir)) or cleanStart:
+    fuzzCmd = &"afl-fuzz -i {inputDir} -o {resultsDir} -M fuzzer01 -- ./{target}"
```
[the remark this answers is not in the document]
kdeme: Yes, you're right, its bad. Normally I would use `quoteShell` here but that's in `os` and not usable for nimscript. (although the nims docs page says it should be)
Good tip on the triple quotes, I'll go for this for now.

tests/fuzzing/fuzz_helpers.nim
```diff
@@ -0,0 +1,101 @@
...
+    libFuzzer
+
+  Compiler* = enum
+    gcc = aflGcc,
+    clang = aflClang,
+    clangFast = aflClangFast
+
+proc aflCompile*(target: string, c: Compiler) =
+  let aflOptions = &"-d:standalone -d:noSignalHandler {$c}"
+  let compileCmd = &"""nim c {defaultFlags} {aflOptions} {target}"""
```
zah: You still have to put quotes around the `target` name here and various directory names below (to solve the problem of having a potential space in the path).

kdeme: woops, indeed. Decided to just copy the quoteShell code as my eyes started burning from all the quotes.

tests/fuzzing/fuzz_helpers.nim
```diff
@@ -42,75 +42,54 @@ type
...
+            path & DirSep & "input",
+            path & DirSep & "results")
 
   of libFuzzer:
     libFuzzerCompile(targetPath)
     # Note: Lets not mix afl input with libFuzzer corpus default. This can have
     # consequences on speed for afl. Better to look into merging afl results &
     # libFuzzer corpus.
-    libFuzzerExec(target, path & "/corpus")
+    libFuzzerExec(path & DirSep & target, path & DirSep & "corpus")
```
zah: `ospaths` defines an operator `/` that combines path with the right separator. It's used like this:

```nim
libFuzzerExec(path / target, path / "corpus")
```

PR #100, opened 2019-10-04
Beginning of bzz protocol + simple handshake tester

kdeme:
I've tried to take the important parts from https://github.com/oskarth/nim-eth/pull/2 and did some rework/clean-up on it.

Basically, it will allow for a bzz handshake on a node running ethersphere/swarm client to complete.
```
DEBUG[10-04|14:46:58.575|p2p/server.go:733]                                         Adding p2p peer                          name=nim-eth-p2p/0.2.0       addr=172.17.0.1:34150    peers=5
DEBUG[10-04|14:46:58.616|github.com/ethersphere/swarm/network/protocol.go:256]      peer created                             addr="0000000000000000000000000000000000000000000000000000000000000000 <enode://1dfcee808c95deecba3ef94d933e1776592d2cfd78ee6a9fc3b82af4147462816ad574fd11dcb6259226955fea61cca7a697682933d0a4f751644c5dec2fed3f@127.0.0.1:30303>"
```
Of course on our side it needs a whole lot of more implementation.

### DIFF
--- a/eth/p2p/rlpx_protocols/bzz_protocol.nim
+++ b/eth/p2p/rlpx_protocols/bzz_protocol.nim
@@ -1,0 +1,85 @@
+import
+  chronos, chronicles, eth/p2p
+
+# Limited bzz protocol that allows for doing a handshake with a peer running
+# ethersphere/swarm client rev. c535b271536d0dee5bd97c2541ca32a42f272d4f
+
+const
+  bzzVersion = 12
+  hiveVersion = 10
+  swarmNetworkId* = 4
+  # Faking our capabilities to make handshake work, "bit" 0, 1, 4, 5, 15.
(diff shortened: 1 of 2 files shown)


zah:
The only concern I have is that we'll be doing false advertising that BZZ is supported. Other clients may try to communicate with us. This can be mitigated if we switch to a different model where the protocol capabilities are added explicitly depending on a command-line switch (Geth have similar switches to enable Whisper for example).

zah:
But since this module is not yet imported from anywhere, it won't end up in the supported protocols, so I'm merging this.

kdeme:
yeah, agreed. Should not be allowed to really use it together with other protocols in some application (like Nimbus), as it can easily cause disconnects from the peer.

PR #101, opened 2019-10-08
Fuzzing updates

kdeme:
- docs
- aflInit + aflLoop
- Fix 4x AssertionError + 2x segfault found when fuzzing on rlpx.

### DIFF
--- a/eth/p2p/rlpx.nim
+++ b/eth/p2p/rlpx.nim
@@ -23,7 +23,13 @@ type
 
   ResponderWithoutId*[MsgType] = distinct Peer
 
+  # We need these two types in rlpx/devp2p as no parameters or single parameters
+  # are not getting encoded in an rlp list.
+  # TODO: we could generalize this in the protocol dsl but it would need an
+  # `alwaysList` flag as not every protocol expects lists in these cases.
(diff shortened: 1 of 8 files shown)


eth/p2p/rlpx.nim
```diff
@@ -512,11 +519,17 @@ proc dispatchMessages*(peer: Peer) {.async.} =
     except PeerDisconnected:
       return
 
+    # TODO: is there a reason why we don't just allow `sendDisconnectMsg`
+    # to do its work here, with the usual thunk checks included?
```
zah: I think this code should be safe for removal indeed. Have you tried doing this?

kdeme: Yes, it works within our own tests.

However, it doesn't when running Nimbus (against other clients). This is because a rlp list with a number (reason) is expected according to https://github.com/ethereum/devp2p/blob/master/rlpx.md#disconnect-0x01 .
We just have `DisconnectionReason` there which doesn't create a list, just the number. The problem is similar to the ping/pong where I had to create that EmptyList for.
Quick test with `seq[DisconnectionReason]` instead seems to work as expected.

PR #104, opened 2019-10-17
WIP Add rlpx unit test concept

kdeme:
Sort of a PoC. The idea would be to change the templates so that we can reuse the code in thunk.nim currently used for fuzzing. 
Either by doing an include of the fuzzing application,or perhaps better, by running the same fuzzing application, but with another define which will then read and test the json file input.


kdeme:
Although the idea of having unit tests and being able to fuzz from those still appeals to me. This will require quite some more work to do properly so closing in favor of some simple test setup at https://github.com/status-im/nim-eth/pull/113

PR #112, opened 2019-10-22
add ingress mac on foo check in rlpx auth test

kdeme:


kdeme:
hmm, latest nimcrypto seems to break some things here? @cheatfate

cheatfate:
@kdeme fixed

Issue #109, opened 2019-10-21
Check rlpx for EIP-8 compliance


kdeme:
Check if rlpx complies with EIP-8 and create unit tests with these test vectors:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-8.md#test-vectors

This was done already for the discovery part.

cheatfate:
This two procedures are already supporting EIP-8 authentication and perform it transparently.
https://github.com/status-im/nim-eth/blob/master/eth/p2p/auth.nim#L436
https://github.com/status-im/nim-eth/blob/master/eth/p2p/auth.nim#L456

Test vectors from link you mentioned already in test suite.
https://github.com/status-im/nim-eth/blob/master/tests/p2p/test_auth.nim#L95-L96

kdeme:
Oh that's great! I should have known that you would have looked into that :)

I did add one little check that seems to be forgotten: https://github.com/status-im/nim-eth/pull/112

PR #113, opened 2019-10-22
Rlpx unittest

kdeme:
Basic rlpx thunk unittests, currently containing payloads of previously triggered crashes + the EIP-8 payload (for https://github.com/status-im/nim-eth/issues/109).


kdeme:
Also changed the msgId to int32 to be the same for x64 & x86 (test originally failed on x86). I don't think we need int64 as I think int32 should be big enough, but we could always change to int64.

PR #114, opened 2019-10-28
[Do not merge] Implement Waku mode PoC

kdeme:
Quick PoC of how the Waku mode could look. See https://github.com/status-im/specs/pull/54

### DIFF
--- a/eth/p2p/rlpx_protocols/whisper_protocol.nim
+++ b/eth/p2p/rlpx_protocols/whisper_protocol.nim
@@ -161,6 +161,13 @@ type
     bloom*: Bloom
     isLightNode*: bool
     maxMsgSize*: uint32
+    wakuMode*: WakuMode
+    topics*: seq[Topic]
+
+  WakuMode* = enum
+    None, # No Waku mode
+    WakuChan, # Waku client
+    WakuSan # Waku node
 
 # Utilities --------------------------------------------------------------------
 
@@ -533,6 +540,11 @@ proc allowed*(msg: Message, config: WhisperConfig): bool =
     warn "Message does not match node bloom filter"
     return false
 
+  if config.wakuMode == WakuChan:
+    if msg.env.topic notin config.topics:
+      warn "Message topic does not match Waku topic list"
+      return false
+
   return true
 
 # NOTE: Hashing and leading zeroes calculation is now the same between geth,
@@ -725,6 +737,8 @@ type
     bloom*: Bloom
     isLightNode*: bool
     trusted*: bool
+    wakuMode*: WakuMode
+    topics*: seq[Topic]
     received: HashSet[Message]
 
   WhisperNetwork = ref object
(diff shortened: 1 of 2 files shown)


kdeme:
This are currently purely the Whisper changes.

It would also require some forced setting to only connect with 1 other (trusted) waku node. And thus don't run additional discovery or listen for incoming connections.

Next to that, it would be best if we could have backwards compatibility with Whisper versions not having the Waku mode.

kdeme:
Another, more ugly, way making sure we have backwards compatibility would be to not put the `WakuMode` and `topics` in the handshake, but having an additional message type which requests the `WakuMode` of the peer. If it turns out the be an "old" node without Waku support it will (must) simply ignore this message. Thus no reply will be send and the node can disconnect from this peer.

adambabik:
We already have incompatible number of arguments in Whisper handshake with regards to EIP-627:
- https://github.com/status-im/whisper/blob/master/whisperv6/peer.go#L93
- https://github.com/ethereum/go-ethereum/blob/master/whisper/whisperv6/peer.go#L89

From our fork of Whisper, I assume that we can handle that gracefully, i.e. if there is no additional param, ignore the error (https://github.com/status-im/whisper/blob/master/whisperv6/peer.go#L141-L144). It needs to be confirmed whether EIP-627 compatible node will handle properly a handshake with more params but it looks like it will. 

In the Waku mode spec (https://github.com/status-im/specs/pull/54), there are two problems: (1) adding a new packet code, (2) identifying as a waku node.

In my opinion, identifying as a Waku Node might be put into the handshake. It is very close logically to identifying as a Light Node. Generally, this light node identification could be extend to be an enum informing about the Whisper node mode (light, waku, possibly others). Alternatively, we can rely on the peers discovery mechanism. A peer would seek for Whisper nodes advertising themselves as Waku nodes, do a handshake and finally verify whether the advertisement is correct using a special packet or by examining its interface.

Adding a new packet is already explored by our Whisper fork, for example, packets to do the sync between mailservers.

So, there is a possibility to do everything we want by just introducing new packets and utilizing peers discovery mechanism and avoid changing the handshake at all.

kdeme:
I did not know about this change in the handshake. Only knew about the added `packetid`s.

I agree on the identification in the handshake. The code change in this PR does that. It adds a `WakuMode` enum. It doesn't use the light node field as that would rather certain break things I believe. The list of topics is also immediately send in the handshake (similar like the bloom filter).

Having peer capability (supported protocols) checking in discovery would in general be better yes. But for us this is not yet possible. This would mean we need to implement discovery v5 or at least the [ENR mechanism](https://eips.ethereum.org/EIPS/eip-778).
Also, I'm not sure if that will resolve the handshake here. Ideally you still send which type of `WakuMode` you are (client or node) and which topics you want. So purely the protocols supported is not enough. Sure, you could add the WakuMode in the discovery information also, but you have to draw a line somewhere. Where does it end?

Adding `packetid`s is indeed never a problem. The node is supposed to ignore what it does not know.

kdeme:
closed in favour of https://github.com/status-im/nim-eth/pull/147

PR #118, opened 2019-11-07
Do not start the peer pool loop when discovery is disabled

kdeme:
Doesn't make sense to have that peer_pool `run` running when there will never be peers presented as discovery is disabled.

### DIFF
--- a/eth/p2p.nim
+++ b/eth/p2p.nim
@@ -104,10 +104,9 @@ proc connectToNetwork*(node: EthereumNode,
   if enableDiscovery:
     node.discovery.open()
     await node.discovery.bootstrap()
+    node.peerPool.start()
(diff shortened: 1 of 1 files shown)


zah:
Woudn't the `PeerPool` be updated from manually established connections as well?
I think the only reason to run without discovery is when you intend to establish connections manually.

kdeme:
Yes, but that is still possible. Connecting manually to predefined peers will still add them to the `PeerPool`. 

This loop that I disable here seems to just be trying to connect to `randomNodes` it gets from discovery. But it will never get any if discovery is disabled. So unless I'm missing here something, that loop runs for no reason.

zah:
OK, thanks for the explanation.

Issue #122, opened 2019-11-18
Refactor whisper message structure + protocol code


kdeme:
Split current Whisper code into message structure and purely protocol code.
This avoids duplicated code & tests for other versions/forks of the protocol such as the one Status uses (soon Waku/0)

PR #128, opened 2019-11-19
Split off whisper config related tests

kdeme:
Move `WhisperConfig` related tests to another file. This way the original test only has to import `whisper_types` and several waku tests don't have to be duplicated in the future.

see issue: https://github.com/status-im/nim-eth/issues/122


kdeme:
I will go ahead and merge this as there are no functional changes.

PR #129, opened 2019-11-20
Support for compressed public key format

yglukhov:
Required by discv5

### DIFF
--- a/eth/keys/libsecp256k1.nim
+++ b/eth/keys/libsecp256k1.nim
@@ -12,6 +12,7 @@ import secp256k1, nimcrypto/sysrand, nimcrypto/utils
 
 const
   KeyLength* = 256 div 8
+  CompressedPubKeyLength* = 33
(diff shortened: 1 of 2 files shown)


eth/keys/libsecp256k1.nim
```diff
@@ -154,16 +155,24 @@ proc recoverPublicKey*(data: openarray[byte],
   ## Unserialize public key from `data`.
   let ctx = getSecpContext()
   let length = len(data)
-  if length < RawPublicKeySize:
+  if length >= RawPublicKeySize:
```
zah: How come the conditional here was changed, but we didn't break any tests? Was this function lacking coverage before? Does it cover the uncompressed case after this PR?

yglukhov: It is covered. I've changed the branch ordering slightly, along with the conditions :)

PR #134, opened 2019-11-28
support trie with multiple root if pruning turned on

jangko:
use case:

sync start from block 4, 5, 6 ,7, 8, oops it's a wrong chain.

resync from block 4, 5b, 6b, 7b, 8b.

if trie pruning turned on, the state trie at block 4 already damaged by wrong chain.

this PR combined with nimbus upcoming PR will make trie pruning less agressive.


jangko:
can we merge this one, @zah?

PR #135, opened 2019-12-02
Replace getCurrentException and getCurrentExceptionMsg

kdeme:
Similar as for the other repos, change to stop using `getCurrentException`. From going over them, didn't actually notice any that could create an issue.

### DIFF
--- a/eth/net/nat.nim
+++ b/eth/net/nat.nim
@@ -71,8 +71,8 @@ proc getExternalIP*(natStrategy: NatStrategy, quiet = false): Option[IpAddress] 
             externalIP = parseIpAddress(ires.value)
             strategy = NatUpnp
             return some(externalIP)
-          except:
-            error "parseIpAddress() exception", err = getCurrentExceptionMsg()
+          except ValueError as e:
+            error "parseIpAddress() exception", err = e.msg
             return
 
   if natStrategy == NatAny or natStrategy == NatPmp:
@@ -89,8 +89,8 @@ proc getExternalIP*(natStrategy: NatStrategy, quiet = false): Option[IpAddress] 
           externalIP = parseIpAddress($(nires.value))
           strategy = NatPmp
           return some(externalIP)
-        except:
-          error "parseIpAddress() exception", err = getCurrentExceptionMsg()
+        except ValueError as e:
+          error "parseIpAddress() exception", err = e.msg
           return
 
 proc doPortMapping(tcpPort, udpPort: Port, description: string): Option[(Port, Port)] {.gcsafe.} =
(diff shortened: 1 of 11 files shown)


eth/net/nat.nim
```diff
@@ -71,8 +71,8 @@ proc getExternalIP*(natStrategy: NatStrategy, quiet = false): Option[IpAddress]
             externalIP = parseIpAddress(ires.value)
             strategy = NatUpnp
             return some(externalIP)
-          except:
-            error "parseIpAddress() exception", err = getCurrentExceptionMsg()
+          except ValueError as e:
```
kdeme: For reference, this is the only error that can be raised currently.
Could change it though to `CatchableError` if we don't trust this not to change as we have no raises pragma to guard it.

eth/p2p/discovery.nim
```diff
@@ -263,10 +263,12 @@ proc processClient(transp: DatagramTransport,
...
     proto.receive(a, buf)
-  except RlpError, DiscProtocolError:
-    debug "Receive failed", err = getCurrentExceptionMsg()
-  except:
-    debug "Receive failed", err = getCurrentExceptionMsg()
+  except RlpError as e:
+    debug "Receive failed", exc = e.name, err = e.msg
+  except DiscProtocolError as e:
+    debug "Receive failed", exc = e.name, err = e.msg
+  except Exception as e:
```
stefantalpalaru: Don't you want CatchableError here?

eth/p2p/peer_pool.nim
```diff
@@ -173,8 +173,7 @@ proc run(p: PeerPool) {.async.} =
       # Most unexpected errors should be transient, so we log and restart from
```
stefantalpalaru: The exception above should probably be CatchableError.

eth/p2p/rlpx.nim
```diff
@@ -443,11 +443,11 @@ proc checkedRlpRead(peer: Peer, r: var Rlp, MsgType: type): auto {.inline.} =
   else:
     try:
       return r.read(MsgType)
-    except:
+    except Exception as e:
```
stefantalpalaru: Guess what?

eth/p2p/discovery.nim
kdeme: It's being reraised, so no, not necessarily. I guess it is open for debate (see libp2p fix discussion).
Anyway, In this case it makes sense to just remove it.

eth/p2p/peer_pool.nim
kdeme: True, I'll change it.

eth/p2p/rlpx.nim
kdeme: Same story, here it is used to add some extra useful logging before the application dies. Is this only useful for CatchableErrors and not for Defects? I'm not so sure, I remember hitting quite some AssertionsErrors and other Defects when fuzzing rlpx and rlp.

Issue #131, opened 2019-11-21
Allow Whisper light client to send regular messages?


kdeme:
Decide on whether we want to change our Whisper implementation to allow this.
More information here: https://github.com/vacp2p/specs/pull/28#discussion_r348976883

For Waku/0 we will most likely have to change this.

kdeme:
Additionally, if we allow this for Waku/0 protocol, we could make it so that the `processQueue` doesn't need to run at all and that the messages are send synchronous with e.g. the `postMessage` call (as is done now for the `sendP2PMessage`)

kdeme:
Done for Waku.

PR #137, opened 2019-12-03
Do not send empty messages packets at interval

kdeme:
While it is allowed to send `messages` packets with empty envelopes list, there is no real reason to do so. So lets disable it to have less drain of battery.

While it might occur for full nodes (probably not often) and probably more for nodes with narrow bloomfilter set, it will be extremely important for light nodes, if they are allowed to send envelopes via the `messages` packet, see also https://github.com/status-im/nim-eth/issues/131

### DIFF
--- a/eth/p2p/rlpx_protocols/waku_protocol.nim
+++ b/eth/p2p/rlpx_protocols/waku_protocol.nim
@@ -270,10 +270,11 @@ proc processQueue(peer: Peer) =
     envelopes.add(message.env)
     wakuPeer.received.incl(message)
 
-  trace "Sending envelopes", amount=envelopes.len
-  # Ignore failure of sending messages, this could occur when the connection
-  # gets dropped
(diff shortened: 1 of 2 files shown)


arnetheduck:
not sure if it's relevant to this pr, but generally whisper has the notion of cover traffic..

kdeme:
Yes. But in this case we are potentially just sending empty rlp lists. If someone is watching the nodes traffic, they can just notice this so I believe there is no difference in not sending the packet at all. So I don't think it is relevant here.

What would be an improvement of "cover traffic" is when we are posting our own envelope from a node, but the moment the Interval hits, there are no new envelopes except our own one. In this case we should probably wait (until next interval) to see if a certain threshold of new envelopes is hit and only then send the envelopes. Or we could inject ourselves extra garbage traffic if there really is not enough regular traffic. 

Although, come to think of it, if that wait is rare (like only at posts), then this will again leak information.

PR #141, opened 2019-12-05
Waku mail client implementation

kdeme:
Starting on mail client as in https://github.com/vacp2p/specs/blob/master/waku.md#mailserver-and-client .

Some parts are unclear:
- cursor -> This means it has to be send back, through the same packet id? Why not some other  packet id (a reponse)? And why the array?
- unclear which validation checking of envelope is done in the P2P Request packet. All like normal messages?
- unclear which validation checking is done on envelopes for P2P Message packet. TTL for sure is skipped, but anything else?

### DIFF
--- a/eth/p2p/rlpx_protocols/waku_mail.nim
+++ b/eth/p2p/rlpx_protocols/waku_mail.nim
@@ -32,6 +32,9 @@ proc requestMail*(node: EthereumNode, peerId: NodeId, request: MailRequest,
   # TODO: Perhaps don't go the recursive route or could use the actual response
   # proc to implement this (via a handler) and store the necessary data in the
   # WakuPeer object.
+  # TODO: Several requestMail calls in parallel can create issues with handling
+  # the wrong response to a request. Can additionaly check the requestId but
+  # that would only solve it half. Better to use the requestResponse mechanism.
 
   # TODO: move this check out of requestMail?
   let peer = node.getPeer(peerId, Waku)
--- a/eth/p2p/rlpx_protocols/waku_mail.nim
+++ b/eth/p2p/rlpx_protocols/waku_mail.nim
@@ -1,0 +1,82 @@
+#
+#            Waku Mail Client & Server
+#              (c) Copyright 2019
+#       Status Research & Development GmbH
+#
+#            Licensed under either of
+#  Apache License, version 2.0, (LICENSE-APACHEv2)
+#            MIT license (LICENSE-MIT)
+#
+import
+  chronos,
+  eth/[p2p, async_utils], eth/p2p/rlpx_protocols/waku_protocol
+
+const
+  requestCompleteTimeout = chronos.seconds(5)
+
+type
+  Cursor = Bytes
+
+  MailRequest* = object
+    lower*: uint32 ## Unix timestamp; oldest requested envelope's creation time
+    upper*: uint32 ## Unix timestamp; newest requested envelope's creation time
+    bloom*: Bytes ## Bloom filter to apply on the envelopes
+    limit*: uint32 ## Maximum amount of envelopes to return
+    cursor*: Cursor ## Optional cursor
+
+proc requestMail*(node: EthereumNode, peerId: NodeId, request: MailRequest,
+    symKey: SymKey, requests = 10): Future[Option[Cursor]] {.async.} =
+  ## Send p2p mail request and check request complete.
+  ## If result is none, and error occured. If result is a none empty cursor,
+  ## more envelopes are available.
+  # TODO: Perhaps don't go the recursive route or could use the actual response
+  # proc to implement this (via a handler) and store the necessary data in the
+  # WakuPeer object.
+
+  # TODO: move this check out of requestMail?
+  let peer = node.getPeer(peerId, Waku)
+  if not peer.isSome():
+    error "Invalid peer"
+    return result
+  elif not peer.get().state(Waku).trusted:
+    return result
+
+  var writer = initRlpWriter()
+  writer.append(request)
+  let payload = writer.finish()
+  let data = encode(Payload(payload: payload, symKey: some(symKey)))
+  if not data.isSome():
+    error "Encoding of payload failed"
+    return result
+
+  # TODO: should this envelope be valid in terms of ttl, PoW, etc.?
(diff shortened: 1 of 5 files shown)


oskarth:
Answers from my understanding, but @cammellos and @adambabik probably know better:


>     cursor -> This means it has to be send back, through the same packet id? Why not some other packet id (a reponse)? And why the array?

Yes, same packet id. Another packet id might make sense, I think in general the API should probably be fixed amended (past v0).

>    unclear which validation checking of envelope is done in the P2P Request packet. All like normal messages?

Believe so.

 >   unclear which validation checking is done on envelopes for P2P Message packet. TTL for sure is skipped, but anything else?

Think that's it

adambabik:
> cursor -> This means it has to be send back, through the same packet id? Why not some other packet id (a reponse)? And why the array?

No, it's different packet code. The documentation misses a fragment about a packet 125 which is [implemented in our for of Whisper](https://github.com/status-im/whisper/blob/master/whisperv6/doc.go#L60). 

I will take care of this today. [Issue](https://github.com/status-im/specs/issues/71).

adambabik:
[The current spec](https://github.com/status-im/specs/issues/71#issuecomment-562528754) explains how to handle this additional packet with the response.

kdeme:
@adambabik 
> 
> [Here is how a regular envelope is validated](https://github.com/status-im/whisper/blob/master/whisperv6/whisper.go#L1213) vs [how a p2p envelope is validated](https://github.com/status-im/whisper/blob/25321b2c035b6e03dbae85a2f54cf89f9f873dd9/whisperv6/whisper.go#L1303:25). p2p messages pretty much are not validated at all because they are assumed that they were already validated by mailserver.

Yes, the first link is for regular message adding to the queue.
However, I don't think the P2P message request goes through that process. 

I'm actually quite confused as there seem to be two implementations for this:
https://github.com/status-im/whisper/blob/25321b2c035b6e03dbae85a2f54cf89f9f873dd9/whisperv6/whisper.go#L449 & https://github.com/status-im/whisper/blob/25321b2c035b6e03dbae85a2f54cf89f9f873dd9/whisperv6/whisper.go#L465
One with an envelope, the other not?

From looking at the server counterpart, I have the impression it just checks the PoW value: https://github.com/status-im/status-go/blob/a636f33109130a9151553abf5e0253e35994ef60/mailserver/mailserver.go#L798 

For the actual message envelopes send via P2P Message, it makes sense that validation is skipped.

kdeme:
If OK, this can be merged and I fill in the TODO gaps in a second part.

eth/p2p/rlpx_protocols/waku_mail.nim
```diff
@@ -0,0 +1,82 @@
...
+    error "Encoding of payload failed"
+    return result
+
+  # TODO: should this envelope be valid in terms of ttl, PoW, etc.?
+  let env = Envelope(expiry:0, ttl: 0, data: data.get(), nonce: 0)
+  # Send the request
+  traceAsyncErrors peer.get().p2pRequest(env)
+
+  # Wait for the Request Complete packet
+  var f = peer.get().nextMsg(Waku.p2pRequestComplete)
```
zah: If you have multiple concurrent requests, `nextMsg(Waku.p2pRequestComplete)` may be completed by a response from one of your other requests. Is this OK for the code here?

zah: The solution would be to either add request IDs to the Waku messages or to implement some sort of queuing that will guarantee that only a single request can be in flight.

eth/p2p/rlpx_protocols/waku_protocol.nim
```diff
@@ -71,11 +71,22 @@ type
...
     filters*: Filters
     config*: WakuConfig
+    p2pRequestHandler*: P2PRequestHandler
 
+  # TODO: In the current specification this is not wrapped in a regular envelope
+  # as is done for the P2P Request packet. If we could alter this in the spec it
+  # would be a cleaner separation between Waku and Mail server / client and then
+  # this could be moved to waku_mail.nim
+  P2PRequestCompleteObject* = object
+    requestId*: Hash
```
zah: Does this need to be a hash? The framework has support for numeric request IDs through the `requestResponse: ` helpers in the p2pProtocol DSL.

eth/p2p/rlpx_protocols/waku_protocol.nim
```diff
@@ -347,7 +363,7 @@ proc postMessage*(node: EthereumNode, pubKey = none[PublicKey](),
 
     # Allow lightnode to post only direct p2p messages
     if targetPeer.isSome():
-      return node.sendP2PMessage(targetPeer.get(), env)
+      return node.sendP2PMessage(targetPeer.get(), @[env])
```
zah: It would be more efficient to use an array here: `[env]`.
This would avoid allocating heap memory for the sequence.

eth/p2p/rlpx_protocols/waku_mail.nim
kdeme: True, didn't think of that but this would be a problem when `requestMail` is called multiple times in parallel. 
Could use the `requestId` (see TODO lower) to check for that, but it would still stop the flow.

eth/p2p/rlpx_protocols/waku_protocol.nim
kdeme: It doesn't need to be, it is just that this is how it is specified now in the Status mail server specification.
Ideally, for Waku, we can change this to what is more convenient.

zah: I suggest following the practice of the LES protocol:
https://github.com/ethereum/devp2p/blob/master/caps/les.md

In our codebase, this looks like this:
https://github.com/status-im/nim-eth/blob/master/eth/p2p/rlpx_protocols/les_protocol.nim#L281

On the usage call-sites, this will automatically wait for the correct response.

kdeme: So yes, ideally we could use the request IDs provided by the requestResponse mechanism. It was my original plan to use that mechanism, however it would not work due to non consecutive packet ID's, see comment lower.

PR #144, opened 2019-12-10
Enr parsing and construction/serialization

yglukhov:
Note, this PR also includes the necessary additions to `eth/keys`:
 - Added `SignatureNR` type for non-recoverable signature. I'd prefer using `Signature` and `RecoverableSignature` instead, but unfortunately `Signature` name is already taken.
 - Added `parseCompact` functions for `Signature` and `SignatureNR`. For signature it kinda does what `recoverSignature` did, so I decided to deprecate `recoverSignature` function because it is now confusing in the context of recoverable/nonrecoverable signatures.


cheatfate:
If base64 encoding/decoding will be used not only in tests i suggest not to use stdlib's base64 version, but https://github.com/status-im/nim-libp2p/blob/master/libp2p/base64.nim

tests/p2p/test_enr.nim
```diff
@@ -0,0 +1,19 @@
...
+    var pk = initPrivateKey("5d2908f3f09ea1ff2e327c3f623159639b00af406e9009de5fd4b910fc34049d")
+    var r = initRecord(123, pk, {"udp": 1234, "ip": 12345})
+    doAssert($r == """(id: "v4", ip: 12345, secp256k1: 0x02E51EFA66628CE09F689BC2B82F165A75A9DDECBB6A804BE15AC3FDF41F3B34E7, udp: 1234)""")
+    let uri = r.toURI()
+    var r2: Record
+    let sigValid = r2.fromURI(uri)
+    doAssert(sigValid)
+    doAssert($r2 == $r)
+
+  test "Parsing":
```
cheatfate: Is there only one test vector available? What about malicious input?

eth/keys/libsecp256k1.nim
```diff
@@ -274,6 +303,22 @@ proc toRaw*(s: Signature, data: var openarray[byte]) =
     raiseSecp256k1Error()
   data[64] = uint8(recid)
 
+proc getRaw*(s: Signature): array[RawSignatureSize, byte] {.noinit.} =
+  ## Converts signature `s` to serialized form.
+  s.toRaw(result)
+
+proc toRaw*(s: SignatureNR, data: var openarray[byte]) =
+  ## Converts signature `s` to serialized form and store it in `data`.
+  doAssert(len(data) == RawSignatureNRSize)
```
cheatfate: I think there no need to have `==` here, `>=` will also work.

eth/keys/libsecp256k1.nim
```diff
@@ -175,12 +184,23 @@ proc recoverPublicKey*(data: openarray[byte],
...
+    setErrorMsg(InvalidSignature)
+    return(EthKeysStatus.Error)
+  result = EthKeysStatus.Success
+
+proc parseCompact*(signature: var Signature, data: openarray[byte]): EthKeysStatus =
   ## Unserialize signature from `data`.
   let ctx = getSecpContext()
   let length = len(data)
-  if length < RawSignatureSize:
+  if length != RawSignatureSize:
```
cheatfate: Why so strict comparison used?

yglukhov:
@cheatfate yes, `libp2p/base64` is nice, but `nim-eth` doesn't depend on `libp2p` currently. But I suppose it will soon? @zah?

cheatfate:
I think we can move base64, base58, base32 to nim-stew.

yglukhov:
Alright, but libp2p doesn't depend on stew currently :)

yglukhov:
@cheatfate regarding the strict length checks. It just serves better correctness and semantics. Usually functions taking arrays utilize all of the range, not the "first K elements". As I recall the length checks were relaxed back when we didn't have `toOpenArray` but now we do, so the user can pass whatever part of his data without performance implications.

tests/p2p/test_enr.nim
yglukhov: > Is there only one test vector available?

The "Parsing" test vector is the only one from the spec, yes.
>  What about malicious input?

Good point, I'll add a couple more tests.

eth/keys/libsecp256k1.nim
```diff
@@ -212,6 +240,16 @@ proc initSignature*(data: openarray[byte]): Signature =
   if recoverSignature(data, result) != EthKeysStatus.Success:
     raise newException(EthKeysException, libsecp256k1ErrorMsg())
 
+proc ecdhAgree*(seckey: PrivateKey, pubkey: PublicKey,
+                    secret: var SharedSecretFull): EthKeysStatus =
```
arnetheduck: generally, this code would be a lot more ergonomic to use with `Result[ShardSecretFull, EthKeysStatus]`: easier to call, no partially/uninitialized `secret` on fail, easy to convert to exception with result `[]` operator etc

cheatfate:
@yglukhov sorry but i still can't understand why this strict size comparison is needed. You have enough data to proceed correctly without any implications.

Also how i do not understand how it helps to stay correct?

```nim
let res = parseCompact(signature, dataBuffer.toOpenArray(offset, len(dataBuffer) - 1))
```
vs
```nim
let res = parseCompact(signature, dataBuffer.toOpenArray(offset, offset + signatureRawSize - 1))
```
If `dataBuffer` do not have enough bytes, in first case res will have `InvalidSignature` error, but in second case you will have `IndexError` exception.

`IndexError` looks like `Defect` while `InvalidSignature` looks like `CatchableError` here.

arnetheduck:
er, I hope we're not raising `Defect` on parsing??

arnetheduck:
+1 to make libp2p depend on `stew` and move encodings there

eth/p2p/discoveryv5/enr.nim
```diff
@@ -0,0 +1,223 @@
...
+    case k
+    of "id":
+      id = rlp.read(string)
+      r.pairs.add((k, Field(kind: kString, str: id)))
+    of "secp256k1":
+      pubkeyData = rlp.read(seq[byte])
+      r.pairs.add((k, Field(kind: kBytes, bytes: pubkeyData)))
+    of "tcp", "udp", "tcp6", "udp6", "ip":
+      let v = rlp.read(int)
+      r.pairs.add((k, Field(kind: kNum, num: v)))
```
zah: Why don't we map the pre-defined keys to strongly typed fields.
It's a bit inconvenient to read the values through the dynamically typed `getField` API.

zah: Why don't we map the pre-defined keys to strongly typed fields?
It's a bit inconvenient to read the values through the dynamically typed `getField` API.

yglukhov: No strong preference, just tried to somewhat avoid duplicating data. But we could add typed accessors (wrappers around getField) for common fields.

yglukhov:
> If dataBuffer do not have enough bytes, in first case res will have InvalidSignature error, but in second case you will have IndexError exception.

Correct, but I would argue that `InvalidSignature` is a wrong error in this case, because it's not the signature which is invalid, but the buffer doesn't even have the signature in it. Whereas `IndexError` would obviously indicate a bug, that we don't want to hide. And indeed in your code you're missing a validation step.

Regardless the length check, both APIs imply that the user **must** know the signature length and **must** validate sufficient buffer length before parsing. In your case technically we could try parse, get the error, and only then check the length, but it's not how it is done in real world scenario, neither we want to encourage anyone to do that.

Here's a more real world scenario:
```nim
if myMessage.len >= MIN_MESSAGE_LE:
  let res = parseCompact(sig, dataBuffer.toOpenArray(offset, len(dataBuffer) - 1))
  # vs
  let res = parseCompact(sig, dataBuffer.toOpenArray(offset, offset + signatureRawSize - 1))
  let packet = decodePacket(dataBuffer.toOpenArray(offset + signatureRawSize, dataBuffer.len - 1))
```
Note how `decodePacket` requires us to use `signatureRawSize`. So we don't gain a lot more simplicity with your approach.

arnetheduck:
> And indeed in your code you're missing a validation step.

should the calling code really have to know the detailed required length? I mean, if it has to check for it every time anyway, why not let the parser deal with it?

yglukhov:
Well the other way would be to feed an array to the parser and make the parser return the "consumed" length, is that what you mean, @arnetheduck? That would be correct api-wise and kinda allows the user not to know the length in advance, but since the length is constant, knowing it allows for arguably simpler and potentially more optimized user code. Either way, the user must know the length of the blob to parse. Whether it is defined as a constant, or returned by `parse` proc.

eth/p2p/discoveryv5/enr.nim
```diff
@@ -0,0 +1,225 @@
...
+    of "v4":
+      result = verifySignatureV4(r, sigData, content)
+    else:
+      # Unknown Identity Scheme
+      discard
+
+proc fromBytesAux(r: var Record): bool =
+  var rlp = rlpFromBytes(r.raw.toRange)
+  let sz = rlp.listLen
+  if sz < 5 or sz mod 2 != 0:
```
kdeme: Could you add a comment why the size < 5? 
In the [spec](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-778.md#specification) it says `All keys except id are optional`. So that would make 4 also valid?
Although I guess when the v4 scheme is used then 4 is not valid as it can't be verified without the secp256k1 id.

eth/p2p/discoveryv5/enr.nim
```diff
@@ -0,0 +1,225 @@
...
+    # Wrong rlp object
+    return false
+
```
kdeme: From spec: `the maximum encoded size of a node record is 300 bytes. Implementations should reject records larger than this size.` 
Should we also not add this total size check somewhere here or do you think it should be checked at a layer lower?
I think it is especially important not to forget considering that the `verifySignature` is done at end, after all key/value pairs are read.

zah:
Closing in favor of https://github.com/status-im/nim-eth/pull/151

eth/keys/libsecp256k1.nim
yglukhov: @kdeme, yes, this is a new function :)

eth/keys/libsecp256k1.nim
```diff
@@ -190,6 +210,14 @@ proc recoverSignature*(data: openarray[byte],
     return(EthKeysStatus.Error)
   result = EthKeysStatus.Success
 
+proc recoverSignature*(data: openarray[byte],
+                       signature: var Signature): EthKeysStatus {.deprecated.} =
+  ## Deprecated, use `parseCompact` instead
+  if data.len < RawSignatureSize:
```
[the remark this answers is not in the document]
yglukhov: Yes. Thought it would be fine for a deprecated proc

eth/p2p/discoveryv5/enr.nim
yglukhov: Good point. The required list elements are signature, seq, "id" key, id value. Then depending on how you count an enr without a pubkey is kinda invalid for us, but tbh I'm not sure why I wrote 5 here, instead of 4 or 6 :).

eth/p2p/discoveryv5/enr.nim
yglukhov: I must have missed it from the spec, will add a check.

eth/keys/libsecp256k1.nim
kdeme: A right, I missed the `toOpenArray` part where the openArray ranges are changed. My bad.

PR #146, opened 2019-12-11
implement hexary trie replicate iterator

jangko:
this iterator helps 'rebuild' the entire trie without going through a trie algorithm, but it will pull the entire low level KV pairs. Thus the target db will only use `put` operations without `del` or `contains`, can speed up huge trie replication.

this iterator probably will not be used directly in nimbus, but I need it for salvaging and cleaning my staging db and move the content to a new db.

using upper level KV trie iterator wasting too much time and db operations.

### DIFF
--- a/eth/trie/hexary.nim
+++ b/eth/trie/hexary.nim
@@ -270,6 +270,50 @@ iterator pairs*(self: HexaryTrie): (BytesRange, BytesRange) =
     let res = getPairsAux(self.db, stack)
     yield res
 
+iterator replicate*(self: HexaryTrie): (BytesRange, BytesRange) =
+  # this iterator helps 'rebuild' the entire trie without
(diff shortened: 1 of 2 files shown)


eth/trie/hexary.nim
```diff
@@ -270,6 +270,50 @@ iterator pairs*(self: HexaryTrie): (BytesRange, BytesRange) =
     let res = getPairsAux(self.db, stack)
     yield res
 
+iterator replicate*(self: HexaryTrie): (BytesRange, BytesRange) =
```
zah: This is fine for merging, but you should be aware that an inline iterator produces significant code bloat when there is more than one `yield` present inside it. The reason for this is that the for loop body that uses the iterator will be inlined in place of every `yield` statement. For this reason, you should strive to keep the number of `yields` to the minimum.

PR #151, opened 2019-12-16
Discv5

yglukhov:
Includes #144 
Node discovery works, but needs a bit more polish, and tests, which is in progress.


eth/p2p/discoveryv5/enr.nim
```diff
@@ -10,6 +10,16 @@ type
 
   EnrUri* = distinct string
 
+  TypedRecord* = object
```
zah: @yglukhov, in the previous PR, I've suggested using a more strongly-typed ENR records. What I meant is that we can deserialize the RLP value directly into an object type like the one here. If you think about it, it would actually use less memory than the `seq` based solution most of the time.

For me, the main motivation was the highly conditional code that I had to write in order to consume the information in the ENR in order to feed it into a LibP2P MultiAddress:

https://github.com/status-im/nim-beacon-chain/pull/639/files#diff-375c1c836a137011a0ce0c56d40d208eR153-R174

eth/p2p/discoveryv5/enr.nim
```diff
@@ -0,0 +1,283 @@
...
+    # signature: seq[byte]
+    raw*: seq[byte] # RLP encoded record
+    pairs: seq[(string, Field)] # sorted list of all key/value pairs
+
+  EnrUri* = distinct string
+
+  TypedRecord* = object
+    id*: string
+    secp256k1*: Option[array[33, byte]]
+    ip*: Option[array[4, byte]]
```
zah: I was a bit confused about the spec here. It says that the `ip` value is 4 bytes and the `ip6` value is 16 bytes, but is only one of them a byte blob prefixed with a length? I can see that you've passed the only official test vector by treating the `ip` field as a numeric value?

eth/p2p/discoveryv5/enr.nim
yglukhov: Alright, I can see a case where it would be easier to work with a typed record, but your particular impl of TypedRecord is not serializable, as it doesn't carry the signature and the custom keys around.

zah: Yes, the requirement for the keys to be sorted on serialization was a PITA to consider with this approach. My thinking now is that perhaps we should create a short-cut for deserializing a typed record that just ignores any other fields. After all, in our networking code we'll do ENR deserialization only with the purpose of obtaining the specific fields that we need in order to connect to another peer.

eth/p2p/discoveryv5/enr.nim
yglukhov: Yes, I too find the spec a bit vague on this. Apparently ipv4 is encoded as a big-endian int. ipv6 - not sure, but i'd expect it to be the same, i.e. int128.

eth/p2p/discoveryv5/enr.nim
yglukhov: But why don't you want to provide just the accessors to the `Record` that will lookup the field in the pairs? That way you would keep less types, `Record` would be both "typed" and serializable.

zah: Defining all the accessors that would use a linear search on each access seemed a bit awkward, but I guess this shouldn't matter much in practice considering how processing the ENR data is a relatively rare one-off event.

zah: If you would like to go with the accessors, I don't mind that. As far as I can tell, the code in the beacon node will need a 1-line change.

eth/p2p/discoveryv5/enr.nim
cheatfate: Usually ipv6 is stored like array of bytes.

PR #152, opened 2019-12-17
Fix PoW to be specifically uint64 as it is a float64 value that is be…

kdeme:
…ing passed


PombeirP:
@kim, tested this on Android x86 emulator and it worked.

Issue #156, opened 2020-01-20
Wakusim revealed duplicate valid_envelopes for node sending traffic


kdeme:
Testing simulation that is done here https://github.com/status-im/nimbus/pull/437 revealed that the traffic sending node has more `valid_envelopes` than there are actually being send. Investigate how these duplicates got added.

PR #158, opened 2020-01-23
Waku fixes

kdeme:
issue #156  was occurring due len field not being the same for a message posted. Original intent however was to just store the hashes, seems I forgot about changing that.


zah:
I still feel you would have been better off without the code duplication. Perhaps this won't be the last time where you have to carefully carry out the same edits in two files.

kdeme:
Yeah, things could probably improved here by more refactoring. I did the first refactor to whisper_types.nim very quickly by just moving all non "protocol" related parts. But sure, more can be done here.

PR #162, opened 2020-02-04
Add StatusOptions to Waku Status packet

kdeme:
Use the association list status-options as in spec: https://github.com/vacp2p/specs/blob/master/waku/waku.md#packet-usage
Also remove anything WakuMode related as this was still leftover from the PoC when compatibility with Whisper was still a thing.


eth/p2p/rlpx_protocols/waku_protocol.nim
```diff
@@ -116,6 +104,89 @@ type
...
+    except RlpTypeMismatch:
+      # skip unknown keys and their value
+      rlp.skipElem()
+      rlp.skipElem()
+      continue
+
+    case k
+    of powRequirementKey:
+      let pow = rlp.read(uint64)
+      result.powRequirement = some(cast[float64](pow))
```
zah: https://stackoverflow.com/questions/9056373/what-is-the-difference-between-big-and-little-endian-floats

This source suggests that the binary representation of floats may vary across machines. The Waku protocol should specify what encoding is being used and the source code should probably feature some test cases with sanity checks that the expected bits are being produced.

eth/p2p/rlpx_protocols/waku_protocol.nim
```diff
@@ -175,49 +245,37 @@ p2pProtocol Waku(version = wakuVersion,
...
+    let list = StatusOptions(
+      powRequirement: some(wakuNet.config.powRequirement),
+      bloomFilter: some(wakuNet.config.bloom),
+      lightNode: some(wakuNet.config.isLightNode),
+      confirmationsEnabled: some(wakuNet.config.confirmationsEnabled),
+      rateLimits: some(wakuNet.config.rateLimits),
+      topicInterest: wakuNet.config.topics)
+
+    let m = await peer.status(wakuVersion, list,
+      timeout = chronos.milliseconds(500))
```
zah: Isn't this timeout rather low for internet usage?

kdeme: Yes, increased now.

eth/p2p/rlpx_protocols/waku_protocol.nim
kdeme: Good point, we'll have to add this to the spec + and provide a proper call for it per endianness.

PR #163, opened 2020-02-05
Discv5 integration progress

zah:


eth/trie/trie_defs.nim
```diff
@@ -8,6 +8,18 @@ type
...
   BytesContainer* = ByteRange | Bytes | string
 
+  TrieError* = object of CatchableError
+    # A common base type of all Trie errors.
+
+  PersistenceFailure* = object of TrieError
+    # The backing database of the trie was not able to carry out
+    # the storage or retrieval request.
+
+  CorruptedTrieDatabase* = object of Defect
```
arnetheduck: I find it a bit unsettling that we build in expected crashes for "normal" user behaviour - corrupted databases is one of those things that happen out there and are not really internal programming errors - just like we don't expect data that arrives over the network or command line to be fully correct in all cases - it's part of "normal" program flow really

arnetheduck: ie we should expect that while the app is running, operations involving the database "work" - but between restarts, anything may happen and it becomes just another external data source whose consistency is questionable and should be handled deliberately

zah: My reasoning is explained in the comment just below this line.

> We consider this a Defect, because the software cannot safely
    operate if its database has been tampered with. A swift crash
    will be a more appropriate response.

This surely can be disputed, but your comments are not addressing the presented argument in any way.

arnetheduck: the point is more that these situations _can_ be handled (by validating db up front or passing a manageable error to the application layer - for example for discv5 one could simply drop whatever ENR's are not loading correctly etc), and we're just not doing it because the  surrounding code is not ready / hasn't been architected with this in mind.. 

crashing with a call stack would certainly not instil a sense of security and trust if I was running this software - it would instead tell me that it's indeed an omission on part of the developer .

zah: This change is not related to discv5 in any way. I've just used the opportunity to review all exception types in nim-eth, because I noticed a particular discv5-related one that was not right.

The release builds of Nimbus will surely handle crashes like this in a more user-friendly way. For example, the user can be presented with just the error string of the exception, while a script packages more details about the problem (e.g. core dump, executable version, OS details, etc), which can be then submitted automatically to a crash collection service maintained by us.

I agree that this is not a clear `Defect`. In an ideal world, we would have one more exception type with a name such as `CatastraphicFailure` as a parent of `Defect`. Such failures will be considered non-recoverable as well.

You can surely try to detect that the database is corrupted upfront and then handle the discovered corruption with a custom panic, but this doesn't conflict with the code here. I argue that any code that tries to look up Ethereum state data (that's what the Trie is used for), only to discover that the database has been tampered with, cannot safely recover from this situation. It would be inappropriate to continue execution because the risk of doing something wrong is too high. If we make this a recoverable error, it may be handled by the application in an unintended way.

arnetheduck: > `CatastraphicFailure`

well, there begins the slippery slope from "logic programming errors" to.. anything really -> more pressure to catch them -> moving away from their intended use

> it may be handled by the application in an unintended way.

well, this is an outcome of the architecture: error abstraction leakage and lack of intentional error discovery/api visibility/handling - it does not make clear to the caller that an unreliable external resource is involved - because clearly there exist cases where "better" error handling is imaginable and not a programmer error. anyway, sure - it's not a blocker for this PR but the comment would do well to point it out, so that future generations can do something better.

zah: Well, you are arguing that I'm twisting my definitions, but I beg to disagree. Here is what I've written in the past regarding this type of failures: 

https://gist.github.com/zah/d2d729b39d95a1dfedf8183ca35043b3
> **catastrophic system failures** These errors represent situations where it's simply not reasonable for us to recover - we have ran out of memory, a hardware device appears to malfunction, etc. While these situations are not bugs, mostly everything we've said about bugs applies here well. The code should not attempt to handle the problem, and the more context information we are able to gather the better.

Picture the code for a moment as if it was written according to your ideals. It's using an error result and this reaches the application code extracting data from the trie? What kind of code would you write there to handle the error? Remember that you'll have to do it at every single call site of the trie APIs.

PR #164, opened 2020-02-11
fixes #95

### LINKED ISSUE #95: 'payloadBytesCount' in rlp.nim behaves differently on 32 bit vs 64 bit
I found ['payloadBytesCount' in eth/rlp.nim](https://github.com/status-im/nim-eth/blob/bef28a18e58f52f24bcc50f0cadefac0224d4630/eth/rlp.nim#L117) behaves differently when compiled to 32bit. The [spec](https://github.com/ethereum/wiki/wiki/RLP) doesn't say anything about decoding to platform dependent 'int'. 

it's a bit confusing whether to fail on 32 bit or 64 bit. need deeper investigation.

jangko:
according  to nimbus tx test, `payloadBytesCount` should return 32 bit int

### DIFF
--- a/eth/rlp.nim
+++ b/eth/rlp.nim
@@ -115,6 +115,10 @@ template nonCanonicalNumberError =
   raise newException(MalformedRlpError, "Small number encoded in a non-canonical way")
 
 proc payloadBytesCount(self: Rlp): int =
+  type
+    ResType = type result
+    IntType = int32
+
   if not hasData():
     return 0
 
@@ -122,7 +126,7 @@ proc payloadBytesCount(self: Rlp): int =
   if marker < BLOB_START_MARKER:
     return 1
   if marker <= LEN_PREFIXED_BLOB_MARKER:
-    result = int(marker - BLOB_START_MARKER)
+    result = ResType(marker - BLOB_START_MARKER)
     readAheadCheck(result)
     if result == 1:
       if bytes[position + 1] < BLOB_START_MARKER:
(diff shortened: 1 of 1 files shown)


eth/rlp.nim
```diff
@@ -114,15 +116,15 @@ template readAheadCheck(numberOfBytes) =
 template nonCanonicalNumberError =
   raise newException(MalformedRlpError, "Small number encoded in a non-canonical way")
 
-proc payloadBytesCount(self: Rlp): int =
+proc payloadBytesCount(self: Rlp): RlpBytesInt =
```
zah: Can you show me the code affected by this change?

I'm asking because in Nim the length of something is usually returned as an `int` (e.g. `seq.len`). If the user of the API needs the length as a different type, she can easily convert it at the call site.

jangko: https://github.com/status-im/nim-eth/blob/4c82a31b9ad5b07c1b6450919354e23bdd4fa0e8/eth/rlp.nim#L146-L147

the value of `sizeof(result)` can be 4 or 8. probably we should replace it with a constant '4' or `sizeof(int32)`

jangko: but this is also not entirely correct. the nimbus tx test tell us the max payload len should not exceed `high(int32)`. We need to fix the comparison too.

jangko:
how about this new changes, the API stay the same, we isolate the changes inside the proc?

zah:
I've realized only now that this proc is internal to the library (private), so my concerns about the API are less important in this particular case. I'm happy to merge this.

zah:
Last question: are we sure the max payload length is `max(int32)` instead of `max(uint32)`?

jangko:
The spec silent about it. But I found something interesting in both `testeth` generator and runner.
The overflow error should be treated in platform dependent manner rather than uniform behaviour for all platform. If we sided with `testeth` as a `trusted reference` this PR is actually useless.
Sorry for all the trouble. I will close this PR. We only need to modify the test runner in Nimbus.

PR #175, opened 2020-02-24
Add lookupLoop and other fixes

kdeme:
- add lookupLoop
- protocol close / closeWait
- randomize randomNodes
- Use lookupRequestLimit
- Remove invalid check in neighbours proc
- Add lookup test

### DIFF
--- a/eth/p2p/discoveryv5/protocol.nim
+++ b/eth/p2p/discoveryv5/protocol.nim
@@ -1,7 +1,7 @@
 import
   std/[tables, sets, endians, options, math, random],
-  stew/byteutils, eth/[rlp, keys], chronicles, chronos, stint,
-  ../enode, types, encoding, node, routing_table, enr
+  json_serialization/std/net, stew/byteutils, chronicles, chronos, stint,
+  eth/[rlp, keys], ../enode, types, encoding, node, routing_table, enr
 
 import nimcrypto except toHex
 
@@ -168,7 +168,7 @@ proc receive*(d: Protocol, a: Address, msg: Bytes) {.gcsafe,
   # debug "Packet received: ", length = msg.len
 
   if d.isWhoAreYou(msg):
-    trace "Received whoareyou", localNode = d.localNode, address = a
+    trace "Received whoareyou", localNode = $d.localNode, address = a
     let whoareyou = d.decodeWhoAreYou(msg)
     var pr: PendingRequest
     if d.pendingRequests.take(whoareyou.authTag, pr):
@@ -194,7 +194,7 @@ proc receive*(d: Protocol, a: Address, msg: Bytes) {.gcsafe,
       if node.isNil:
         node = d.routingTable.getNode(sender)
       else:
-        debug "Adding new node to routing table", node, localNode = d.localNode
+        debug "Adding new node to routing table", node = $node, localNode = $d.localNode
         discard d.routingTable.addNode(node)
 
       doAssert(not node.isNil, "No node in the routing table (internal error?)")
@@ -212,7 +212,7 @@ proc receive*(d: Protocol, a: Address, msg: Bytes) {.gcsafe,
           debug "TODO: handle packet: ", packet = packet.kind, origin = $node
     elif decoded == DecodeStatus.PacketError:
       debug "Could not decode packet, respond with whoareyou",
-        localNode = d.localNode, address = a
+        localNode = $d.localNode, address = a
       d.sendWhoareyou(a, sender, authTag)
     # No Whoareyou in case it is a Handshake Failure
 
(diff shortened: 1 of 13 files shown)


eth/p2p/discoveryv5/protocol.nim
```diff
@@ -69,38 +78,47 @@ proc send(d: Protocol, a: Address, data: seq[byte]) =
...
+  # If there is already a handshake going on for this nodeid then we drop this
+  # new one. Handshake will get cleaned up after `handshakeTimeout`.
+  # If instead overwriting the handshake would be allowed, the handshake timeout
+  # will need to be canceled each time.
+  # TODO: could also clean up handshakes in a seperate call, e.g. triggered in
+  # a loop.
+  # Use toNode + address to make it more difficult for an attacker to occupy
+  # the handshake of another node.
+
+  let key = HandShakeKey(nodeId: toNode, address: $address)
```
zah: Since this key is not unique for the request, there is a remote chance for a race condition here:

1) You send a request to a particular peer
2) Another request is sent to the same peer
3) The first request times out and cleans up `d.codec.handshakes` for the peer key
4) We receive an answer for the second request, but there is no handshake record anymore

eth/p2p/discoveryv5/protocol.nim
```diff
@@ -333,17 +355,56 @@ proc revalidateNode(p: Protocol, n: Node) {.async.} =
...
-      await p.revalidateNode(n)
+  try:
+    # TODO: We need to handle actual errors still, which might just allow to
+    # continue the loop. However, currently `revalidateNode` raises a general
+    # `Exception` making this rather hard.
+    while true:
+      await sleepAsync(rand(10 * 1000).milliseconds)
+      let n = p.routingTable.nodeToRevalidate()
+      if not n.isNil:
+        await p.revalidateNode(n)
```
zah: May be a good idea to do these in parallel (a TODO comment would suffice for now).
Otherwise, a high number of high latency peers could easily slow the progress of this loop enough to produce unexpected effects (peers being re-examined in significantly more than 1-second intervals).

zah: In the long-term, we probably need some kind of general throttling mechanism that can execute a list of async tasks by calling at most N in parallel at any given time.

eth/p2p/discoveryv5/protocol.nim
```diff
@@ -333,17 +355,56 @@ proc revalidateNode(p: Protocol, n: Node) {.async.} =
...
+  debug "Closing discovery node", n = d.localNode
+  d.revalidateLoop.cancel()
+  d.lookupLoop.cancel()
+  # TODO: unsure if close can't create issues in the not awaited cancellations
+  # above
+  d.transp.close()
+
+proc closeWait*(d: Protocol) {.async.} =
+  doAssert(not d.lookupLoop.isNil() or not d.revalidateLoop.isNil())
+  doAssert(not d.transp.closed)
```
zah: Ditto

eth/p2p/discoveryv5/routing_table.nim
```diff
@@ -130,6 +130,7 @@ proc computeSharedPrefixBits(nodes: openarray[Node]): int =
 proc init*(r: var RoutingTable, thisNode: Node) {.inline.} =
   r.thisNode = thisNode
   r.buckets = @[newKBucket(0.u256, high(Uint256))]
+  randomize() # for later `randomNodes` selection
```
zah: Should we use a more secure random source in `randomNodes`? Perhaps, yes.

eth/p2p/discoveryv5/protocol.nim
kdeme: Well, there are different ways of handling these multiple requests from the same peer. I decided (for now?) to solve by not allowing a second handshake to be started.  So in your example point 2. will not happen, the second whoareyou message will not be send.

eth/p2p/discoveryv5/protocol.nim
```diff
@@ -333,17 +355,56 @@ proc revalidateNode(p: Protocol, n: Node) {.async.} =
...
   let ta = initTAddress(IPv4_any(), d.localNode.node.address.udpPort)
   d.transp = newDatagramTransport(processClient, udata = d, local = ta)
-  asyncCheck d.revalidateLoop() # TODO: This loop has to be terminated on close()
+  # Might want to move these to a separate proc if this turns out to be needed.
+  d.lookupLoop = lookupLoop(d)
+  d.revalidateLoop = revalidateLoop(d)
+
+proc close*(d: Protocol) =
+  doAssert(not d.lookupLoop.isNil() or not d.revalidateLoop.isNil())
+  doAssert(not d.transp.closed)
```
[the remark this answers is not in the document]
kdeme: Good point, not sure..

eth/p2p/discoveryv5/routing_table.nim
kdeme: Yeah, thought about that very briefly while adding `randomize()`. Not sure how difficult it would be to abuse right now. But perhaps.

eth/p2p/discoveryv5/protocol.nim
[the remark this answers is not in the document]
kdeme: @cheatfate confirmed that `closed` will only return true after `close` got called.

However, calling `close` twice is allowed and will be just a nop, so I could make this close also a nop.

PR #176, opened 2020-02-24
safeEnterList() renamed to enterList()

stefantalpalaru:
...in order to avoid unhandled exceptions due to malformed data received
over the network, like sometimes when running `make testnet1` in nim-beacon-chain:

```text
DBG 2020-02-24 23:37:56+01:00 Got discv5 lookup response                 topics="discv5" tid=26545 idx=1
DBG 2020-02-24 23:37:56+01:00 discv5 pending queries                     topics="discv5" tid=26545 total=3
DBG 2020-02-24 23:37:56+01:00 Got discv5 lookup response                 topics="discv5" tid=26545 idx=2
DBG 2020-02-24 23:37:56+01:00 discv5 pending queries                     topics="discv5" tid=26545 total=3
TRC 2020-02-24 23:37:56+01:00 Decoding failed (no keys)                  tid=26545
DBG 2020-02-24 23:37:56+01:00 Could not decode, respond with whoareyou   topics="discv5" tid=26545
TRC 2020-02-24 23:37:56+01:00 sending who are you                        topics="discv5" tid=26545 to=71903746117165936395423247506562441065595177867592707754147235596739826250520 toAddress="(ip: 3.126.82.94, udpPort: 9100, tcpPort: 9100)"
 peers: 0 ❯ epoch: 3403, slot: 2/8 (27226) ❯ finalized epoch: 0 (1bdb7c6b)                                                                      ETH: 0 Traceback (most recent call last, using override)
/mnt/sda3/storage/CODE/status/nim-beacon-chain/beacon_chain/beacon_node.nim(1172) beacon_node
/mnt/sda3/storage/CODE/status/nim-beacon-chain/beacon_chain/beacon_node.nim(931) start
/mnt/sda3/storage/CODE/status/nim-beacon-chain/beacon_chain/beacon_node.nim(889) run
/mnt/sda3/storage/CODE/status/nim-beacon-chain/vendor/nim-chronos/chronos/asyncloop.nim(941) runForever
/mnt/sda3/storage/CODE/status/nim-beacon-chain/vendor/nim-chronos/chronos/asyncloop.nim(278) poll
Error: unhandled exception: /mnt/sda3/storage/CODE/status/nim-beacon-chain/vendor/nim-eth/eth/rlp.nim(246, 12) `isList()` 
Async traceback:
Exception message: /mnt/sda3/storage/CODE/status/nim-beacon-chain/vendor/nim-eth/eth/rlp.nim(246, 12) `isList()` 
Exception type: [AssertionError]
```

### DIFF
--- a/eth/p2p/rlpx_protocols/waku_protocol.nim
+++ b/eth/p2p/rlpx_protocols/waku_protocol.nim
@@ -147,17 +147,9 @@ proc append*(rlpWriter: var RlpWriter, value: StatusOptions) =
   rlpWriter.append(rlpFromBytes(bytes.toRange))
 
 proc read*(rlp: var Rlp, T: typedesc[StatusOptions]): T =
-  if not rlp.isList():
-    raise newException(RlpTypeMismatch,
-      "List expected, but the source RLP is not a list.")
-
   let sz = rlp.listLen()
   rlp.enterList()
   for i in 0 ..< sz:
-    if not rlp.isList():
-      raise newException(RlpTypeMismatch,
-        "List expected, but the source RLP is not a list.")
-
     rlp.enterList()
     var k: KeyKind
     try:
--- a/eth/trie/hexary.nim
+++ b/eth/trie/hexary.nim
@@ -384,7 +384,7 @@ proc replaceValue(data: Rlp, key: NibblesRange, value: BytesRange): Bytes =
 
   var r = initRlpList(17)
 
-  # XXX: This can be optmized to a direct bitwise copy of the source RLP
+  # XXX: This can be optimized to a direct bitwise copy of the source RLP
   var iter = data
   iter.enterList()
   for i in 0 ..< 16:
--- a/eth/p2p/rlpx.nim
+++ b/eth/p2p/rlpx.nim
@@ -567,7 +567,7 @@ proc p2pProtocolBackendImpl*(protocol: P2PProtocol): Backend =
     read = bindSym("read", brForceOpen)
     checkedRlpRead = bindSym "checkedRlpRead"
     startList = bindSym "startList"
-    safeEnterList = bindSym "safeEnterList"
+    enterList = bindSym "enterList"
     finish = bindSym "finish"
 
     messagePrinter = bindSym "messagePrinter"
@@ -677,7 +677,7 @@ proc p2pProtocolBackendImpl*(protocol: P2PProtocol): Backend =
 
     let
       paramCount = paramsToWrite.len
-      readParamsPrelude = if paramCount > 1: newCall(safeEnterList, receivedRlp)
+      readParamsPrelude = if paramCount > 1: newCall(enterList, receivedRlp)
                           else: newStmtList()
 
     when tracingEnabled:
--- a/eth/rlp.nim
+++ b/eth/rlp.nim
@@ -243,13 +243,9 @@ proc currentElemEnd*(self: Rlp): int =
     result += payloadOffset() + payloadBytesCount()
 
 proc enterList*(self: var Rlp) =
-  doAssert isList()
-  position += payloadOffset()
-
-proc safeEnterList*(self: var Rlp) =
   if not isList():
     raise newException(RlpTypeMismatch, "List expected, but source RLP is not a list")
-  enterList()
+  position += payloadOffset()
 
 proc skipElem*(rlp: var Rlp) =
   rlp.position = rlp.currentElemEnd


eth/trie/hexary.nim
```diff
@@ -386,7 +386,7 @@ proc replaceValue(data: Rlp, key: NibblesRange, value: BytesRange): Bytes =
 
   # XXX: This can be optmized to a direct bitwise copy of the source RLP
   var iter = data
-  iter.enterList()
+  iter.safeEnterList()
```
kdeme: Ok for me, but perhaps the ones in hexary need to be seen as programming errors. @jangko could probably better comment on these.

arnetheduck: seems like an api design flaw that you can end up with unsafe constructs like this / that it's not default that things are checked - this should be part of any untainting done as data enters from outside - what can be done here? rather than changing the call here, why not make `enterList` safe? if there's a special need for an unsafe one, make it the non-default, ie `unsafeEnterList` or something.

 @zah

arnetheduck: in part, this is why not making errors plain in the visible API of a function is so important: developers _do not realise_ that it can fail because most of the time, correct data arrives, and only when there is a deliberate attempt (or a bug/version mismatch/etc) do they realise that something else might happen - unless reminded, it's easy to forget testing malicious inputs - this is what designing an API with safety in mind achieves

kdeme: > if there's a special need for an unsafe one, make it the non-default, ie `unsafeEnterList` or something.

Yeah, that would be better. I actually added the safe one a while back when fixing similar errors as this PR fixes. Should have probably done the inverse like you said, as in the meanwhile new cases of misuse in discovery v5 were added.
And in Waku code I actually added checks manually, I must have forgotted about the safe one I introduced already...

eth/p2p/rlpx_protocols/waku_protocol.nim
```diff
@@ -152,13 +152,13 @@ proc read*(rlp: var Rlp, T: typedesc[StatusOptions]): T =
       "List expected, but the source RLP is not a list.")
 
   let sz = rlp.listLen()
-  rlp.enterList()
+  rlp.safeEnterList()
```
kdeme: Checks above for isList can be removed

eth/p2p/rlpx_protocols/waku_protocol.nim
```diff
@@ -152,13 +152,13 @@ proc read*(rlp: var Rlp, T: typedesc[StatusOptions]): T =
...
   for i in 0 ..< sz:
     if not rlp.isList():
       raise newException(RlpTypeMismatch,
         "List expected, but the source RLP is not a list.")
 
-    rlp.enterList()
+    rlp.safeEnterList()
```
kdeme: idem ^

eth/trie/hexary.nim
zah: The difference between `enterList` and `safeEnterList` is that the former has a pre-condition that the rlp is known to be a list, while the later will raise a recoverable error when given a non-list.

In other words, the correct way to use `enterList` is the following:

```nim
if rlp.isList:
  rlp.enterList
  ...
```

while the correct way to use `safeEnterList` is like this:

```nim
try:
  rlp.safeEnterList
  ...
except RlpTypeMismatchError:
  ...
```

It's not possible to create an API that's inherently safer and doesn't require some form of "is this a list?" check. Since RLP is quite a dynamic format (like JSON), I personally think the first version of the code is more natural to write and that's why it's considered the default. `safeEnterList` could have been better named `tryEnterList` to downplay the "safe" meaning and to stress the fact that exceptions may be raised.


Usually, you argue for a world without pre-conditions. An API that is used like this:

```nim
let wasThisAList = rlp.enterList
if wasThisAList:
  ...
```

You have a point that in the current version of Nim, this is slightly more ergonomic, but you have to recognize that it's also a bit silly. I'd still argue that the better way to write the code would have been:

```nim
if rlp.isList:
  discard rlp.enterList
```

For this API in particular, we might make this change if we didn't have other more important issues to worry about, but for me the long-term solution is to embrace pre-conditions and to improve the ergonomics by allowing the compiler to point out when they are violated (that's what the upcoming Z3 integration is about).

stefantalpalaru:
Strategy changed. `safeEnterList()` was renamed to `enterList()` and those couple of now redundant `isList()` checks have been removed.

zah:
With this new strategy, now the users are forced to use the exceptions-based API to write correct code.

Consider a function that has a `raises` list. If you try to use up-front `isList` checks, the compiler will be unhappy because it will insist that the `RlpTypeMismatchError` is still possible. You'll have to write the code like this:

```nim
proc handleRlp(rlp: Rlp) {. raises: [Defect] .} =
  if rlp.isList:
    try:
      rlp.enterList
    except RlpTypeMismatchError:
      discard # this should never happen
```

stefantalpalaru:
> now the users are forced to use the exceptions-based API

They always were. Both functions were raising exceptions: one implicitly from `doAssert`, the other explicitly with `raise`.

zah:
`doAssert` does not produce a recoverable error. The users are not supposed to handle `Defects` and that's why the `raises` signature in the example above is allowing them to pass through.

kdeme:
The doAssert should raise a `Defect` or atleast not a `CatchableError`, hence the remark of @zah

eth/trie/hexary.nim
arnetheduck: @zah both your api force the user to remember to do something magic (inserting an if or remembering to stick a `raises` in there) before calling the function, which simply doesn't happen (consistently) - people don't read documentation, and they don't remember strange things they need to do to make unusal cases work because typically, they don't see them and cannot imagine what will go wrong. they may with all the best intentions try to be thorough, but in practice the problem is too complex for humans to deal with well, specially when they're busy trying to just make it work under the pressure of deadlines.

This is the source of security bugs - ie this is why `strcpy` is considered unsafe: you have to validate that it's a zero-terminated string or it will overflow and by and large, developers don't - not consistently enough - it's an elementary property of secure design.

Since parsing deals with external data that adversaries control can input, it's simply wrong to design api like this - there should be no way for this to assert unless it's painfully obvious that you're doing something unusual - awkward naming, compiler errors that need disarming etc

we can talk all we want about magic unicorns that maybe will be available in 2-3 years in the nim compiler (at best) - but we're releasing the software to audit in 3 months - will your Z3 features be ready and stable by then? keep in mind that nim is not the first time z3 and proving systems are considered for proving things in languages, but there are theoretical reasons that make this difficult except for the most trivial of cases.

It is entirely possible to design a safe API without these magic pills in the compiler - all they do is make things a little more ergonomic - it was even possible to design them in C.

eth/nimbus is not the right sandbox to by trying out things in - they can be tried in experimental branches that are not meant for production, or other projects that don't deal with cryptocurrencies - when the features become available in nim, rewriting is cheap - you only have to refactor some code which is easy/trivial (because now you have a working and well-tested codebase that's perfectly prepared for it).

arnetheduck: ...

```
let wasThisAList = rlp.enterList
if wasThisAList:
```

why is this silly? it perfectly expresses what can happen and forces the caller to deal with it - they implicitly learn that entering a list sometimes may fail, and can deal with the problem deliberately, instead of after a security issue in the wild.

if you're doing exceptions, raising a catchable exception is equally correct: you're trying to amortize error-handling costs by aggregating them at some boundary - ideally, the composite parse function would be `raises[]` so that you group all parsing errors and teach the caller (that's probably trying to load an entire object) that the format is not total - it's less good because again, you're forcing the caller of your API to remember to do something, which people are bad at.

zah:
Why was this merged? There were clear outstanding objections against the latest version of the code

stefantalpalaru:
> Why was this merged?

Because bug fixes shouldn't have to wait for the conclusion of those tangent discussions they inspired.

> There were clear outstanding objections against the latest version of the code

Like what? The hypothetical future use of the {.raises.} pragma? Fix that in the hypothetical future and let me have my fix in the present.

By all means, improve the code, but do it after merging the good-enough fix. We can then discuss all the fascinating aspects of surprise doAssert runtime exceptions at our leisure.

Issue #168, opened 2020-02-14
Allow for changing between topic-interest and bloom filter by means of new packet type


kdeme:
When topic-interest is used, we currently do now allow to switch back to usage of bloom filter and vice versa.

See also comment here: https://github.com/status-im/nim-eth/blob/master/eth/p2p/rlpx_protocols/waku_protocol.nim#L368

This issue has been raised and a new packet type with a `StatusOptions` parameter will be added to to specification.
As soon as this is done, this needs to be added to the waku_protocol here to be compliant.
This might include the removal of some of the current packets types.

See also here for the status-go version: https://github.com/status-im/status-go/issues/1853

kdeme:
closed by #178

PR #178, opened 2020-02-26
Handle waku status-update code 22 & increase max topics

cammellos:
Handle waku status-update code 22

This commits adds handling of the status-update code, 22, according to waku specs 0.4 https://github.com/vacp2p/specs/commit/a6a9d6bcb127044e13b47c229da7a5b3f66eadf1
    
Removes unused codes 20,21

Fields not used (RateLimits & ConfirmationsEnabled) are currently ignored.

### DIFF
--- a/eth/p2p/rlpx_protocols/waku_protocol.nim
+++ b/eth/p2p/rlpx_protocols/waku_protocol.nim
@@ -69,7 +69,7 @@ const
   ## send to peers, in ms.
   pruneInterval* = chronos.milliseconds(1000)  ## Interval at which message
   ## queue is pruned, in ms.
-  topicInterestMax = 1000
+  topicInterestMax = 10000
 
 type
   WakuConfig* = object
--- a/eth/p2p/rlpx_protocols/waku_protocol.nim
+++ b/eth/p2p/rlpx_protocols/waku_protocol.nim
@@ -74,7 +74,7 @@ const
 type
   WakuConfig* = object
     powRequirement*: float64
-    bloom*: Bloom
+    bloom*: Option[Bloom]
     isLightNode*: bool
     maxMsgSize*: uint32
     confirmationsEnabled*: bool
@@ -199,7 +199,7 @@ proc allowed*(msg: Message, config: WakuConfig): bool =
       warn "Message topic does not match Waku topic list"
       return false
   else:
-    if not bloomFilterMatch(config.bloom, msg.bloom):
+    if config.bloom.isSome() and not bloomFilterMatch(config.bloom.get(), msg.bloom):
       dropped_bloom_filter_mismatch_envelopes.inc()
       warn "Message does not match node bloom filter"
       return false
@@ -213,7 +213,7 @@ proc initProtocolState*(network: WakuNetwork, node: EthereumNode) {.gcsafe.} =
   new(network.queue)
   network.queue[] = initQueue(defaultQueueCapacity)
   network.filters = initTable[string, Filter]()
-  network.config.bloom = fullBloom()
+  network.config.bloom = some(fullBloom())
   network.config.powRequirement = defaultMinPow
   network.config.isLightNode = false
   # RateLimits and confirmations are not yet implemented so we set confirmations
@@ -237,7 +237,7 @@ p2pProtocol Waku(version = wakuVersion,
 
     let list = StatusOptions(
       powRequirement: some(wakuNet.config.powRequirement),
-      bloomFilter: some(wakuNet.config.bloom),
+      bloomFilter: wakuNet.config.bloom,
       lightNode: some(wakuNet.config.isLightNode),
       confirmationsEnabled: some(wakuNet.config.confirmationsEnabled),
       rateLimits: wakuNet.config.rateLimits,
@@ -328,40 +328,24 @@ p2pProtocol Waku(version = wakuVersion,
         # notify filters of this message
         peer.networkState.filters.notify(msg)
 
-  proc powRequirement(peer: Peer, value: uint64) =
+  nextID 22
+
+  proc statusOptions(peer: Peer, options: StatusOptions) =
     if not peer.state.initialized:
-      warn "Handshake not completed yet, discarding powRequirement"
+      warn "Handshake not completed yet, discarding statusOptions"
       return
 
-    peer.state.powRequirement = cast[float64](value)
+    if options.topicInterest.isSome():
+      peer.state.topics = options.topicInterest
+    elif options.bloomFilter.isSome():
+      peer.state.bloom = options.bloomFilter.get()
+      peer.state.topics = none(seq[Topic])
 
-  proc bloomFilterExchange(peer: Peer, bloom: Bytes) =
-    if not peer.state.initialized:
-      warn "Handshake not completed yet, discarding bloomFilterExchange"
-      return
+    if options.powRequirement.isSome():
+      peer.state.powRequirement = options.powRequirement.get()
 
-    if bloom.len == bloomSize:
-      peer.state.bloom.bytesCopy(bloom)
-
-  nextID 20
-
-  proc rateLimits(peer: Peer, rateLimits: RateLimits) = discard
-
-  proc topicInterest(peer: Peer, topics: openarray[Topic]) =
-    if not peer.state.initialized:
-      warn "Handshake not completed yet, discarding topicInterest"
-      return
-
-    if topics.len > topicInterestMax:
-      error "Too many topics in the topic-interest list"
-      return
-
-    # TODO: We currently do not allow changing topic-interest.
-    # We could allow this by also resetting topic-interest on a bloom filter
-    # exchange. But we don't for now and will instead introduce another packet
-    # type in the Waku Specification.
-    if peer.state.topics.isSome():
-      peer.state.topics = some(topics)
+    if options.lightNode.isSome():
+      peer.state.isLightNode = options.lightNode.get()
 
   nextID 126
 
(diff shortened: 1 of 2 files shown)


eth/p2p/rlpx_protocols/waku_protocol.nim
```diff
@@ -328,40 +328,25 @@ p2pProtocol Waku(version = wakuVersion,
...
 
-    if bloom.len == bloomSize:
-      peer.state.bloom.bytesCopy(bloom)
-
-  nextID 20
-
-  proc rateLimits(peer: Peer, rateLimits: RateLimits) = discard
+    if options.topicInterest.isSome():
+      peer.state.topics = options.topicInterest
+      peer.state.bloom = fullBloom()
```
kdeme: The `bloom` doesn't necessarily need to be changed here right? It should not be used anyhow as topicInterest is set.

eth/p2p/rlpx_protocols/waku_protocol.nim
```diff
@@ -609,13 +599,13 @@ proc setTopicInterest*(node: EthereumNode, topics: seq[Topic]):
   if topics.len > topicInterestMax:
     return false
 
-  if node.protocolState(Waku).config.topics.isNone():
-    return false
-
   node.protocolState(Waku).config.topics = some(topics)
+  node.protocolState(Waku).config.bloom = fullBloom()
```
kdeme: Same here as in the `statusOptions`. Not really needed?

eth/p2p/rlpx_protocols/waku_protocol.nim
```diff
@@ -638,13 +628,18 @@ proc setPeerTrusted*(node: EthereumNode, peerId: NodeId): bool =
       peer.state(Waku).trusted = true
       return true
 
-proc setLightNode*(node: EthereumNode, isLightNode: bool) =
+proc setLightNode*(node: EthereumNode, isLightNode: bool) {.async.} =
   ## Set this node as a Waku light node.
-  ##
-  ## NOTE: Should be run before connection is made with peers as this
-  ## setting is only communicated at peer handshake.
   node.protocolState(Waku).config.isLightNode = isLightNode
```
kdeme:
```suggestion
# TODO: Add starting/stopping of `processQueue` loop depending on value of isLightNode.
```

tests/p2p/test_waku_connect.nim
```diff
@@ -44,6 +44,119 @@ suite "Waku connections":
...
+    # Set topic interest
+    discard await wakuBloomNode.setTopicInterest(topics)
+
+    wakuNode.startListening()
+    await wakuBloomNode.peerPool.connectToNode(newNode(
+      initENode(wakuNode.keys.pubKey, wakuNode.address)))
+
+    # Sanity check
+    check:
+      wakuNode.peerPool.len == 1
```
kdeme: The CI failure here happens because you are not certain that `wakuNode` has gone through the whole process of connecting and adding `wakuBloomNode` to the peerPool.
You only know this for the `wakuBloomNode` that is connecting & adding `wakuNode`

kdeme:
For reference: this implements https://github.com/status-im/nim-eth/issues/168

tests/p2p/test_waku_connect.nim
kdeme: I think you can get around it in your test case by changing the direction of connecting.

eth/p2p/rlpx_protocols/waku_protocol.nim
```diff
@@ -638,12 +628,17 @@ proc setPeerTrusted*(node: EthereumNode, peerId: NodeId): bool =
...
+# TODO: Add starting/stopping of `processQueue` loop depending on value of isLightNode.
+  var futures: seq[Future[void]] = @[]
+  for peer in node.peers(Waku):
+    let list = StatusOptions(lightNode: some(isLightNode))
+    futures.add(peer.statusOptions(list))
+
+  # Exceptions from sendMsg will not be raised
+  await allFutures(futures)
```
zah: Please note that if any of the async operations (futures) terminates with an error, `await allFutures` won't report this in any way. To handle the errors, you are supposed to iterate over the list of `futures` after the await operation and to check whether any of them has a non-nil `error` field. Depending on the intention of the code, you can log something, reduce peer reputation, retry the operation, etc.

kdeme: True, and in the already existing calls I specifically used `allFutures` for that reason. See also the documentation/comments.

It is done this way so that the caller doesn't need to care, as in fact, there is not much you can do anyway as caller.  An error that would/could typically occur here is e.g. that one of the peers got disconnected in the meanwhile.
Yes, if it turns out to be needed these calls could do additional things such as retry on error, etc.

cammellos:
@kdeme I have noticed that the tests failed in the previous run, this is likely due to the fact that we `sleepAsync(timeout)` and check the condition, assuming that `timeout` is enough (looks like it might not be in some cases).

 I have changed the code to use `eventually` https://github.com/status-im/nim-eth/pull/178/commits/93445756595d1729566a9b24880102e3cc1c461c 

Which will poll for the condition every `100ms`, and `sleepAsync` so it's not busy-waiting and should release control to the main thread, for a maximum of 1 second, returning true/false based on the condtion passed. (at least that's what I was trying to do, whether I actually achieved that or made a mess of it, not sure :) )

The plus would be that is less likely to be flaky  (as long as max timeout is enough) and should be faster (as we won't wait always for the max amount), the drawback is complexity of course and the fact that we lose expressiveness in the check, as which check has failed is actually swallowed (in case of multiple checks). Ideally it'd be nicer to have a macro, similar to `check`, like `eventually:`.

I have only changed a few for now as not sure what you think of it, if it's ok I can change the rest of the tests otherwise we can figure out a different way.
Thanks!

cammellos:
didn't help at all :D , back to the drawing board I guess

kdeme:
@cammellos Yeah, that's what I meant with my comment here https://github.com/status-im/nim-eth/pull/178#discussion_r384378909 . And it is quite brittle due to possible too short waiting time.

As mentioned, for the postMessage we can use the filters to notice an incoming message.
For the other packet types, we can't, so your `eventually` solution is indeed better there. At first glance, this should work. I'll have a better look later. There could also still be something else wrong with the packet being send/received.

cammellos:
@kdeme thanks,

I have added a few debug statements and see what is happening exactly. So far I was not able to replicate locally, and all the builds have passed at least once (windows builds have failed more consistently but not always), so looks like it might be still a timing issue.

Some of the builds are also hanging on the `[Suite] sparse binary trie` tests (happened twice), I suppose it's unrelated, but not sure.

Thanks!

cammellos:
@kdeme I have figured out what was wrong (I think :) ), essentially we need to wait until on the other side the peers are connected, otherwise the update won't be propagated.
I'll run a few times the build to check if it's stable now.

kdeme:
I can reproduce it in my Windows VM. The bloomFilter update never gets send to the peer as the peer is not yet in the peerPools list (same issue kind of that we tried to work around). So I guess you'll need an eventually for the peerPool.len also.

cammellos:
@kdeme just figured out as well, pushed a fix :)

eth/p2p/rlpx_protocols/waku_protocol.nim
arnetheduck: perhaps we should deprecate `allFutures`? it's hardly a safe approach - it also sounds like it will swallow `Defect`s?

tests/p2p/test_waku_connect.nim
```diff
@@ -15,6 +15,18 @@ import
...
   waitInterval = messageInterval + 150.milliseconds
+  conditionTimeoutMs = 3000
+
+# check on a condition until true or return a future containing false
+# if timeout expires first
+proc eventually(timeout: int, condition: proc(): bool): Future[bool] =
+  let wrappedCondition = proc(): Future[bool] {.async.} =
+    let f = newFuture[bool]()
+    while not condition():
+      await sleepAsync(100.milliseconds)
```
arnetheduck: if `sleepAsync` is needed, it's usually a design smell in the api that you cannot deterministically wait for an operation to be performed - ideally, async operations should return futures that can be waited for - we've thought about deprecating `sleepAsync` for this reason as it often gets misused, just like `sleep` often gets misused in multi-threaded programs.. 

what's the reason it's needed here?

eth/p2p/rlpx_protocols/waku_protocol.nim
kdeme: Well, I certainly hope it doesn't swallow `Defect`s, it shouldn't.

And iirc @cheatfate introduce `allFutures` because he wanted to deprecate `all` as it is considered "broken" (In the sense that you don't know about all the possible failures, so if you want that information, you'd have to do like @zah suggests).

tests/p2p/test_waku_connect.nim
kdeme: Yes, you are right, it should be avoided. The tests added here are sending status update messages, and it is checked on the receiving peer if these values get set. 
For envelopes we can do without `sleepAsync` as we can register on received envelopes (filters). However, for the other packet types we don't have this, so it is perhaps slightly more cumbersome. 

However, it can be done I believe with the `nextMsg` call, as I do e.g. here:
https://github.com/status-im/nim-eth/blob/master/tests/p2p/test_waku_mail.nim#L47 , so yes, no reason I guess to keep the `sleepAsync`.

Anyway, there is an issue open to rework and get rid of all `sleepAsync`'s that remain in the tests.

cammellos: Basically the test is an integration test:

1) Spin up node A
2) Spin up node B
3) Update settings in node A -> Node A will send a message to node B, which I believe is not a simple synchronous Request/Response pattern, and B will process this message asynchronously at their leisure. i.e even if we wait for a future on A (which I believe we do), B might have not yet processed the message and updated its state.

So the only way I can see to know this, is to watch a piece of state in B to make sure it's processed (I used sleep async not to busy wait, but a `while` loop would be essentially the same)

@kdeme correct me if anything is inaccurate

cammellos: oh didn't know about `nextMsg`, thanks

PR #180, opened 2020-02-27
`enterList` now returns a bool

zah:


eth/p2p/rlpx_protocols/waku_protocol.nim
```diff
@@ -147,10 +147,18 @@ proc append*(rlpWriter: var RlpWriter, value: StatusOptions) =
...
+    raise newException(RlpTypeMismatch,
+      "List expected, but the source RLP is not a list.")
+
   let sz = rlp.listLen()
-  rlp.enterList()
+  # We already know that we are working with a list
+  discard rlp.enterList()
   for i in 0 ..< sz:
-    rlp.enterList()
+    if not rlp.enterList():
```
kdeme: I guess here `tryEnterList` could be used

tests/rlp/test_api_usage.nim
```diff
@@ -102,7 +102,7 @@ test "encode and decode lists":
   var list = rlpFromBytes encodeList(rlp.listELem(1), rlp.listELem(0)).toRange
 
   # test that iteration with enterList/skipElem works as expected
-  list.enterList
+  discard list.enterList # We alreay know that we are working with a list
```
arnetheduck: why not assert it instead of discarding?

PR #201, opened 2020-03-26
enhancement/enr-nodeid

decanus:


decanus:
@kdeme I think this does it, I will add tests and reduce duplication

eth/p2p/discoveryv5/node.nim
```diff
@@ -40,10 +40,7 @@ proc newNode*(r: Record): Node =
   except KeyError:
     discard
 
-  var pk: PublicKey
-  if recoverPublicKey(r.get("secp256k1", seq[byte]), pk) != EthKeysStatus.Success:
-    warn "Could not recover public key"
-    return
+  let pk = r.toNodeID()
```
kdeme: But we need the public key here, not the NodeId?

decanus: @kdeme you're right I wasn't fully thinking straight

eth/p2p/discoveryv5/enr.nim
```diff
@@ -43,6 +43,8 @@ type
     of kBytes:
       bytes: seq[byte]
 
+  ToNodeIDError = Exception
```
kdeme:
```suggestion
  ToNodeIDError = object of CatchableError
```

decanus:
@kdeme finally

decanus:
@kdeme can we merge?

arnetheduck:
test case missing

eth/p2p/discoveryv5/enr.nim
```diff
@@ -2,9 +2,9 @@
 # https://github.com/ethereum/EIPs/blob/master/EIPS/eip-778.md
 
 import
-  net, strutils, macros, algorithm, options,
+  net, stint, strutils, macros, algorithm, options,
   nimcrypto, stew/base64,
-  eth/[rlp, keys], ../enode
+  eth/[rlp, keys], ../enode, ../kademlia
```
kdeme, line 7: Currently discoveryv5 is not well integrated with the rest of nim-eth codebase at all. And lots of duplication exists. But we shouldn't import from kademlia here (which is discv4, and eventually discv4 should use also ENR). I understand it is currently the only way, due to recursive import if you take it from types.nim 
I'll see for a better (common) location for NodeId.

eth/p2p/discoveryv5/enr.nim
```diff
@@ -285,6 +287,13 @@ template fromURI*(r: var Record, url: EnrUri): bool =
 proc toBase64*(r: Record): string =
   result = Base64Url.encode(r.raw)
 
+proc toNodeID*(r: Record): NodeId =
+  var pk: PublicKey
+  if recoverPublicKey(r.get("secp256k1", seq[byte]), pk) != EthKeysStatus.Success:
+    raise newException(ToNodeIDError, "Could not recover public key")
```
kdeme, line 293: Sorry for the delay for second review, but in the meanwhile some changes got in and you should use `PublicKey.fromRaw`. Or better, just reuse the existing `get` call.

Also, I know the current codebase does this kind of error handling all over the place. But I think ideally we move this code to a more `Result` and/or `Option` oriented error handling. It will also fit better with the underlying calls that were adapted to use `Result` / `Option` here (see latest changes done by @arnetheduck)

kdeme:
By the way, currently a `Node` has the `NodeId` information redundantly available as an `id` field outside the `Record`. This field is public, so if the Node is correctly initialized, you could get the `NodeId` through there. Disclaimer: Not sure if we will keep that redundant information around in this form. :)

arnetheduck:
code has moved on, this would need a reboot

PR #203, opened 2020-03-31
refactor secp interface

arnetheduck:
* avoid mix of exceptions and return codes
* introduce Result
* use deprecated compatibility API to avoid big-bang refactoring
* loosely based on nim-libp2p secp as well as the rust-secp256k1 wrapper
* oh, boy does our secp256k1 wrapper need updating - we're _far_
behind...


arnetheduck:
depends on https://github.com/status-im/nim-stew/pull/23

arnetheduck:
this is a first step to not disrupt too much code - there's more to do on this front

eth/keys.nim
```diff
@@ -6,105 +6,127 @@
...
+    ## Size of uncompressed public key without format marker (0x04)
+  RawSignatureSize* = 65
 
-proc isZeroKey*(pubkey: PublicKey): bool =
-  ## Check if public key `pubkey` contains only 0 bytes.
-  result = true
-  for i in pubkey.data:
-    if i != byte(0):
-      result = false
+proc toPublicKey*(seckey: PrivateKey): SkResult[PublicKey] =
```
mratsim, line 46: I call that `privToPub` in blscurve, following py-ecc lead

https://github.com/ethereum/py_ecc/search?q=privToPub&unscoped_q=privToPub

eth/keys.nim
```diff
@@ -6,105 +6,127 @@
...
+  SkPrivateKey(seckey).toPublicKey().mapConvert(PublicKey)
```
mratsim, line 47: Not sure where `mapConvert` came from but checking if something is the default value should be either `isDefault` or `isZero` or in functional approach `all(it == 0)`

eth/keys.nim
```diff
@@ -6,105 +6,127 @@
...
 
-proc signMessage*(seckey: PrivateKey,
-                  data: openarray[byte]): Signature {.inline.} =
-  ## Sign message of arbitrary length `data` using private key `seckey`.
-  let hash = keccak256.digest(data)
-  if signRawMessage(hash.data, seckey, result) != EthKeysStatus.Success:
-    raise newException(EthKeysException, ekErrorMsg())
+proc fromRaw*(T: type PrivateKey, data: openArray[byte]): SkResult[PrivateKey] =
+  SkPrivateKey.fromRaw(data).mapConvert(PrivateKey)
```
mratsim, line 50: same remark on `mapConvert`

Furthermore, since we are working on private key it is absolutely forbidden to return early like the previous implementation

```Nim
proc isZeroKey*(seckey: PrivateKey): bool =
  ## Check if private key `seckey` contains only 0 bytes.
  result = true
  for i in seckey.data:
    if i != byte(0):
      result = false
```
which is not constant-time and operating on secret data

eth/keys.nim
```diff
@@ -6,105 +6,127 @@
...
+
+proc verifyMessage*(data: openarray[byte], hash: MDigest[256]): bool =
   ## Verify binary data blob `data` has properly signed hash `hash`.
-  result = verifyMessage(data, hash.data)
+  let signature = SkRecoverableSignature.fromRaw(data)
+
+  signature.isOk() and recover(signature[], SkMessage(data: hash.data)).isOk()
+
+proc verifyMessage*(data: openarray[byte], message: openarray[byte]): SkResult[void] =
+  ## Verify binary data blob `data` has properly signed message `message`.
```
mratsim, line 86: I'm not too sure if the ECDSA spec proposes a name for this but the deterministic ECDSA at https://tools.ietf.org/html/rfc6979#section-2.2 doesn't.

Barring that I would be consistent with the BLS signature scheme spec: https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-01#section-2.5

`verify` is to check that a signature is valid for a message under a public key so I would just call this `verify`

eth/keys.nim
```diff
@@ -6,105 +6,127 @@
...
+  SkPublicKey.fromRaw(d).mapConvert(PublicKey)
+
+proc fromHex*(T: type PublicKey, data: string): SkResult[PublicKey] =
+  try:
+    # TODO strip string?
+    T.fromRaw(hexToSeqByte(data))
+  except CatchableError:
+    err("keys: cannot parse eth public key")
+
+proc random*(t: type KeyPair): KeyPair =
```
mratsim, line 75: The BLS draft proposes `keyGen` this would make it consistent with the BLS scheme: 

https://github.com/status-im/nim-blscurve/blob/9ce0266e0a56e990b68d29d199080b16ae6288de/blscurve/bls_signature_scheme.nim#L451

eth/keys.nim
```diff
@@ -6,105 +6,127 @@
...
+  nimcrypto/hash, nimcrypto/keccak, ./keys/secp, sequtils,
+  stew/[byteutils, objects, result], strformat
+
+export secp, result
 
 type
-  EthKeysStatus* = enum
-    Success,  ## Operation was successful
-    Error     ## Operation failed
+  PrivateKey* = distinct SkPrivateKey
```
mratsim, line 22: Both the underlying secp256k1 library and IETF specs (deterministic ECDSA https://tools.ietf.org/html/rfc6979 and BLS https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-01#section-2.3) calls it secret key.

The NIST standard and Ethereum calls it private key.

In the BLS signature scheme I went with secret key, makes it possible to use `sk` and `pk` instead of having 2 `pk` for private/public.

Open question.

eth/keys.nim
arnetheduck: `mapConvert` is a `Result` utility to convert using conversion operator, basically to change to the distinct type: https://github.com/status-im/nim-stew/pull/23/files#diff-5c66d6831b4e2a35d6aa76b2475cd161R239

eth/keys.nim
arnetheduck: well, `keys.nim` is intended to work with the "ethereum flavour" of keys so that naming should apply. 

In general, I'd prefer using secret key as well - I'll switch `secp.nim` to use `SecretKey` while `keys.nim` will "translate" that to the ethereum name

eth/keys.nim
arnetheduck: yeah I'd tend to agree - let's do a round of renaming stuff in `keys.nim` in the next PR - in fact, I think I'll just remove all non-deprecated functions from here to make it clear that these do not exist  yet

eth/keys.nim
arnetheduck: we have a half-baked Nim convention that uses `toTypeName` for these things.. at some point, I'm hoping to introduce something like generic explicit type converters akin to the To/From traits in rust - maybe `proc from(T: type X, v: Y): T` and `proc to(v: Y, T: type X)` - but until then, I think the nim convention we have already might serve us better

eth/keys.nim
arnetheduck: fixed, though I wonder if the zero check is sound at all?

eth/keys.nim
arnetheduck: the other name I thought would fit would be `init` per our Nim convention.. rust does this stuff in its standard `new` function - this would ensure that whoever uses the `init` convention gets a valid and working key / keypair instead of default-init-to-0 and weird behaviour (I wish there was a way to kill the zero-init construction)

mratsim: I guess we could propose a type-level `{.noinit.}` pragma so that tagged type are never zero-initialized.

arnetheduck: what I mean is that we need constructors to enforce at compile time that the default zero-init is never called.. `noinit` is just a way to introduce bugs at runtime - a sufficiently smart compiler has all the information it needs to inject it automatically where it's safe

arnetheduck:
Regarding naming:

If we go with `sign`, `verify` etc, how do we differentiate functions that take a hash vs functions that take an arbitrary data blob (and do the hash operation internally)?

We can go by type - the former can take an `MDigest` while the latter take an `openArray[byte]` but this is slightly dangerous in that it's a fairly fine line between the two..

eth/keys/secp.nim
```diff
@@ -0,0 +1,423 @@
...
+
+proc random*(T: type SkKeyPair): SkResult[T] =
+  ## Generates new random key pair.
+  let seckey = ? SkSecretKey.random()
+  ok(T(
+    seckey: seckey,
+    pubkey: seckey.toPublicKey().expect("random key should always be valid")
+  ))
+
+proc `==`*(lhs, rhs: SkSecretKey): bool =
```
cheatfate, line 326: This operations should not be present, because there no such use-case for secret keys.

arnetheduck: pushed

mratsim:
> Regarding naming:
> 
> If we go with `sign`, `verify` etc, how do we differentiate functions that take a hash vs functions that take an arbitrary data blob (and do the hash operation internally)?
> 
> We can go by type - the former can take an `MDigest` while the latter take an `openArray[byte]` but this is slightly dangerous in that it's a fairly fine line between the two..

In signature schemes `sign`, `verify` have the following parameters:
- A "cipher suite" which is a combination of hash function + tags + usage defined parameters, for example SHA256. In our case this is hardcoded by secp256k1 but it could be a typedesc or a context parameter.
- A message as a sequence of bytes
- A secret key for signing or a public key for verify.

i.e. The hashing is internal. Exposing a hashed message is an implementation defined optimization so it should be an opaque type. Changing the definition from

```Nim
  SkMessage* = MDigest[SkMessageSize * 8]
    ## Message that can be signed or verified
```

to

```Nim
  SkMessage* =  object
    ## Message that can be signed or verified
    # blob is private to enforce proper API usage
    blob: MDigest[SkMessageSize * 8]
```

should be safe enough.

arnetheduck:
> hardcoded by secp256k1 
not really - it'll take any 32-byte message

> A message as a sequence of bytes

well, `secp` interface operats on 32 bytes of whatever, and though a hash seems reasonable, it's not required

PR #209, opened 2020-04-02
keys: make hex parsing more lenient

arnetheduck:
also clean up keys a little


arnetheduck:
turns out strict parsing wasn't an issue, so closing this branch - will come back with some better error checks soon instead

PR #210, opened 2020-04-03
secp: handle invalid keys better

arnetheduck:
we can't guarantee with the type system that invalid keys don't exist,
so we have to introduce error handling for it


arnetheduck:
merging this to avoid conflicts - if there are comments, add them anyway and I'll address in separate PR's

PR #213, opened 2020-04-04
keys: get rid of deprecated calls

arnetheduck:
the mechanical changes needed to move to new keys api


arnetheduck:
same - comments welcome, merging to avoid conflicts - will fix in second round

PR #217, opened 2020-04-09
add goerli bootnodes

jangko:


jangko:
the CI failures because of #218

PR #220, opened 2020-04-11
secp: adapt to secp256k1 update

arnetheduck:


jangko:
the `DateTime` issue is fixed in #218.

arnetheduck:
thanks @jangko !

PR #227, opened 2020-04-19
Rlp experimental

arnetheduck:
removes experimental feature and most of the range library usage


zah:
As it stands, this code will result in millions of new allocations and `memcpy` calls during Eth1 syncing. RLP and the cheap-to-slice bit ranges are used heavily during Trie traversal which is one of the low-level operations invoked thousands of times during every Eth1 state transition.

Perhaps the problems can be fixed with some careful use of the `seq.shallow` API, but this needs to be verified more carefully.

Personally, I don't think we should introduce so significant inefficiencies only to get slightly safer language constructs. After all, the RLP library has gone through significant amount of fuzzing which is another viable way to gain confidence in the code.

arnetheduck:
the problem would not lie in RLP but rather in the proliferation of shallow copies through RLP copies outside of the rlp code itself - for auditing, we would want to minimize surface like this.

I have an ugly short-term fix in mind for the bit ranges in case this turns out to be a real problem which is based on using a `HSlice`-like type and operating on openarray instead - it's used in so few places that it's fine for now, but really, the bigger problem here is how the code calls RLP which should be fixed by your planned faststreams update - ie `if rlp.listItem(0).isBlob: rlp.listItem(0).getBytes` will never approach "fast" no matter how many shallowcopy tricks you perform - it's the usage of the code that needs fixing and this branch will merely bring to light where.

zah:
I would recommend at least running the following test before and after the PR and to see what the impact is:

https://github.com/status-im/nim-eth/blob/master/tests/trie/test_hexary_trie.nim#L55-L74

The test is not enabled by default, because it takes some time to run. You need to download the file mentioned in the comments.

arnetheduck:
new code:
```
[arnetheduck@tempus nim-eth]$ time /home/arnetheduck/status/nimbus/vendor/nim-eth/tests/trie/test_hexary_trie "genesis hash"

[Suite] hexary trie
  [OK] genesis hash

real	0m0,442s
user	0m0,430s
sys	0m0,010s
```
old:
```
[arnetheduck@tempus nim-eth]$ time /home/arnetheduck/tmp/nimbus/vendor/nim-eth/tests/trie/test_hexary_trie "genesis hash"

[Suite] hexary trie
  [OK] genesis hash

real	0m0,489s
user	0m0,471s
sys	0m0,017s
```

arnetheduck:
also, my impression was that eth1 syncing bottleneck (by far) is disk access - that said, rewriting the way path lookups are done would help everything here - in fact, most of it doesn't need a seq at all. 

Using a bitrange for example is really bad from a performance point of view - as such, I didn't really consider that the code was written with performance in mind - the overhead that the range introduces on a typical bit path is quite bad (ie if bit paths are up to 20 bytes, a bitrange object is.. 60 - 24 for the seq, 16 for the counters and 20 for the path data - this is ridiculously expensive + the refcounting)

arnetheduck:
```
+   91,12%     0,00%  test_hexary_tri  test_hexary_trie  [.] _start                                                                                                 ▒
+   91,12%     0,00%  test_hexary_tri  libc-2.30.so      [.] __libc_start_main                                                                                      ▒
+   91,12%     0,00%  test_hexary_tri  test_hexary_trie  [.] main                                                                                                   ▒
+   91,12%     0,00%  test_hexary_tri  test_hexary_trie  [.] NimMain                                                                                                ▒
+   91,10%     0,00%  test_hexary_tri  test_hexary_trie  [.] NimMainModule                                                                                          ▒
+   84,78%     0,06%  test_hexary_tri  test_hexary_trie  [.] put__17uWyyfDECoo0rbpmmBkZA                                                                            ◆
+   81,65%     0,06%  test_hexary_tri  test_hexary_trie  [.] put__u07y8aavCxWeCwphW9c1y9aA                                                                          ▒
+   69,91%     0,40%  test_hexary_tri  test_hexary_trie  [.] mergeAt__h589cO0vgAVWT4LyYUqeBiw                                                                       ▒
+   65,76%     0,06%  test_hexary_tri  test_hexary_trie  [.] mergeAtAux__tlpSdBOBTXwiFey0qBuwXQ                                                                     ▒
+   55,25%     4,25%  test_hexary_tri  test_hexary_trie  [.] digest__kUuxyH0Wbp0DnT9anHK9b9c2g                                                                      ▒
+   53,65%     0,00%  test_hexary_tri  test_hexary_trie  [.] mergeAt__NzzpWrk7yTDoNKuEDNqDQA                                                                        ▒
+   43,83%    15,72%  test_hexary_tri  test_hexary_trie  [.] keccakTransform__sVKJmssKMutEY1QO9bMnrnQkeccak                                                         ▒
+   28,83%     0,00%  test_hexary_tri  test_hexary_trie  [.] update__MiO1UCidJInmJ0SMkYUExQkeccak (inlined)                                                         ▒
+   24,01%     0,00%  test_hexary_tri  test_hexary_trie  [.] appendAndSave__PPSr29ah9aQD89bLAaQYTIrHA (inlined)                                                     ▒
+   23,72%     0,00%  test_hexary_tri  test_hexary_trie  [.] dbPut__Y1s44fJdT9aGp7cikCyHzRg (inlined)                                                               ▒
+   21,76%     0,00%  test_hexary_tri  test_hexary_trie  [.] keccak__VHBLaRIyZExhKFxxkgcCMQ (inlined)                                                               ▒
+   20,18%     0,00%  test_hexary_tri  test_hexary_trie  [.] keccak__VHBLaRIyZExhKFxxkgcCMQ (inlined)                                                               ▒
+   19,49%     0,00%  test_hexary_tri  test_hexary_trie  [.] finish__P4Gs9bIYxb2bVUt0SYoDVmQkeccak (inlined)                                                        ▒
+   19,49%     0,00%  test_hexary_tri  test_hexary_trie  [.] finish__vDb2CouTV34uwH9b5pTy74Akeccak (inlined)                                                        ▒
+   11,54%     0,24%  test_hexary_tri  test_hexary_trie  [.] append__7c9bAIt0G0h7H89b5CUmtIww                                                                       ▒
+   10,75%     0,00%  test_hexary_tri  test_hexary_trie  [.] dbPut__Y1s44fJdT9aGp7cikCyHzRg (inlined)                                                               ▒
+    9,58%     0,00%  test_hexary_tri  test_hexary_trie  [.] RHOPI__7qP3aIAGgrP4LeHFZAvPRQkeccak (inlined)                                                          ▒
+    9,26%     0,00%  test_hexary_tri  test_hexary_trie  [.] keccak__VHBLaRIyZExhKFxxkgcCMQ (inlined)                                                               ▒
+    7,39%     4,63%  test_hexary_tri  test_hexary_trie  [.] add__NYeR1yR6tUXiD6WGihr9cpw                                                                           ▒
+    7,28%     0,06%  test_hexary_tri  test_hexary_trie  [.] appendRawBytes__pxdeWX5a8lLVwyciPIo0mA_2                                                               ▒
+    7,10%     0,00%  test_hexary_tri  test_hexary_trie  [.] rawNewObj__ehkAaLROrd0Hc9aLROWt1nQ (inlined)                                                           ▒
+    6,70%     6,70%  test_hexary_tri  test_hexary_trie  [.] burnMem__haAOjkJjOuRy8ux9bV57dXw                                                                       ▒
+    6,64%     0,00%  test_hexary_tri  test_hexary_trie  [.] clear__9crVzInbDoQCnwlFos9bhraw_2keccak (inlined)                                                      ▒
+    6,64%     0,00%  test_hexary_tri  test_hexary_trie  [.] burnMem__3kHY0BwnzMmCH3xJFhlgPAutils (inlined)                                                         ▒
+    4,99%     2,68%  test_hexary_tri  test_hexary_trie  [.] genericResetAux__9bFXTAJ8caOL4Ioy2o3B3aw                                                               ▒
+    4,97%     0,17%  test_hexary_tri  test_hexary_trie  [.] genericSeqAssign                                                                                       ▒
+    4,91%     0,00%  test_hexary_tri  test_hexary_trie  [.] genericAssign (inlined)                                                                                ▒
```

guess the hexary tree test might be too small to stress the malloc - there is a 5% perf cost to the assigns

PR #230, opened 2020-04-23
port kvstore from nim-beacon-chain

arnetheduck:
* remove old database backends
* use kvstore in trie database


eth/db/kvstore.nim
```diff
@@ -0,0 +1,115 @@
...
+    ## Key-Value store virtual interface
+    obj: RootRef
+    putProc: PutProc
+    getProc: GetProc
+    delProc: DelProc
+    containsProc: ContainsProc
+
+template put*(db: KvStoreRef, key, val: openArray[byte]): KvResult[void] =
+  ## Store ``value`` at ``key`` - overwrites existing value if already present
+  db.putProc(db.obj, key, val)
```
zah, line 42: Double evaluation of the `db` parameter. You can go for the usual fix:

```nim
template put*(dbParam: KvStoreRef, key, val: openArray[byte]): KvResult[void] =
  ## Store ``value`` at ``key`` - overwrites existing value if already present
  let db = dbParam
  db.putProc(db.obj, key, val)
```

zah: Ditto for the rest of the file.

eth/db/kvstore_sqlite3.nim
```diff
@@ -0,0 +1,158 @@
...
+  ./kvstore
+
+export kvstore
+
+type
+  SqStoreRef* = ref object of RootObj
+    env: ptr sqlite3
+    selectStmt, insertStmt, deleteStmt: ptr sqlite3_stmt
+
+template checkErr(op, cleanup: untyped) =
```
zah, line 17: How come you never use the `cleanup` parameter in this file? None of the sqlite calls below require cleanup?

eth/db/kvstore.nim
```diff
@@ -0,0 +1,115 @@
...
+  KvStoreRef(
+    obj: x,
+    putProc: putImpl[T],
+    getProc: getImpl[T],
+    delProc: delImpl[T],
+    containsProc: containsImpl[T]
+  )
+
+proc get*(db: MemStoreRef, key: openArray[byte], onData: DataProc): KvResult[bool] =
+  let key = @key
```
zah, line 87: This should get the same TODO comment as the two functions below.

eth/trie/hexary.nim
```diff
@@ -564,7 +564,7 @@ proc mergeAt(self: var HexaryTrie, orig: Rlp, origHash: KeccakHash,
   if orig.isEmpty:
     return origWithNewValue()
 
-  doAssert orig.isTrieBranch
+  doAssert orig.isTrieBranch, $orig
```
zah, line 567: Is `doAssert` lazy with regard to the string parameter? `$(rlp)` is not cheap

zah: I've just checked that it is, so it's OK

eth/db/kvstore_sqlite3.nim
arnetheduck: https://github.com/status-im/nim-eth/pull/230/files#diff-50ae87a573f6f1c02f8e6e155b691ba8R115
https://github.com/status-im/nim-eth/pull/230/files#diff-50ae87a573f6f1c02f8e6e155b691ba8R148

poor man's destructors I guess - but without move support and SqStore being `ref`..

eth/trie/hexary.nim
arnetheduck: yeah, I checked as well ;)

Issue #185, opened 2020-03-04
Improve/fix Error handling in discovery v5


kdeme:
for example [here](https://github.com/status-im/nim-eth/blob/master/eth/p2p/discoveryv5/protocol.nim#L154), [here](https://github.com/status-im/nim-eth/blob/master/eth/p2p/discoveryv5/protocol.nim#L359) [and here](https://github.com/status-im/nim-eth/blob/master/eth/p2p/discoveryv5/protocol.nim#L373). But there are more locations with problems

kdeme:
This has been handled in #240.

Of course there are still the many TODO items regarding raises pragma and async: Issue is created for that here: https://github.com/status-im/nim-chronos/issues/98
Related follow-up issue regarding error handling of the send call: https://github.com/status-im/nim-eth/issues/241

PR #240, opened 2020-05-26
Discv5 eh2

kdeme:


eth/p2p/discoveryv5/discovery_db.nim
```diff
@@ -22,13 +24,16 @@ proc makeKey(id: NodeId, address: Address): array[keySize, byte] =
   copyMem(addr result[1], unsafeAddr id, sizeof(id))
   case address.ip.family
   of IpAddressFamily.IpV4:
-    copyMem(addr result[sizeof(id) + 1], unsafeAddr address.ip.address_v4, sizeof(address.ip.address_v4))
+    copyMem(addr result[sizeof(id) + 1], unsafeAddr address.ip.address_v4,
```
arnetheduck, line 27: `import stew/byteutils` + `result[pos..pos+sizeof(v4)] = address.ip.address_v4`

eth/p2p/discoveryv5/encoding.nim
```diff
@@ -42,7 +42,7 @@ type
     response*: seq[byte]
 
   DecodeError* = enum
-    HandshakeError = "discv5: handshake failed"
+    HandshakeError = "discv5: handshake failed",
```
arnetheduck, line 45: can just remove all `,` instead? I think we usually don't put them there

eth/p2p/discoveryv5/node.nim
```diff
@@ -1,64 +1,59 @@
 import
-  std/[net, hashes], nimcrypto, stint, chronicles,
-  types, enr, eth/keys, ../enode
+  std/[net, hashes], nimcrypto, stint, chronos,
+  eth/keys, enr
 
 {.push raises: [Defect].}
 
 type
+  NodeId* = UInt256
```
arnetheduck, line 8: is the discv5 nodeid linked to eth1 nodeid in the spec, or just conveniently compatible?

eth/p2p/discoveryv5/discovery_db.nim
arnetheduck: or perhaps introduce some other range-checked helper that ensures there's space enough in the array like the `[]` operator does, but with a single argument

eth/p2p/discoveryv5/protocol.nim
```diff
@@ -152,15 +158,31 @@ func privKey*(d: Protocol): lent PrivateKey =
...
+        # interface underneath gets disconnected or similar.
+        # TODO: Should this kind of error be propagated upwards? Probably, but
+        # it should not stop the process as that would reset the discovery
+        # progress in case there is even a small window of no connection.
+        # One case that needs this error available upwards is when revalidating
+        # nodes. Else the revalidation might end up clearing the routing tabl
+        # because of ping failures due to own network connection failure.
+        debug "Discovery send failed", msg = f.readError.msg
+  except Exception as e:
+    # General exception still being raised from Chronos.
```
arnetheduck, line 178:
```suggestion
    # TODO General exception still being raised from Chronos.
```

eth/p2p/discoveryv5/protocol.nim
```diff
@@ -177,16 +199,19 @@ proc isWhoAreYou(d: Protocol, packet: openArray[byte]): bool =
...
     result = d.whoareyouMagic == packet.toOpenArray(0, magicSize - 1)
 
-proc decodeWhoAreYou(d: Protocol, packet: openArray[byte]): Whoareyou =
+proc decodeWhoAreYou(d: Protocol, packet: openArray[byte]):
+    Whoareyou {.raises: [RlpError].} =
   result = Whoareyou()
   result[] = rlp.decode(packet.toOpenArray(magicSize, packet.high), WhoareyouObj)
 
-proc sendWhoareyou(d: Protocol, address: Address, toNode: NodeId, authTag: AuthTag) =
+proc sendWhoareyou(d: Protocol, address: Address, toNode: NodeId,
```
arnetheduck, line 207: this looks pretty odd though - it looks more like this should be an `{.async.}` proc - or a pure proc that registers a handshake without any async component to it - the dangling callback it creates is hard to reason about

eth/p2p/discoveryv5/node.nim
kdeme: It is not explicitly linked afaik. Discv5 specs just defines it as a 32-byte identifier.
In practise however it is the same currently.

I'd guess it is written like this because in theory ENR can have several identity schemes.

eth/p2p/discoveryv5/protocol.nim
kdeme: Yeah, the callbacks, both here and when registering the pending requests are there to clean up the handshakes/pending requests in case of time-out on reply.

I agree it makes the code more hard to reason about, but I had to add 2 of them to fix some bugs (there was already 1 callback doing cleanup). At that time I thought also about just cleaning them up in some loop called now and then, but I wasn't so sure if that was an actual improvement.

That aside, I might (have to) make the call {.async.} anyhow as I'm thinking about moving send errors up in the `send` proc with a `Future[Result[..]]` type. Although I could do that also only for `sendMessage`, as that information in `sendWhoareyou` will only be discarded (or just used for logging) anyhow, and working with results and await is a bit cumbersome right now.

kdeme:
This should improve on https://github.com/status-im/nim-eth/issues/185, so I think I will close that one and create a new issue regarding all the TODOs. Which are mostly about the async-raises fun but also about the `send` proc eating its errors in that async callback.

kdeme:
So, tests actually fail but `require` makes it pass.
Fun fact: The test that fails, basically had same behaviour before, but it wasn't spotted as a sequence was returned that was empty.

PR #245, opened 2020-06-02
Add discoveryv5 debugging tool: dcli

kdeme:
Little tool I've been using to debug discovery v5.

First, is nim-eth a good location? Or should we drop it in the nim-beacon-chain repo? The latter would allow it to be certain to work for the repo versioning of nim-beacon-chain.
If here, I'll add a `bin` in eth.nimble. But currently that gives issues, several wrong paths + what looks like some chronicles difference between latest version and version that nim-beacon-chain uses (that import of `topics_registry` I think).

ENR tooling could also be added in this application, or create yet another tool for that.
Thoughts?

### DIFF
--- a/eth/p2p/discoveryv5/dcli.nim
+++ b/eth/p2p/discoveryv5/dcli.nim
@@ -37,16 +37,20 @@ type
       discard
     of ping:
       pingTarget* {.
+        argument
         desc: "ENR URI of the node to a send ping message"
         name: "node" .}: Node
     of findnode:
-      findNodeTarget* {.
-        desc: "ENR URI of the node to send a findNode message"
-        name: "node" .}: Node
       distance* {.
         defaultValue: 255
         desc: "Distance parameter for the findNode message"
         name: "distance" .}: uint32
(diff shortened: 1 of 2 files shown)


eth/p2p/discoveryv5/dcli.nim
```diff
@@ -0,0 +1,147 @@
...
+
+    case cmd* {.
+      command
+      defaultValue: noCommand }: DiscoveryCmd
+    of noCommand:
+      discard
+    of ping:
+      pingTarget* {.
+        desc: "ENR URI of the node to a send ping message"
+        name: "node" .}: Node
```
zah, line 41: Wouldn't this work better as an `argument`? This is a confutils pragma that allows the following usage:

```
dcli ping enr://121312...
```

eth/p2p/discoveryv5/dcli.nim
```diff
@@ -0,0 +1,147 @@
...
+    of findnode:
+      findNodeTarget* {.
```
zah, line 43: Ditto

PR #248, opened 2020-06-09
Fix discovery after introduction of ValidIpAddress

kdeme:
will need https://github.com/status-im/nim-stew/pull/42

### DIFF
--- a/eth/p2p/discoveryv5/node.nim
+++ b/eth/p2p/discoveryv5/node.nim
@@ -1,5 +1,5 @@
 import
-  std/[net, hashes], nimcrypto, stint, chronos,
+  std/hashes, nimcrypto, stint, chronos, stew/shims/net,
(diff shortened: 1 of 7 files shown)


eth/p2p/discoveryv5/node.nim
```diff
@@ -32,7 +32,8 @@ proc newNode*(r: Record): Result[Node, cstring] =
   let tr = ? r.toTypedRecord()
   if tr.ip.isSome() and tr.udp.isSome():
     let
-      ip = IpAddress(family: IpAddressFamily.IPv4, address_v4: tr.ip.get())
+      ip = ValidIpAddress.init(
+        IpAddress(family: IpAddressFamily.IPv4, address_v4: tr.ip.get()))
```
zah, line 36: You can use `ipv4(tr.ip.get())` here

PR #251, opened 2020-06-10
Add predicate filter option for randomNodes

kdeme:


eth/p2p/discoveryv5/node.nim
```diff
@@ -32,8 +32,7 @@ proc newNode*(r: Record): Result[Node, cstring] =
   let tr = ? r.toTypedRecord()
   if tr.ip.isSome() and tr.udp.isSome():
     let
-      ip = ValidIpAddress.init(
-        IpAddress(family: IpAddressFamily.IPv4, address_v4: tr.ip.get()))
+      ip = ValidIpAddress.init(ipv4(tr.ip.get()))
```
zah, line 35: `ipv4` returns a `ValidIpAddress` (no need for the extra `init` call`)

eth/p2p/discoveryv5/routing_table.nim
```diff
@@ -248,25 +248,28 @@ proc nodeToRevalidate*(r: RoutingTable): Node =
     if b.len > 0:
       return b.nodes[^1]
 
-proc randomNodes*(r: RoutingTable, count: int): seq[Node] =
-  var count = count
+proc randomNodes*(r: RoutingTable, maxAmount: int,
+    pred: proc(x: Node): bool {.closure.} = nil): seq[Node] =
```
zah, line 252: It's better if all proc types are specified more precisely - `gcsafe, raises: [Defect]` as a minimum

arnetheduck: `gcsafe, noSideEffect, raises: [Defect]` really - which becomes quite an impressive list of junk/noise one needs to add just to get the ability to do basic reasoning about it.. I wonder if this shouldn't be added to some generic `callback` annotation which outlines "minimal" requirements for callbacks

zah: `noSideEffect` is not strictly mandatory here as this function is a regular proc. It would impose some limitations on what can be written in the predicate, perhaps for a good reasons, but it's definitely arguable.

arnetheduck: so is gcsafe, and so is raises ;) they all relate to the ability to reason about what will happen by constraining the space of options

eth/p2p/discoveryv5/routing_table.nim
```diff
@@ -248,25 +248,28 @@ proc nodeToRevalidate*(r: RoutingTable): Node =
...
-  result = newSeqOfCap[Node](count)
+  result = newSeqOfCap[Node](maxAmount)
   var seen = initHashSet[Node]()
 
   # This is a rather inneficient way of randomizing nodes from all buckets, but even if we
   # iterate over all nodes in the routing table, the time it takes would still be
   # insignificant compared to the time it takes for the network roundtrips when connecting
   # to nodes.
-  while len(seen) < count:
+  while len(seen) < maxAmount:
```
arnetheduck, line 267: won't this loop forever if the space of nodes that match the filter is less than `r.len`?

eth/p2p/discoveryv5/routing_table.nim
arnetheduck: I'm guessing discv5 doesn't have a query that can include filters, so that one can let the remote end do the filtering?

kdeme: Indeed. You can remove all this and just filterIt in nbc if you want it that way. (right now you only avoid adding it to a seq and then removing again, not much of a deal). 
Just thought it was slightly more practical, but perhaps not.

However, there is talk about adding a `FINDNODEATT` which specifically requests nodes with certain ENR attribute. But not sure if that will get in discv5.

kdeme: But the point of leaving the `pred` option open, is because I thought we eventually might want to do custom filtering based on certain bits in the `attnets` field. But yeah again, you can just filterIt in nbc directly also.

eth/p2p/discoveryv5/routing_table.nim
kdeme: Hmm no, because the `seen.incl(node)` is done no matter what the filter returns (if that is what you meant?). This is also why I changed the name from `count` to `maxAmount`, in an attempt to make it more clear what this call does.
As similar issue already existed, e.g. when the routing table had less nodes than the amount you requested (also without filter). So it has always been a `maxAmount` basically.

PR #253, opened 2020-06-15
Stricter error handling

zah:
Handle the changes introduced in status-im/nim-serialization#27

### DIFF
--- a/eth/common/eth_types_json_serialization.nim
+++ b/eth/common/eth_types_json_serialization.nim
@@ -2,11 +2,16 @@ import
   times, net,
   json_serialization, nimcrypto/[hash, utils], eth_types
 
-proc writeValue*(w: var JsonWriter, a: MDigest) {.inline.} =
+{.push raises: [SerializationError, IOError, Defect].}
+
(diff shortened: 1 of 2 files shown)


eth/common/eth_types_json_serialization.nim
```diff
@@ -2,11 +2,16 @@ import
...
-proc writeValue*(w: var JsonWriter, a: MDigest) {.inline.} =
+{.push raises: [SerializationError, IOError, Defect].}
+
+proc writeValue*(w: var JsonWriter, a: MDigest) =
   w.writeValue a.data.toHex(true)
 
 proc readValue*(r: var JsonReader, a: var MDigest) {.inline.} =
-  a = fromHex(type(a), r.readValue(string))
+  try:
+    a = fromHex(type(a), r.readValue(string))
```
arnetheduck, line 12: `fromHex` from std lib btw is more lenient than the `byteutils` variant that we have - afair, it's more lenient than json in general as well, meaning using it here is not really compliant...

arnetheduck: ah nevermind this is the mdigest version

PR #257, opened 2020-06-20
ignoreSignalsInThread()

stefantalpalaru:


mratsim:
That seems like a Pandora box. Assuming we start logging with signals to manage logrotation (cc @zah) or we add any new features that require signal do we have to edit that file?

Will we need to transform that into some kind of supervisor where we register signal handlers?

stefantalpalaru:
https://www.man7.org/linux/man-pages/man7/signal.7.html :

>A process-directed signal may be delivered to any one of the threads that does not currently have the signal blocked.  If more than one of the threads has the signal unblocked, then the kernel chooses an arbitrary thread to which to deliver the signal.

So, this here is just common sense. I don't want to process signals in the thread I'm creating, so I'm masking them.

PR #260, opened 2020-06-24
More rand fixes

narimiran:
Sorry for yet another PR, I missed these instances in the previous PR.


kdeme:
No problem, thanks for cleaning these up!

Issue #206, opened 2020-04-02
Improve (re)validation of nodes


kdeme:
Currently nodes are (re)validated by a periodic ping and removed from the routing table at first failure to reply.
This should be improved, e.g.:
1. No action should be taken on first failure. It should be marked `stale` after n (=5 ?) failures.
2. The replacement cache should be used: When a node goes stale, it should be replaced with a (non stale?) entry from the replacement cache, if there is one.
3. If there is no entry in the replacement cache, no immediate removal of the node should happen. Instead it should be removed when a new node for that bucket is added.

Considering that this replacement system (and kbucket ordering in general) is not really used currently (nodes only get added to it) and that there are no tests, it is possible that there are some hidden issues in the current code.
There also seems to be no current limitation on adding to the replacement cache.

Also the `nodeToRevalidate` should probably target least recently updated bucket.

kdeme:
Use of replacement cache is in place, and replacement cache is limited to 8 nodes.

`nodeToRevalidate` targeting least recently updated bucket has not shown to give any improvement on current testnets.
More tests are in place now.

1. and 3. are not currently done. 3. specifically might be an issue when the own connection drops or is just terrible. 

Issue for those is created here: https://github.com/status-im/nim-eth/issues/262

Issue #183, opened 2020-03-04
Liveness verified functionality


kdeme:
Nodes added in the routing table should have a liveness flag that is set after liveness is verified at least once.
Peers where liveness is not validated should not be used yet:
- Not forwarded in lookup 
- Perhaps also not provided when the local application asks for random peers?

kdeme:
Point 1 is added here: https://github.com/status-im/nim-eth/pull/261
Point 2 is decided not to do default.

PR #261, opened 2020-06-29
Add routing table metrics + tweaks + fixes

kdeme:
- routing table metrics + option in dcli
- only forward "seen" nodes on a findNode request
- setJustSeen & replace on ping AND findnode
- self lookup only at start
- revalidate 10x more
- use bitsPerHop (b) of 5
- small fix in resolve
- small fix in bucket split

### DIFF
--- a/eth/p2p/discoveryv5/dcli.nim
+++ b/eth/p2p/discoveryv5/dcli.nim
@@ -1,6 +1,6 @@
 import
   sequtils, options, strutils, chronos, chronicles, chronicles/topics_registry,
-  stew/byteutils, stew/shims/net, confutils,
+  stew/byteutils, confutils, confutils/std/net, metrics,
   eth/keys, eth/trie/db, eth/net/nat,
   eth/p2p/discoveryv5/[protocol, discovery_db, enr, node]
 
@@ -29,6 +29,26 @@ type
       desc: "Specify method to use for determining public address. " &
             "Must be one of: any, none, upnp, pmp, extip:<IP>."
       defaultValue: "any" .}: string
+
+    nodeKey* {.
+      desc: "P2P node private key as hex.",
+      defaultValue: PrivateKey.random().expect("Properly intialized private key")
+      name: "nodekey" .}: PrivateKey
+
+    metricsEnabled* {.
+      defaultValue: false
+      desc: "Enable the metrics server."
+      name: "metrics" .}: bool
+
+    metricsAddress* {.
+      defaultValue: ValidIpAddress.init("127.0.0.1")
+      desc: "Listening address of the metrics server."
+      name: "metrics-address" .}: ValidIpAddress
+
+    metricsPort* {.
+      defaultValue: 8008
+      desc: "Listening HTTP port of the metrics server."
+      name: "metrics-port" .}: Port
 
     case cmd* {.
       command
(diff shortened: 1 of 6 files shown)


kdeme:
This PR basically tackles https://github.com/status-im/nim-eth/issues/183 and https://github.com/status-im/nim-eth/issues/206.

Regarding liveness, I've only peers that are not "seen" yet are not forwarded in the FindNode requests as it should be.
They are still passed with `randomNodes` call though, as else providing peers at start-up will be much slower. (a custom `randomNodes` call could still filter these).

Regarding staleness, I've done some quick testing with allowing nodes with a failed request/response to stay in the routing table until after n failures. In the somewhat naive implementation of a simple counter, and the same revalidation code, this has sort of a negative effect.
Because of the extra pings required, the actual nodes that are "alive" take much longer to be "seen". A lot of time is wasted on stale nodes. And this is worse because the network seems to have still a lot of stale nodes in the DHT (This is probably due to a combination of some clients not cleaning up stale nodes fast and typically having a lot of stale nodes due to the nature of a testnet, this was tested on witti).
So an alternative solution (e.g. separate exp. back-off ping not obstructing current revalidation) would be good here, I'll create a ticket for that. As a result, nodes are currently still replaced (or removed) immediately on failure.

Here is the state on witti with the current settings, nodes in routing table versus "seen" nodes in routing table.
Screenshot_2020-06-29-current_state
The current setting of revalidation frequency is required to keep routing table not to have too many unseen nodes.

The setting of bitsPerHop (b) is chosen mostly based on how many nodes in the DHT, see following screenshot:
Screenshot_2020-06-29_withnoselflookup-max1000ms-60s-nostale-2-fixed-2-4-3-b
This is with settings of b=2, b=4, b=3. b=4 would probably still be fine. Either way, these can be lowered, it will just give less nodes in the routing table. But it should be noted that we currently use `randomNodes` in nbc to get our nodes, so the newly discovered nodes through lookups, that are not added because the table is e.g. full, will not be used.

The selflookup is not only done at startup for an initial list of "close" nodes. Running that continuously has not proven to increase the routing table seen nodes.

kdeme:
And here when running on altone. (These are tests where discovery is launched with only 1 bootstrap node)

Screenshot_2020-06-30 New dashboard - Grafana
You can see some continuously happening small drops. Some (most?) of these can possibly be avoided by better stale re-pinging, see comment above. But as we continuously do lookups, it doesn't matter much currently.

zah:
This looks like quite a complicated way to draw a dinosaur :P

LTGM

arnetheduck:
What's the cost of a) keeping a node in the table and b) pinging it?

arnetheduck:
also, an option might be to score nodes in the table - ie those that have "recently" been validated get a better score, so stuff like randomnodes returns high-score peers first and low-score peers only if there aren't that many

kdeme:
> What's the cost of a) keeping a node in the table and b) pinging it?

Direct cost of a) is just memory, ENR size (which is max 300 bytes) + (duplicated from enr) data stored in `Node` object. Of course there is a hidden cost in the sense that it possibly takes up a spot for another more useful node, in the case that this node is offline. As mentioned before, this could be changed by not using `randomNodes` but instead `lookupRandom` directly in the nbc networking code. I eventually want to test that better.

b) is just a request response. And if it is the first ping, a handshake + request response (If we move to an LRU cache for the secrets, possibly not just the first ping that does a handshake). Can't give numbers in time as I haven't benchmarked this, but should be small. I think go implementation benchmarked the handshake part.

kdeme:
> also, an option might be to score nodes in the table - ie those that have "recently" been validated get a better score, so stuff like randomnodes returns high-score peers first and low-score peers only if there aren't that many

Nodes are ordered on least recently seen in the bucket. So you could easily get the least recently validated nodes first if you want. And it would change all the time as long as the revalidation runs with a low interval, so perhaps enough randomisation. Not sure though. I guess it depends also on the application. If you want to waste more trial&error connections in favour of having maximum of nodes, or if you want to connect to the least possible but be more sure that it would work (e.g. for mobile).

PR #263, opened 2020-06-30
Return on getMessage failure and do address conversion afterwards

kdeme:


eth/p2p/discoveryv5/protocol.nim
```diff
@@ -428,11 +419,21 @@ proc processClient(transp: DatagramTransport, raddr: TransportAddress):
   except TransportOsError as e:
     # This is likely to be local network connection issues.
     error "Transport getMessage error", exception = e.name, msg = e.msg
+    return
```
arnetheduck, line 422: so this is better written as :+1: 
```
let buf = try: raddr.address()
  except TOE as e: ...
  excpet Exception as ...
```
which avoids this kind of bugs completely

eth/p2p/discoveryv5/protocol.nim
```diff
@@ -428,11 +419,21 @@ proc processClient(transp: DatagramTransport, raddr: TransportAddress):
...
   except Exception as e:
     if e of Defect:
       raise (ref Defect)(e)
     else: doAssert(false)
 
+  try:
```
arnetheduck, line 428: and this is `let ip = try: raddr.address() except: ...`

eth/p2p/discoveryv5/protocol.nim
kdeme: Cool, didn't know could do a return in that construction, I'll adjust.

PR #265, opened 2020-07-06
use bearssl rng throughout

arnetheduck:


arnetheduck:
> is now avoided.

strictly, it's moved from being spread out all across the codebase to a single location - the newRng function

> technical benefit

the main objective is to use a single method across the codebase, and bear happened to satisfy that criteria and already be included. technically, it could be wrapped in our own type / interface if we wanted to isolate it further (like `crypto.nim` does for bls in eth2) but this seemed onerous.

> terms of security

https://bearssl.org/gitweb/?p=BearSSL;a=blob;f=inc/bearssl_rand.h;h=0a9f544fc4c58a82805d9b759eb6b21d8a001511;hb=dda1f8a0c46e15b4a235163470ff700b2f13dcc5#l76

> never need to reseed

see above - also, since it's a single instance, it's easy to reseed periodically

kdeme:
> it could be wrapped in our own type / interface if we wanted to isolate it further

Yes, was about to suggest that but then didn't as I realized we would still depend on bearssl in nim-eth for using it in `dcli` or in the tests (if we want to set proper example). Nevertheless, it might still be a sound good idea, we already use it in nbc, but we might to not want to push to other projects.  But then again, those projects would also need a good rng, so might end up with bearssl again :).

eth/keys.nim
```diff
@@ -46,12 +46,29 @@ type
...
 
-proc random*(T: type PrivateKey): SkResult[T] =
-  SkSecretKey.random().mapConvert(T)
+template pubkey*(v: KeyPair): PublicKey = PublicKey(SkKeyPair(v).pubkey)
+template seckey*(v: KeyPair): PrivateKey = PrivateKey(SkKeyPair(v).seckey)
+
+proc newRng*(): ref BrHmacDrbgContext =
+  # You should only create one instance of the RNG per application / library
+  # Ref is used so that it can be shared between components
+  # TODO consider moving to bearssl
```
kdeme, line 57: Yeah, I think that would be better. Looking at the nbc code now that calls this, it looks weird and confusing

eth/keys.nim
```diff
@@ -46,12 +46,29 @@ type
...
+  var
+    seeder = brPrngSeederSystem(nil)
```
kdeme, line 59: Looks like this could also fail in theory

eth/keys.nim
arnetheduck: yeah - I've thought of refactoring bearssl a little, would do it then

arnetheduck:
> good rng

yes - and at the core of the matter lies that some of our key generation is tied to the bear rng explicitly so until that is rewritten, this is a freedom we can't have - I'd leave it for another day

arnetheduck:
> RNG in a high-level API

yes - different PR though - one thing to consider is that bear functions internally depend on this being the bear API specifically

Issue #233, opened 2020-05-02
Do not use stdlib.random for Discovery5


cheatfate:
We are still using stdlib.random in discovery5 code. Current version of nim stdlib's `random` module is implementation of xoroshiro128+.

> XorShift and Xoroshiro are LFSRs. LFSRs are a component of some hardware-implementation-focused algorithms. LFSRs are not secure on their own. They are linear, so their state (and thus future output) can be predicted from a few known output bits. Additional operations have to be used in stream ciphers that utilize LFSRs.
> 
> Mersenne twister is based on a generalized feedback shift register, so it has similar properties. (Note that generalized does not mean better. Algorithms recommended by Sebastiano Vigna are faster, smaller, and have better quality ouput.)

kdeme:
True, it is still used for selecting "random" nodes from the routing table.
More specifically here: https://github.com/status-im/nim-eth/blob/master/eth/p2p/discoveryv5/routing_table.nim#L267 , there is already a TODO note about it there.

I think it is less harmful (much more difficult to abuse) than on other locations, as the routing table itself already has some "randomness" to it. Hence the question form in the TODO, but sure, better be safe and change this.

Issue #271, opened 2020-07-09
Adjust approach of randomNodes to a more lookup based system for providing peers


kdeme:
Currently lookups are done automatically at specific interval.
Applications using discoveryv5 can then request nodes with `randomNodes`. This will select nodes from the routing table.
If wanted, `lookup` or `randomLookup` can still be called to provide potentially extra peers.

It might however be of interest to only use `randomLookup` (or `lookup`) to provide peers. Discovery could have a timer to be sure that a lookup is done every x minutes, in case it would not be trigger by the application.

EDIT: Some discussion and thinking has brought us to the conclusion (see comments below) that `randomNodes` in its current form is not only inefficient, but also likely to have security flaws. Additionally, the idea of using `randomLookup` directly does not seem to get much approval. The alternative idea is something proposed a while back with an additional structure holding latest discovered nodes through lookup, which would be triggered to provide these peers after a new lookup.

kdeme:
The interest might be two-folded:
- More usefull nodes might be passed to the application using them, especially in case of a small routing table (say bitsPerHop=1).
- It sounds like a more resilient way in case of a big amount of malicious nodes in the routing table. Say that the routing table is practically full, and contains a lot of malicious nodes, then `lookups` should still provide a decent amount of other nodes (at least if there are enough good peers still in the routing table).

kdeme:
On the topic of efficiency, suggestions have been made (mostly by @cheatfate) to work on improving the `randomNodes` call.
One such suggestion was to trigger an event when new nodes are added to the routing table, and to make `randomNodes` an `async proc` waiting on this event. Next to that, to return only nodes that were not returned in previous call (or perhaps even calls).

Giving this some thought, I believe this can introduce a security risk. Especially when the number of nodes in the routing table is still low or almost none, e.g. before or during first lookup. 
A malicious actor could make it so that the first set of nodes to be passed on for the application to be used, be a set of nodes that it controls (or at least a big part of them). And it could probably make the next sets biased.
- This could be done passively, e.g. if it would control a bootstrap node or manage to poison a bootstrap node with many of its nodes. And then if that node would be within the first ones used to do a lookup. But this would not be so easy achievable (as findNode requests are done in parallel by 3).
[...]

PR #272, opened 2020-07-10
Use brHmacDrbgGenerate for all random operations in discovery

kdeme:
Alright, this is an attempt to resolve https://github.com/status-im/nim-eth/issues/233.

Random is currently used for 3 things in discovery v5:
1. `randomNodes`, to get a random selection of nodes from the routing table (random node in random bucket)
2. `nodeToRevalidate`, to get a random bucket
3. `revalidateLoop`, to ping on an average interval, but not always the same interval.

I believe all 3 would be rather difficult to abuse with current implementation, but perhaps not impossible. And with the idea of not using stdlib random in general to set good example and avoid mistakes, I decided to remove it. 

The idea of 1. is that it could perhaps be abused if an attacker manages to fill our routing table with a specific set of nodes, with node ids that would then be triggered to be selected in the first rounds of `randomNodes`. This is quite difficult because there is more randomization going on then just the `random` call. Discovery itself has a difficulty to it that you have to create specific node ids to be placed in the right buckets. And then the locations of your nodes in the buckets will highly depend on the order of network calls made. But if such effort could be done, it would be troublesome, yes.

Not sure how 2. could be abused, but it could be used to have your nodes being revalidated "as late as possible", but eventually they will be revalidated and removed if not reachable.

Then number 3., not sure why it was placed there originally (I noticed the go code also does this). I would assume it is against an attacker for making revalidating (pinging) fail each time in case the interval is fixed by timing some network congestion on that moment. Not sure about other scenarios.

Regarding the random helper calls, aside from the rng, I've based myself on the stdlib implementation mostly, however I noticed what I believe to be a bug in the stdlib at the location where possible modulo bias is removed. Please have a good look at this, I don't want to worsen things. :)
I've tested it with a `randMax=255` (uint8) version and it "seems" to be correct now. (testing with uint64 version would not easily give you practical differences).

Aside from these changes, 1. and 2. could probably also avoid using random in general:
1. I've thought about passing a random node ID and then selecting all neighbours, but I think this too is similarly playable. Perhaps better would be just to not use that call at all and start doing `loopupRandom`, see: https://github.com/status-im/nim-eth/issues/271. It will probably be more resilient in general, especially in case you use smaller routing tables (smaller b factor, less bucket splits, less nodes)
2. Here we could select the bucket that was last validated, I guess the random part is more of a simpler convenient way.

### DIFF
--- a/eth/p2p/discoveryv5/random2.nim
+++ b/eth/p2p/discoveryv5/random2.nim
@@ -1,0 +1,22 @@
+import bearssl
+
+## Random helpers: similar as in stdlib, but with BrHmacDrbgContext rng
+# TODO: Move these somewhere else?
+const randMax = 18_446_744_073_709_551_615'u64
+
+proc rand*(rng: var BrHmacDrbgContext, max: Natural): int =
+  if max == 0: return 0
+
+  var x: uint64
+  while true:
+    brHmacDrbgGenerate(addr rng, addr x, csize_t(sizeof(x)))
+    if x < randMax - (randMax mod (uint64(max) + 1'u64)): # against modulo bias
+      return int(x mod (uint64(max) + 1'u64))
+
+proc sample*[T](rng: var BrHmacDrbgContext, a: openarray[T]): T =
+  result = a[rng.rand(a.high)]
+
+proc shuffle*[T](rng: var BrHmacDrbgContext, a: var openarray[T]) =
+  for i in countdown(a.high, 1):
+    let j = rng.rand(i)
+    swap(a[i], a[j])
--- a/eth/p2p/discoveryv5/protocol.nim
+++ b/eth/p2p/discoveryv5/protocol.nim
@@ -73,10 +73,11 @@
 ## This might be a concern for mobile devices.
 
 import
-  std/[tables, sets, options, math, random, sequtils],
+  std/[tables, sets, options, math, sequtils],
   stew/shims/net as stewNet, json_serialization/std/net,
   stew/[byteutils, endians2], chronicles, chronos, stint, bearssl,
-  eth/[rlp, keys, async_utils], types, encoding, node, routing_table, enr
+  eth/[rlp, keys, async_utils],
+  types, encoding, node, routing_table, enr, random2
 
 import nimcrypto except toHex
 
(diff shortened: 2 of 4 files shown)


eth/p2p/discoveryv5/random2.nim
```diff
@@ -0,0 +1,22 @@
...
+# TODO: Move these somewhere else?
+const randMax = 18_446_744_073_709_551_615'u64
+
+proc rand*(rng: var BrHmacDrbgContext, max: Natural): int =
+  if max == 0: return 0
+
+  var x: uint64
+  while true:
+    brHmacDrbgGenerate(addr rng, addr x, csize_t(sizeof(x)))
+    if x < randMax - (randMax mod (uint64(max) + 1'u64)): # against modulo bias
```
arnetheduck, line 13: this looks a bit complicated - I don't really remember how its done but:

https://github.com/python/cpython/blob/3d1c06e8b9eec5fc1ea2ed4dc1ea79c705da8ab8/Lib/random.py#L250

ping @mratsim @cheatfate

kdeme: So that's the part that I believe to have a mistake in the stdlib by the way.

The python version looks simpler, but it relies on having a `getrandbits`.

arnetheduck: I guess without `randbits` it's harder to avoid long loops, so yeah, that's probably significant - std c++ `uniform_int_distribution` could be another source of inspiration

kdeme: Well, this implementation is avoiding long loops too.
Here, we cut off the range of numbers to be a multiple of `max`. Be aware that `max` is inclusive, hence the additions.

The version that would not avoid long loops would be `if x <= uint64(max):`

I'll have a look at `uniform_int_distribution`.
There is also `arc4random_uniform`: http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/lib/libc/crypt/arc4random_uniform.c?rev=1.3&content-type=text/x-cvsweb-markup

mratsim: This is my techniques to have unbiaised number in a range:

https://github.com/mratsim/constantine/blob/504e2a9c25f36db4d8357a6ec83626c14ee548d7/helpers/prng_unsafe.nim#L87-L106

```Nim
func random_unsafe*(rng: var RngState, maxExclusive: uint32): uint32 =
  ## Generate a random integer in 0 ..< maxExclusive
  ## Uses an unbiaised generation method
  ## See Lemire's algorithm modified by Melissa O'Neill
  ##   https://www.pcg-random.org/posts/bounded-rands.html
  let max = maxExclusive
  var x = uint32 rng.next()
  var m = x.uint64 * max.uint64
  var l = uint32 m
  if l < max:
    var t = not(max) + 1 # -max
    if t >= max:
      t -= max
      if t >= max:
        t = t mod max
    while l < t:
      x = uint32 rng.next()
      m = x.uint64 * max.uint64
      l = uint32 m
  return uint32(m shr 32)
```

Source: https://www.pcg-random.org/posts/bounded-rands.html

PR #276, opened 2020-07-11
Comments and cleanup

kdeme:


eth/p2p/discoveryv5/dcli.nim
```diff
@@ -32,7 +32,7 @@ type
 
     nodeKey* {.
       desc: "P2P node private key as hex.",
-      defaultValue: PrivateKey.random().expect("Properly intialized private key")
+      defaultValue: PrivateKey.random(keys.newRng()[])
```
arnetheduck, line 35: generally good to use a single instance per application

arnetheduck: I'll just  assume the value is calculated at runtime :) ping @zah

kdeme: > generally good to use a single instance per application

Yeah, I realized that, but didn't find to time to look into how, see my comment on the same issue here: https://github.com/status-im/nim-waku/blob/7ccbcf17c9f905c2b8900fc9411d8a748634448e/waku/node/v1/config.nim#L96

kdeme: I guess I could do it by creating it into a `Some` value and then creating that key later in the application. Just slightly annoying.

zah: It's OK, the code follows established practices:
https://xkcd.com/221/

p.p.
It's fine really, the defaultValue expression is evaluated inside the `load Configuration` call.

PR #281, opened 2020-07-14
Add discv5 fuzzing test targets

kdeme:
Next to the test added here https://github.com/status-im/nim-eth/pull/274, some other fuzzing tests I've been running


arnetheduck:
where is it documented how to run the fuzzing tests? the readme doesn't mention them

kdeme:
hmm, right, it only gives the `rlp_inspect` example. I'll clarify the readme a bit.

Issue #267, opened 2020-07-08
Allow for updating of ENR of nodes in the routing table


kdeme:
In case we receive an ENR of a node (either through handshake or through findNode) which has a seqNumber > than the current one. This node its ENR should be updated.

Additionally, we should not always request an ENR at handshake, which we currently enforce by setting the sequence number always to 0. See https://github.com/status-im/nim-eth/blob/master/eth/p2p/discoveryv5/encoding.nim#L253 and https://github.com/status-im/nim-eth/blob/master/eth/p2p/discoveryv5/protocol.nim#L225

kdeme:
Might also want to add the request of a new ENR on ping/pong: https://github.com/status-im/nim-eth/blob/master/eth/p2p/discoveryv5/protocol.nim#L663

kdeme:
Tagged a security implication, as nodes could store lots of old ENRs and then send them on requests.

PR #283, opened 2020-07-17
Enr updating

kdeme:
This should fix https://github.com/status-im/nim-eth/issues/267 

Still want to update/add some tests on this.


eth/p2p/discoveryv5/encoding.nim
```diff
@@ -103,22 +103,26 @@ proc encodeAuthHeader*(rng: var BrHmacDrbgContext,
   var resp = AuthResponse(version: 5)
   let ln = c.localNode
 
-  # TODO: What goes over the wire now in case of no updated ENR?
```
kdeme, line 106: So without these changes, nothing was being send (as record.raw is empty).
With the changes an empty rlp list is send, as per specification.

Issue #242, opened 2020-05-28
Store handshake secrets in an LRU cache instead of current database


kdeme:
Handshake negotiated secrets that are required for passing messages to and from nodes are currently stored in `DiscoveryDB` (which is basically `TrieDatabaseRef`), and for which an in memory database is used currently. 

The key used to store the secrets is a combination of `NodeId` and `Address` (as suggested by the spec).
Now, secrets can be stored for nodes that are in our routing table, but also for nodes that are not in our Routing Table. Reason can be that e.g. a bucket is full, or a node is not eligible for being added e.g. because it has no valid address in its ENR. 
This means that for the nodes not stored in the routing table we can't actually delete the secrets unless we stored them in some other table.

I think a LRU cache would be more simple here to replace the database and solve the above issue, and the nice side effect of limiting the entries (e.g. not an entry necessarily for each node in the routing table) 

I believe a persistent (non memory db) storage for these secrets is not needed as these secrets are supposed to be ephemeral.
An optional persistent db will still be useful of course for storing "highly rated" peers to seed from after a restart: https://github.com/status-im/nim-eth/issues/189

kdeme:
> This means that for the nodes not stored in the routing table we can't actually delete the secrets unless we stored them in some other table.

To be clear, this is currently not done and thus some handshakes will be stored and never deleted (mem leakage as it is a memory db typically).

PR #292, opened 2020-09-08
Replace memory db with lrucache for temporary storage of sessions

kdeme:
to resolve https://github.com/status-im/nim-eth/issues/242


kdeme:
Screenshot_2020-09-08 New dashboard - Grafana(1)
CPU (rated over 60s) & memory metric of two nodes (one with this PR, one without) that are being "attacked" by creating continiously new nodes and starting handshakes.

(Actually cheated on the node without this PR, as I disabled also shared secrets removal on replaceNode, to make it easier :), but nevertheless it would be possible by creating nodes in certain nodeid space )

PR #293, opened 2020-09-10
support closing kvstore databases

tersec:


eth/db/kvstore.nim
```diff
@@ -108,6 +120,9 @@ proc put*(db: MemStoreRef, key, val: openArray[byte]): KvResult[void] =
   db.records[@key] = @val
   ok()
 
+proc close*(db: MemStoreRef): KvResult[void] =
+  ok()
```
arnetheduck, line 124: shouldn't this clear the memory store at least?

arnetheduck (approved):

eth/db/kvstore.nim
tersec: https://github.com/status-im/nim-eth/pull/293/commits/9759b4b7b85acccd2efc7d6b243a67a729089c42

PR #294, opened 2020-09-18
close sqlite transactions earlier

arnetheduck:
If the database is locked for reading, as it is when step returns ROW,
writes cannot checkpoint the wal leading to ever-increasing wal sizes
and a long delay at shutdown.

By resetting the transaction early, writes become more independent of
reads, memory is released earlier and wal can be checkpointed.


stefantalpalaru (approved):
Works for me. The WAL file does not shrink when checkpointed, but it seems to have hit its maximum size at 6.8 MiB.

arnetheduck:
databases rarely shrink files, but the wal in particular is removed on regular shutdowns

Issue #279, opened 2020-07-13
Update to discovery v5.1 (New packet format) + refactor receive and decodePacket


kdeme:
There is currently a proposal for a new packet format, which will change the outer packet encoding.

Currently some of the decoding is done directly in the `receive` call, and within the `decodePacket` (and encode counterpart for that matter), the handshake vs normal packet is done. 
This feels error prone and should probably be separated better. Specifically in the usage of the `AuthTag` and `Node` in different scenarions (like when needing to send a whoareyou because of decrypt failure, etc.). But also within the decode en encode calls.

The new packet format however deals with this as the tag will be removed in its current form, and instead it will be a header containing a `flag` (among other fields) that indicates the packet type. This should motivate for a cleaner separation in the code too, which should be considered when implementing the new packet format.

kdeme:
Done in https://github.com/status-im/nim-eth/pull/295

PR #295, opened 2020-09-25
Add discovery v5.1 implementation

kdeme:
Here's a go at the discovery v5.1 implementation. Closing https://github.com/status-im/nim-eth/issues/279

Didn't give too much thought right now at the passing/copying around of the data, might revise that at some time (if it is worth it), but feel free to suggest obvious improvements.
Might also restructure/split/move some code a bit at some point.

More important TODOs:
- [x] Two of the current existing test vectors still fail at the decrypt step, more specifically because of different generated secrets, which is rather weird as that is a part of the code that remain the same. TBI
- [x] FindNodes request with multiple distances
- [x] Allow for the not supported (new) packet types to not create a failure.
- [x] Revise / improve input size checks
- [x] Spec changes that are still being looked at
- [x] Fuzzing targets
- [x] Interop
- [x] Fix commented out handshake tests

Switch between discv5 and discv5.1 can be made with `UseDiscv51` compile time flag on `dcli` and the `test_discoveryv5`.

### DIFF
--- a/eth/p2p/discoveryv5/encodingv1.nim
+++ b/eth/p2p/discoveryv5/encodingv1.nim
@@ -478,8 +478,6 @@ proc decodeHandshakePacket(c: var Codec, fromAddr: Address, srcId: NodeId,
     ephKey, challenge.whoareyouData.idNonce)
 
   swap(secrets.readKey, secrets.writeKey)
-  c.sessions.store(srcId, fromAddr, secrets.readKey,
-    secrets.writeKey)
 
   let pt = decryptGCM(secrets.readKey, nonce, ct, header)
   if pt.isNone():
@@ -489,6 +487,11 @@ proc decodeHandshakePacket(c: var Codec, fromAddr: Address, srcId: NodeId,
     return err(DecryptError)
 
   let message = ? decodeMessage(pt.get())
+
+  # Only store the session secrets in case decryption was successful and also
+  # in case the message can get decoded.
+  c.sessions.store(srcId, fromAddr, secrets.readKey,
+    secrets.writeKey)
 
   return ok(Packet(flag: Flag.HandshakeMessage, message: message, srcId: srcId,
     node: newNode))
--- a/eth/p2p/discoveryv5/encodingv1.nim
+++ b/eth/p2p/discoveryv5/encodingv1.nim
@@ -187,10 +187,10 @@ proc encodeMessagePacket*(rng: var BrHmacDrbgContext, c: var Codec,
     # We might not have the node's keys if the handshake hasn't been performed
     # yet. That's fine, we send a random-packet and we will be responded with
     # a WHOAREYOU packet.
-    # TODO Minumum packet size is 63, we have here 16 + 23 + 32 = 71, so in theory
-    # we don't need to add random data? But then how do we know if decryption
-    # fails. Empty message automatically means -> whoareyou?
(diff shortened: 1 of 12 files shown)


eth/p2p/discoveryv5/encodingv1.nim
```diff
@@ -0,0 +1,574 @@
...
+
+  let message = ? decodeMessage(pt.get())
+
+  return ok(Packet(flag: Flag.OrdinaryMessage,
+    messageOpt: some(message), requestNonce: nonce, srcId: srcId))
+
+proc decodeWhoareyouPacket(c: var Codec, nonce: AESGCMNonce,
+    iv, header: openArray[byte]): DecodeResult[Packet] =
+  # TODO improve this
+  let authdata = header[staticHeaderSize..header.high()]
```
zah, line 412: Wouldn't it be nicer if you define object types that correspon to these fixed-size messages?
Then you would be able to write code along these lines:

```nim
let authDataMsg = cast[ptr AuthDataMsg](unsafeAddr header[0])
networkToHostOrder(authDataMsg[].recordSeq)
let whoareyou = WhoareyouData(
  requestNonce: authDataMsg[].nonce,
  idNonce: authDataMsg[].idNonce,
  recordSeq: authDataMsg[].recordSeq,
  ...
)
```

You can hide the pattern behind a safer construct that checks the size of the reinterpreted buffer.

eth/p2p/discoveryv5/encodingv1.nim
```diff
@@ -0,0 +1,574 @@
...
+  var ectx: CTR[aes128]
+  ectx.init(id.toByteArrayBE().toOpenArray(0, 15), iv)
+  result = newSeq[byte](header.len)
+  ectx.encrypt(header, result)
+  ectx.clear()
+
+proc hasHandshake*(c: Codec, key: HandShakeKey): bool =
+  c.handshakes.hasKey(key)
+
+proc encodeStaticHeader*(flag: Flag, nonce: AESGCMNonce, authSize: int):
```
zah, line 143: A lot of intermediate sequence allocations can be eliminated if the code was appending to an `OutputStream` instead.

eth/p2p/discoveryv5/encodingv1.nim
```diff
@@ -0,0 +1,574 @@
...
+  ## Decodes to the specific `Message` type.
+  if body.len < 1:
+    return err("No message data")
+
+  if body[0] < MessageKind.low.byte or body[0] > MessageKind.high.byte:
+    return err("Invalid message type")
+
+  # This cast is covered by the above check (else we could get enum with invalid
+  # data!). However, can't we do this in a cleaner way?
+  let kind = cast[MessageKind](body[0])
```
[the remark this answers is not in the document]
kdeme: Great :+1:  , I'll change it to use that call.

eth/p2p/discoveryv5/encodingv1.nim
kdeme: You mean of the https://github.com/status-im/nim-faststreams module?

I actually thought about it for rather the input parsing of some of the code, specifically during one of the many (later) changes to the spec. But didn't want to update anymore at that point to do such change. Don't think the benefit for this specific example is that big though? Anyway, perhaps something to benchmark and adjust later.

PR #302, opened 2020-10-12
Add support for multiple keyspaces in the Sqlite keystore

zah:
A compatible API can be introduced in RocksDB through the [Column Families](https://github.com/facebook/rocksdb/wiki/Column-Families) feature.

### DIFF
--- a/eth/db/kvstore_sqlite3.nim
+++ b/eth/db/kvstore_sqlite3.nim
@@ -3,16 +3,21 @@
 {.push raises: [Defect].}
 
 import
-  os,
+  os, strformat,
   sqlite3_abi,
   ./kvstore
 
 export kvstore
 
 type
-  SqStoreRef* = ref object of RootObj
-    env: ptr sqlite3
+  Sqlite3Ptr* = ptr sqlite3
+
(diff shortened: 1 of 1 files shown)


arnetheduck (approved):

arnetheduck:
actually, I just remembered another reason why I didn't put keyspaces in the kvstore interface: as long as you don't need cross-keyspace transactions, the virtual interface does not need to know about them - it can all be done at setup time which removes the need for an additional keyspace parameter that you must ship around, in addition to the store reference.

zah:
I considered creating additional instances of the key-value store (as far as I understand, this is what you meant), but it seemed more difficult to define a solution for managing the lifetime of the shared `sqlite3*` instance. Sqlite allows you to have multiple connections to the same file and that's one option to avoid the problem, but the documentation was not clear regarding the additional overhead introduced in this mode.

Another reason I liked this approach better is that it's more clear what should happen if we decide to bump the `user_version` stored in the database. The needed migration procedure will have a clear place in the code to execute during the initialization of the single store instance. Otherwise, something much more ad-hoc will have to be implemented.

PR #303, opened 2020-10-13
Richer Sqlite API

zah:
* Adds a strongly typed API for creating custom SQL queries and
  executing them

* Uses destructors to simplify the error handling in the init logic

### DIFF
--- a/eth/db/kvstore_sqlite3.nim
+++ b/eth/db/kvstore_sqlite3.nim
@@ -10,14 +10,42 @@ import
 export kvstore
 
 type
-  Sqlite3Ptr* = ptr sqlite3
+  RawStmtPtr = ptr sqlite3_stmt
+
+  AutoDisposed[T: ptr|ref] = object
+    val: T
+
+  # TODO: These should become AutoDisposed
+  #       This is currently considered risky due to the destructor
+  #       problem found in FastStreams (triggered when objects in
+  #       the GC heap have destructors)
+  Sqlite* = ptr sqlite3
+  SqliteStmt*[Params; Result] = distinct RawStmtPtr
+  NoParams* = tuple # this is the empty tuple
+  ResultHandler*[T] = proc(val: T) {.gcsafe, raises: [Defect].}
 
   KeySpaceStatements = object
-    getStmt, putStmt, delStmt, containsStmt: ptr sqlite3_stmt
+    getStmt, putStmt, delStmt, containsStmt: RawStmtPtr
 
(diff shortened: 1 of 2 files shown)


arnetheduck:
meh, while I agree the lack of destructors is somewhat ugly, nim 1.2 destructors don't work so writing code that... will maybe use them in the future, if they ever end up working, should be done _after_ they work, not before - that's a bridge to cross when and if we switch to 1.4 - the current code is trivially verifiable to be correct - not so with the broken codegen in 1.2, so it's a matter of optics / aesthetics really.

Also, the kvstore was written to be trivially auditable, predictable and understood without any magic, on purpose - the memory handling is explicit and easy to track this way, and easy to control/manipulate per statement - it is not a generic query API so introducing 100 lines of generic code just for the sake of it doesn't really buy anything - this can be seen because _no code is removed_ from the additional layering/abstraction - behind the virtual interface, it can't be used anyway by "end users"

zah:
Destructors work in 1.2. We've seen issues with them when they are used as finalizers by the GC which is not the case here.

The added generic code is already used in NBC for the sequence-like tables that don't fit well in the key value model.

arnetheduck:
they don't work, if they don't work with related features such as finalizers - it's impossible to remember that some rare but used case is broken for a feature like that, specially when the breakage is random and at runtime, along _with all the other broken stuff in Nim_ - it's not possible for a group of humans to keep track of all the breakage every time and every commit - this is not something we want to waste time and resources on - what happens if the type is included in a type with a finalizer? what if it's in some generic code? etc.  

this was a kvstore that used sqlite - it was written with that use case in mind and was simply and reliable - now it's morphing into an ORM and a language experiment instead which goes far beyond its intent.

zah:
To avoid these risks, the `AutoDisposed` type that uses a destructor has not been exported or added to a general purpose library like `stew`. Its use is local to this module and it achieves the goal of simplifying the code here. YMMV.

The Sqlite facilities added here will be expanded further with support for transactions because we have specific problems in NBC that can be solved appropriately with such APIs. If we ever need to switch to a different data store, we can *then* spend the time to implement an analogous API for it, following the YAGNI principle. We discussed this topic during out latest dev call and the consensus was that we should generally strive to keep our options open by relying on abstractions that can be more easily ported to different back-ends, but we should also not cripple ourselves into using the wrong tool for the job when the development costs or the run-time costs associated with this get prohibitively high.

@mratsim plans to apply the newly added tools to the fork-choice problem as well.

Issue #269, opened 2020-07-08
IP limiting nodes / ENRs in discovery v5


kdeme:
To make a sybil or eclipse attack more difficult it would probably be a good idea to have IP limits in the routing table of discovery v5.

Limits that we could set:
- IP limit per routing table
- IP limit per bucket
- ...?

PR #308, opened 2020-11-06
Discv5 ip limits for routing table

kdeme:
Attempt on https://github.com/status-im/nim-eth/issues/269. 
This turned out to be more cumbersome than expected. A lot of corner cases and issues possible due to wrong ordering of checks because of replacement cache, bucket splitting and then also the fact that a newer version of ENR with different ip address is possible. Added tests for all those cases but it still feels rather flaky (could be improved still though).

To make things worse, while creating this PR I realized that the current approach might actually introduce another issue. That is, a malicious peer could temporary make a bucket or the routing table hit an ip limit by responding with records with ip-addresses it doesn't own/control. Causing a node that does own the ip address not to be allowed to be added, until those others are evicted when they can't be verified at revalidation.

So looks like the ip limit checks should probably only be applied when setting a node to verified (currently named "seen"). Either that or we need to verify nodes immediately before adding. But this is typically not recommended (e.g. at least not by original Kademlia paper) as it could/would introduce another DoS vector.

This would require some additional changes, such as only using verified nodes, else the ip limits have not much use. But that is probably a good change (security wise). It will delay the amount of discovered nodes though.

I'll rethink this and keep the PR in draft for now. Input welcome of course.

### DIFF
--- a/eth/net/utils.nim
+++ b/eth/net/utils.nim
@@ -1,0 +1,27 @@
+import
+  std/[tables, hashes],
+  stew/shims/net as stewNet
+
+{.push raises: [Defect].}
+
+type
+  IpLimits* = object
+    limit*: uint
+    ips: Table[ValidIpAddress, uint]
+
+proc hash(ip: ValidIpAddress): Hash = hash($ip)
+
+proc inc*(ipLimits: var IpLimits, ip: ValidIpAddress): bool =
+  let val = ipLimits.ips.getOrDefault(ip, 0)
+  if val < ipLimits.limit:
+    ipLimits.ips[ip] = val + 1
+    true
+  else:
+    false
+
+proc dec*(ipLimits: var IpLimits, ip: ValidIpAddress) =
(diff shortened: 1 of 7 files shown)


kdeme:
Decided to go ahead with the current implementation.

It is still true that a malicious node could fill another node its routing table / bucket with ENRs with IP's he doesn't own. The side effect however, would not be too bad. The node that actually owns the IP would basically not be reached from the outside (or less at least), as other nodes their tables will be poisoned with the wrong ones. However, it can still search and find nodes to connect to itself. So it would practically be similar to a node that didn't set up an "external address".

However, if you would apply the limits only on "verified" (= contacted) nodes then worse is possible.
Currently, lookups are done also with not verified nodes (as a way to verify them), this means that if ip limits are not applied on these, lookups are possible on a high set of nodes owned by the same identity. I consider this worse then the above.
The other option here would be to not use unverified nodes for lookups, but that likely slow down discovery start up a lot.

PR #311, opened 2020-11-17
wait for cancellation to be processed on timeout

arnetheduck:


eth/async_utils.nim
```diff
@@ -10,7 +10,10 @@ template awaitWithTimeout*[T](operation: Future[T],
   let f = operation
   await f or deadline
   if not f.finished:
-    cancel f
```
zah, line 13: So, is it ever safe to call `cancel` now?

zah (approved):

eth/async_utils.nim
arnetheduck: well, cancel _schedules_ a cancellation but doesn't enact it - that creates a random ordering down the line and because of all the shared state in async code in general, it's simply not _feasible_ (for a human) to reason about the code down the line - ie I'd consider `cancel` on its own completely unsafe as far as constructs go and avoid it, but that's mainly because nim lacks the features (ownership/borrowing) that complement it. another way to put this is that there exist cases where cancel is safe, but when it's hidden like this, bugs will happen.

Issue #280, opened 2020-07-13
Stale bootstrap nodes will block the bucket for revalidation


kdeme:
Bootstrap nodes are currently never removed. Also not when pinging them fails.
The idea is here that they should be highly available (trusted) nodes that one can always fallback to.

However, if a bootstrap node is unreachable (goes stale), then this has the side-effect that it will never be put in front of the bucket and practically block revalidation of all nodes in front of that bucket (until the bucket gets split at least).

Those nodes will still be "tested" for availability by doing lookups (findNode requests) though, so it is not that terrible.

Better solution would be to remove the bootstrap nodes, but always keep them aside as fallback for when the routing table drops to `< bootstrapnodes.len` peers.

PR #314, opened 2020-11-27
Fix issue #280 by repopulating the routing table

kdeme:
* Remove also bootstrap nodes if they fail to respond

* Repopulate the routing table (with bootstrap nodes) when the

routing table drops low


kdeme:
Will close as some of the changes to prepare for reseed/refreshing are added in https://github.com/status-im/nim-eth/pull/322

And the actual refresh on drop of nodes should probably be revised and done when the count actually drops (say node gets removed) below threshold.

PR #321, opened 2021-01-05
add github action

jangko:
can't reopen previous PR due to force push.


kdeme (approved):
@jangko So the two workarounds currently are:
- Linux amd64: Use distro rocksdb package instead of self compile
- Windows 64bit: Run rlp tests in `-d:debug` mode
Correct?

jangko:
@kdeme: correct

PR #323, opened 2021-01-18
Improve network address discovery / NAT setup

kdeme:
Testing out logic for now. Moving also the `setupNat` (now `setupAddress`) inside nim-eth, as the same or similar code is used for nimbus-eth2, waku v1, waku v2, nimbus-eth1 and dcli.

Needs further polish on error passing, log messages, etc.


eth/net/nat.nim
```diff
@@ -234,3 +234,88 @@ proc redirectPorts*(tcpPort, udpPort: Port, description: string): Option[(Port,
...
+    (none(ValidIpAddress), tcpPort, udpPort)
+
+proc getBestRoutePublicIp*(): Option[ValidIpAddress] {.raises: [Defect].} =
+  # Avoiding Exception with initTAddress and can't make it work with static.
+  let publicAddress = TransportAddress(family: AddressFamily.IPv4,
+    address_v4: [1'u8, 1, 1, 1], port: Port(0))
+  # Note: `publicAddress` is only used an "example" IP to find the best route,
+  # no data is send over the network to this IP!
+  let route = getBestRoute(publicAddress)
+  if (not route.source.isUnspecified()) and route.source.isPublic():
```
cheatfate, line 266: `getBestRoute()` procedure quires OS routing table to find the best route for connection to specific public IP address.
So its absolutely possible that OS do not know have any routes to the public internet, it is also could have wrong/incorrect routing table, but your error handling do not provide any meaningful error messages about this case. So if `getBestRoute()` returns nothing you can stop even try to communicate with global network.

Also in this PR you forgot about PCs which could not have IPv4 routing at all and only could have IPv6 routing, so its better to query best route for both IPv6 and IPv4 addresses.

kdeme: Yes, good point. The warnings currently are to focused on public/private IP part. If the call fails I should warn/error specifically on this. And perhaps even quit like you state, although I'm unsure what is the best action to take in that case (Perhaps in some startup scenario the route is not set yet, but then again, quitting would usually restart the application in a proper setup and it would get resolved).

Regarding IPv6, I didn't forget about this. Just deliberately decided not to focus on this yet as IPv6 is not really supported higher up in our nbc application (no listeners). I did realize and mentioned however that it would be good to take this in at the same time, and I think I'll still do so in this PR to avoid having to rework parts again. Anyhow, it would be separate calls to figure out the IPv6 "advertised" address and the IPv4 one.

kdeme:
Finished up the logic. Could use another pass for a bigger refactor and changes, but it would depend on some other features that one might want to add so I kept it to a minimum for now and did not go there in this PR.

Some of those items that could be added/changed, but probably better in another PR:
- Allow for tcp or udp port map only (e.g. when running only a discovery service, don't port map TCP)
- Allow for multiple port maps (e.g. Waku bridge has such use case, were a Waku v1 and Waku v2 node needs to be run) (or multiple addresses to be set up basically)
- Allow for custom port maps e.g. 9000:1234 (not sure of there is a real use case for this)
- Depending on whether we want to stick with that `--nat:...` option or rather have several independent options, the code could be reorganized and slightly simplified probably. Could also just start with splitting config parsing and actual logic.
- Additional calls for IPv6 version when we want to support that.

eth/net/nat.nim
```diff
@@ -237,3 +237,109 @@ proc redirectPorts*(tcpPort, udpPort: Port, description: string): Option[(Port,
...
+      return setupNat(NatPmp, tcpPort, udpPort, clientId)
+    else:
+      if nat.startsWith("extip:"):
+        try:
+          # any required port redirection must be done by hand
+          let ip = ValidIpAddress.init(nat[6..^1])
+          return (some(ip), some(tcpPort), some(udpPort))
+        except ValueError:
+          error "Not a valid IP address", address = nat[6..^1]
+          quit QuitFailure
```
arnetheduck, line 342: since this is a function in a library, we really shouldn't be quitting..

kdeme: Yeah, good point. Thought about that actually and was also one of the reasons why I don't quit on for example https://github.com/status-im/nim-eth/pull/323/files#diff-161c148740a4e155d7afca5fd40f23461ba6581e2949c694ffd89aca776b0d0dR293 .
I guess I could already split the parsing of the options with the actual logic, and keep the first in nimbus-eth2. Was planning on doing that together with possible cli option changes (if we decided on that). Either that or returning a result, but that is a bit silly unless I provide the currently logged error messages in (and possibly an enum) and remove the logs there. Not really needed imo.

Issue #207, opened 2020-04-02
Use Pong packet to determine the external IP address and port


kdeme:

oskarth:
This would indeed be super useful for UX as a node runner

PR #325, opened 2021-01-26
Add ENR external IP:Port update based on majority of pong responses

kdeme:
Resolves https://github.com/status-im/nim-eth/issues/207

Also related to https://github.com/status-im/nimbus-eth2/issues/2160

Could/Would also resolve https://github.com/status-im/nimbus-eth2/issues/1954

### DIFF
--- a/eth/p2p/discoveryv5/dcli.nim
+++ b/eth/p2p/discoveryv5/dcli.nim
@@ -14,12 +14,12 @@ type
   DiscoveryConf* = object
     logLevel* {.
       defaultValue: LogLevel.DEBUG
-      desc: "Sets the log level."
+      desc: "Sets the log level"
       name: "log-level" .}: LogLevel
 
     udpPort* {.
       defaultValue: 9009
-      desc: "UDP listening port."
+      desc: "UDP listening port"
       name: "udp-port" .}: uint16
 
     listenAddress* {.
@@ -28,32 +28,39 @@ type
       name: "listen-address" }: ValidIpAddress
 
     bootnodes* {.
-      desc: "ENR URI of node to bootstrap discovery with. Argument may be repeated."
+      desc: "ENR URI of node to bootstrap discovery with. Argument may be repeated"
       name: "bootnode" .}: seq[enr.Record]
 
     nat* {.
       desc: "Specify method to use for determining public address. " &
-            "Must be one of: any, none, upnp, pmp, extip:<IP>."
+            "Must be one of: any, none, upnp, pmp, extip:<IP>"
       defaultValue: "any" .}: string
 
+    enrAutoUpdate* {.
+      defaultValue: false
+      desc: "Discovery can automatically update its ENR with the IP address " &
+            "and UDP port as seen by other nodes it communicates with. " &
+            "This option allows to enable/disable this functionality"
+      name: "enr-auto-update" .}: bool
+
     nodeKey* {.
-      desc: "P2P node private key as hex.",
+      desc: "P2P node private key as hex",
       defaultValue: PrivateKey.random(keys.newRng()[])
       name: "nodekey" .}: PrivateKey
 
     metricsEnabled* {.
       defaultValue: false
-      desc: "Enable the metrics server."
+      desc: "Enable the metrics server"
       name: "metrics" .}: bool
 
     metricsAddress* {.
       defaultValue: defaultAdminListenAddress(config)
-      desc: "Listening address of the metrics server."
+      desc: "Listening address of the metrics server"
       name: "metrics-address" .}: ValidIpAddress
 
     metricsPort* {.
       defaultValue: 8008
-      desc: "Listening HTTP port of the metrics server."
+      desc: "Listening HTTP port of the metrics server"
       name: "metrics-port" .}: Port
 
     case cmd* {.
@@ -163,7 +170,8 @@ proc run(config: DiscoveryConf) =
   let
     (ip, tcpPort, udpPort) = setupNat(config)
     d = newProtocol(config.nodeKey, ip, tcpPort, udpPort,
-      bootstrapRecords = config.bootnodes, bindIp = config.listenAddress)
+      bootstrapRecords = config.bootnodes, bindIp = config.listenAddress,
+      enrAutoUpdate = config.enrAutoUpdate)
 
   d.open()
 
--- a/eth/p2p/discoveryv5/protocol.nim
+++ b/eth/p2p/discoveryv5/protocol.nim
@@ -75,7 +75,7 @@
 import
   std/[tables, sets, options, math, sequtils, algorithm],
   stew/shims/net as stewNet, json_serialization/std/net,
-  stew/[endians2, arrayops], chronicles, chronos, stint, bearssl, metrics,
+  stew/endians2, chronicles, chronos, stint, bearssl, metrics,
   eth/[rlp, keys, async_utils],
   types, encoding, node, routing_table, enr, random2, sessions, ip_vote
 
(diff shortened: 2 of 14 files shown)


eth/p2p/discoveryv5/dcli.nim
```diff
@@ -28,32 +28,39 @@ type
...
+      desc: "ENR URI of node to bootstrap discovery with. Argument may be repeated"
       name: "bootnode" .}: seq[enr.Record]
 
     nat* {.
       desc: "Specify method to use for determining public address. " &
-            "Must be one of: any, none, upnp, pmp, extip:<IP>."
+            "Must be one of: any, none, upnp, pmp, extip:<IP>"
       defaultValue: "any" .}: string
 
+    enrAutoUpdate* {.
```
kdeme, line 39: This option would be added also to nimbus-eth2.

Default should eventually change to `true`

eth/p2p/discoveryv5/protocol.nim
```diff
@@ -897,12 +898,42 @@ proc refreshLoop(d: Protocol) {.async, raises: [Exception, Defect].} =
   except CancelledError:
     trace "refreshLoop canceled"
 
+proc ipMajorityLoop(d: Protocol) {.async, raises: [Exception, Defect].} =
+  try:
+    while true:
+      let majority = d.ipVote.majority()
+      if majority.isSome():
```
kdeme, line 905: Not sure if we should still run this and do the warning below when `enrAutoUpdate` is false.

stefantalpalaru:
This opens up the possibility for a trivial DoS attack, when you control enough nodes to convince the target to advertise a wrong IP, thus depriving it of incoming connections and redirecting traffic from its peers to a secondary target of your choice. Win-win for the attacker.

kdeme:
> This opens up the possibility for a trivial DoS attack, when you control enough nodes to convince the target to advertise a wrong IP, thus depriving it of incoming connections and redirecting traffic from its peers to a secondary target of your choice. Win-win for the attacker.

What are "enough" nodes according to you?

Yes, this opens up the possibility of an attack to make the node advertise the wrong IP. As such there is a risk here with adding and enabling this feature.
The attack will of course always be possible, as long as you control a lot of nodes (and IPs) and thus can control the routing table, but I wouldn't call that trivial. And in that case, the attacker can do worse (or rather more interesting) things probably than just (temporary) cripple a node.

Why is it not trivial? Rationale:
- Only one vote per NodeId is counted, and they are removed over time.
- IP:port values are provided through the pong message. Our node initiates this by first doing a ping message. Unsolicited pong messages are ignored. Thus, malicious nodes can not spam pong us to influence the majority IP.
- At interval (with some randomness) pings are send to the least recently contacted node in a random bucket from the routing table.
- Only messages that our node initiates (ping, findnode, talkreq) and that successfully get a response moves nodes to the front of the bucket. Not the ones that are started from peers. Also findNode requests have typically a randomness to it, as we pick a random nodeid.
- If a peer fails the respond, it gets replaced. It doesn't remain at the end of the bucket. (Side note: This could be changed in the future, as we should probably attempt some more pings. But in this case the node should either be marked as stale or be temporary removed from the table or something similar)
- There are IP limits on the buckets (max 2 the same) and the whole routing table (max 10 the same).

(I'll document this information in comments)

As you can see, a lot of this is controlled or initiated by our local node. This makes me confident that it can not be abused easily (e.g. without owning the routing table). I have not quantified how many nodes it would take to abuse this, this would require some simulations probably.

Perhaps I'm missing something that can still be abused, please indicate if you were thinking of something concretely.

The threshold of minimum votes could also be increased (I took the value that LH uses now).
We could also decide to only update ip address if there is like a 2/3 or even 90% majority instead of the current 50%.
Perhaps some extra measures can be added for protection, please indicate if you can think of any other.

Additionally,  I was considering of not allowing this option to be used in combination with for example the current `--nat:extip:<IP address>` option.

As last, abuse on this is something that could be spotted in the metrics (peer count and incoming peers), but it would not happen immediately, as the peers first have to drop. Once the peer count starts dropping (which would happen not that fast probably), the node would still function somewhat (outgoing peers and leftover incoming), albeit, with known subnet issues that might arise.
Come to think of it, perhaps a metric on IP-address update would be a good add to be able to alarm on.

Regarding redirecting to a secondary target, practically making others DDoS your target.
This would indeed cause some initial redirected traffic from peers trying to contact the node based on the received ENR. However, the target would not respond (it couldn't even if it wanted, as it does not control the key pair from which the NodeId is derived from), and because of this, that ENR will be removed quickly from the routing tables and not forwarded to even more nodes (removing somewhat the Distributed part of DDoS).
Additionally, nothing stops you from spreading invalid ENRs with IPs you do not own but want to target traffic to.

stefantalpalaru:
One way to make attacks harder is to only trust bootnodes to tell you your external IP.

kdeme:
> One way to make attacks harder is to only trust bootstrap nodes to tell you your external IP.

This has crossed my mind, and while it would be a more simple solution, it works against the idea of being less dependable on the bootstrap nodes.

eth/p2p/discoveryv5/protocol.nim
zah: I think we'll probably decide to remove the `enrAutoUpdate` option altogether in the near future (in favor of having it always enabled).

arnetheduck: ENR auto-update is strictly worse for server setups with a static IP: it decreases security without providing benefit - it's arguable that we might want to use the ENR feature by default instead of upnp because it provides a better user experience, but setting a static IP is always better, if you have one - in particular, on ipv6, you're much more likely to have a static ip and not need these shenanigans.

zah: Well, I clarified in my message that having this policy enabled by default makes sense when `extip` is not specified (how you would specify your static IP otherwise?)

kdeme: I think we are all on the same line here :)

The only question is indeed whether to also default enable it in the case of "private local network". It will not replace upnp however as port mapping is still needed.

kdeme: By the way, my original question was actually about whether to still run the counting code, also when the feature is disabled. To show/warn the user but not actually update the address.

adnm9:
Hi, I was wondering if it might be a good idea to add some sort of latency check on the ip that's voted, and compare that to the latency of other nodes, to see if it's really the right ip. You could also randomly send out pings to yourself and monitor the incoming to see if it matches up.

kdeme:
> Hi, I was wondering if it might be a good idea to add some sort of latency check on the ip that's voted, and compare that to the latency of other nodes, to see if it's really the right ip.

Yeah, that's a good idea. Would have to add some monitoring/metrics to get for example the average latency of the nodes, but yes, it could give higher confidence on the voted IP.

> You could also randomly send out pings to yourself and monitor the incoming to see if it matches up.

Not sure if I understand what you mean here but to do so you would need your current IP to begin with. And if you got that, no need to run this in the first place.
Also, I think a (discovery) ping to your external IP when you are behind a NAT will be returned with as source IP your gateway (e.g. 192.168.1.1 or 192.168.1.254, etc.) in some cases.

adnm9:
Regarding the second one, what I mean is this: try the ip address that you're being given by the peers, and ping the nimbus port, and then at the other end monitor the incoming. If when you do this at a random time you get the ping from the ip address that they say is you, it's highly unlikely that the attacker sent it out (if he even controlled that ip) exactly then, and it certainly eliminates the possibility of being used for a DDOS attack, as why would the victim randomly ping you.

kdeme:
Right, I understand you now. Yeah, that would probably work, and the attacker wouldn't even be able to reply in the discovery protocol as it needs your private key for that.
I'm just not 100% sure if this would work with all home gateways.

adnm9:
Here's what I came up with: If some threshold of doubt (perhaps manually adjustable) as to the ip address is met, create a new peer id, and send a ping to the regular peer id. If only one of the ip addresses works (i.e. comes through), than that's the one. Now there is the option for the attacker, if he controls the ip in question or can man-in-the-middle it, to simply forward the ping himself to try to fool you, resulting in more than one ip address working. This can only be somewhat mitigated as follows: 1. Obviously, if one of the possibilities had already been concretely established previously than go with it. 2. Try sending multiple times to each ip and measure latency and reliability of the connections. The real one should generally have less latency, and, in the event of the man-in-the-middle scenario, which is the only real DDOS vector, there should be reliability issues as well (for the fraudulent one), as the attacker won't always be able to intercept the message.

PR #327, opened 2021-02-13
turn networkId into distinct uint

jangko:


eth/p2p/private/p2p_types.nim
```diff
@@ -7,8 +7,10 @@ const
   useSnappy* = defined(useSnappy)
 
 type
+  NetworkId* = distinct uint
```
arnetheduck, line 10: better to avoid platform-specific sizes for external data

jangko: I'll remember it

PR #335, opened 2021-03-22
Fix minor exception effect issues

arnetheduck:
Makes code compatible with
https://github.com/status-im/nim-chronos/pull/166 without requiring it.


arnetheduck:
failure is a download issue, should be fine

PR #336, opened 2021-03-24
Add raises annotations to make exception tracking work

kdeme:
See https://github.com/status-im/nim-chronos/pull/166

### DIFF
--- a/eth/p2p/rlpx.nim
+++ b/eth/p2p/rlpx.nim
@@ -60,7 +60,8 @@ chronicles.formatIt(Peer): $(it.remote)
 
 include p2p_backends_helpers
 
-proc requestResolver[MsgType](msg: pointer, future: FutureBase) {.gcsafe.} =
+proc requestResolver[MsgType](msg: pointer, future: FutureBase)
+    {.gcsafe, raises:[Defect].} =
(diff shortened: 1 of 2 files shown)


eth/p2p/rlpx.nim
```diff
@@ -72,10 +73,13 @@ proc requestResolver[MsgType](msg: pointer, future: FutureBase) {.gcsafe.} =
     # here. The only reasonable explanation is that the request should
     # have timed out.
     if msg != nil:
-      if f.read.isSome:
-        doAssert false, "trying to resolve a request twice"
-      else:
-        doAssert false, "trying to resolve a timed out request with a value"
+      try:
+        if f.read.isSome:
```
zah, line 77: The fact that `f.read` raises a `ValueError` when it's not finished seems silly. This makes it hard to write correct code that relies on exception tracking. `read` should have a pre-condition that the Future is finished and it should raise a `Defect` when it isn't.

cc @cheatfate, @arnetheduck.

cheatfate: I agree it should be `Defect` (its some legacy code from `asyncdispatch`).

Issue #341, opened 2021-04-02
discv4: Kademlia crash when trying to sync


jlokier:
This was reported for status-im/nimbus-eth1#489 back in April 2020 but it properly belongs here.

When using discv4 (Kademlia) to find peers, there is consistently a crash after a few minutes.  It occurs for everyone on Ropsten testnet, and for most of us on mainnet as well.  The error looks like this (Linux server, `--log-level:TRACE`):

```
TRC 2021-03-03 16:47:56.115+00:00 Bonding to peer                            topics="discovery" tid=6599 file=kademlia.nim:329 n=Node[<someone's IP>:30303]
TRC 2021-03-03 16:47:56.115+00:00 Bonded with candidates                     topics="discovery" tid=6599 file=kademlia.nim:391 count=16
TRC 2021-03-03 16:47:56.296+00:00 Discovered nodes                           topics="discovery" tid=6599 file=discovery.nim:293 nodes=273
TRC 2021-03-03 16:47:56.297+00:00 Starting lookup; initial neighbours:       topics="discovery" tid=6599 file=kademlia.nim:399 closest="@[Node[<someone's IP>:30303], Node[<someone's IP>:30303], Node[<someone's IP>:30303], Node[<someone's IP>:30303], Node[<someone's IP>:30303], Node[<someone's IP>:30303], Node[<someone's IP>:30303], Node[<someone's IP>:30303], Node[<someone's IP>:30303], Node[<someone's IP>:30303], Node[<someone's IP>:30303], Node[<someone's IP>:20638], Node[<someone's IP>:20630], Node[<someone's IP>:57955], Node[<someone's IP>:30303], Node[<someone's IP>:30303]]"
TRC 2021-03-03 16:47:56.297+00:00 Node lookup; querying                      topics="discovery" tid=6599 file=kademlia.nim:402 nodesToAsk="@[Node[<someone's IP>:30303], Node[<someone's IP>:30303], Node[<someone's IP>:30303]]"
TRC 2021-03-03 16:47:56.297+00:00 >>> find_node to                           topics="discovery" tid=6599 file=discovery.nim:132 n=Node[<someone's IP>:30303]
TRC 2021-03-03 16:47:56.297+00:00 >>> find_node to                           topics="discovery" tid=6599 file=discovery.nim:132 n=Node[<someone's IP>:30303]
[...]

PR #342, opened 2021-04-02
discv4: Fix Kademlia crash when trying to sync

jlokier:
Fixes status-im/nim-eth#341, fixes status-im/nimbus-eth1#489.

When using discv4 (Kademlia) to find peers, there is a crash after a few minutes.  It occurs for most of us on Eth1 mainnet, and everyone on Ropsten.

The cause is `findNodes` being called twice in succession to the same peer, within about 5 seconds of each other.  ("About" 5 seconds, because Chronos does not guarantee to run the timeout branch at a particular time, due to queuing
and clock reading delays.)

Then `findNodes` sends a duplicate message to the peer and calls `waitNeighbours` to listen for the reply.  There's already a `waitNeighbours` callback in a shared table, so that function hits an assert failure.

Ignoring the assert would be wrong as it would break timeout logic, and sending `FindNodes` twice in rapid succession also makes us a bad peer.

As a simple workaround, just skip `findNodes` in this state and return a fake empty `Neighbours` reply.  This is a bit of a hack as `findNodes` should not be called like this; there's a logic error at a higher level.  But it works.

Tested for about 4 days constant operation on Ropsten.  The crash which occured every few minutes no longer occurs, and discv4 keeps working.

### DIFF
--- a/eth/p2p/kademlia.nim
+++ b/eth/p2p/kademlia.nim
@@ -371,6 +371,13 @@ proc lookup*(k: KademliaProtocol, nodeId: NodeId): Future[seq[Node]] {.async.} =
   var nodesSeen = initHashSet[Node]()
 
   proc findNode(nodeId: NodeId, remote: Node): Future[seq[Node]] {.async.} =
+    if remote in k.neighboursCallbacks:
+      # Sometimes findNode is called while another findNode is already in flight.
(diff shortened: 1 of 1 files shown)


kdeme (commented):
With https://github.com/status-im/nimbus-eth1/pull/576 working, this can be updated to head and then merged

PR #347, opened 2021-04-27
Discv4 with chronos strict usage

kdeme:
Finished some further nim-eth clean-up I had still laying around. More work needed in the rlpx parts though.

Should not change functionality. (However, it would be a good idea to look at that too in comparison with discv5)


eth/p2p/discovery.nim
```diff
@@ -18,6 +18,8 @@ import
 export
   Node, results
 
+{.push raises: [Defect].}
```
arnetheduck, line 21: fyi I've been consistently placing these just under the license, before the imports, elsewhere, as a "language dialect marker" basically

kdeme: Right, I've been consistently not doing that :)

I'll do the same on all locations here.

PR #350, opened 2021-05-13
kvstore fixes

arnetheduck:
Storing large blobs in a "WITHOUT ROWID" table turns out to be extremely
slow when the tree must be rebalanced.

* Split out keystore capability into separate interface, making each
keystore a separate instance
* Disable "WITHOUT ROWID" optimization by default
* Implement prefix lookup that allows iterating over all values with a
certain prefix in their key

### DIFF
--- a/eth/db/kvstore_sqlite3.nim
+++ b/eth/db/kvstore_sqlite3.nim
@@ -3,7 +3,7 @@
 {.push raises: [Defect].}
 
 import
-  std/[os, strformat],
+  std/[os, options, strformat],
   sqlite3_abi,
   ./kvstore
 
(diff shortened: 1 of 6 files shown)


eth/db/kvstore_sqlite3.nim
```diff
@@ -396,68 +464,61 @@ proc init*(
...
+  defer:
+    # We'll "move" ownership to the return value, effectively disabling "close"
+    close(tmp)
+
+  tmp.getStmt = prepare(db.env, "SELECT value FROM " & name & " WHERE key = ?;")
+  tmp.putStmt =
+    prepare(db.env, "INSERT OR REPLACE INTO " & name & "(key, value) VALUES (?, ?);")
+  tmp.delStmt = prepare(db.env, "DELETE FROM " & name & " WHERE key = ?;")
+  tmp.containsStmt = prepare(db.env, "SELECT 1 FROM " & name & " WHERE key = ?;")
+  tmp.findStmt0 = prepare(db.env, "SELECT key, value FROM " & name & ";")
```
zah, line 514: Until now in the review, it was a bit of a mystery what would be the difference between the different find statements. More descriptive names could help (e.g. `findAll`, `findGreater`, `findRange`).

zah (approved):

eth/db/kvstore_sqlite3.nim
arnetheduck: 0, 1 and 2 parameters :) it's described further down where they're used together only

PR #358, opened 2021-05-20
CI: refactor Nim compiler caching

stefantalpalaru:


jangko:
I've seen that `sparse binary trie` stalled for many times. probably we need to wrap some unittest in `tests/trie` calls into a proc instead of let them running in global space. usually that solves.

stefantalpalaru:
I can't make that test hang on my machine. At first, I thought it lacks random bytes inside that CI VM, but that's not it.

arnetheduck:
> wrap some unittest in tests/trie calls into a proc

@jangko `unittest2` wraps every test in a separate proc - ideally, all tests can be switched to it

PR #360, opened 2021-05-26
sqlite3: support option type

arnetheduck:
.. and any integer when writing but don't support int32 on reads -
internally sqlite will truncate on overflow which isn't nice.

### DIFF
--- a/eth/db/kvstore_sqlite3.nim
+++ b/eth/db/kvstore_sqlite3.nim
@@ -93,10 +93,13 @@ proc bindParam(s: RawStmtPtr, n: int, val: auto): cint =
         sqlite3_bind_blob(s, n.cint, nil, 0.cint, nil)
     else:
       {.fatal: "Please add support for the '" & $typeof(val) & "' type".}
-  elif val is int32:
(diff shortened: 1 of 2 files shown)


eth/db/kvstore_sqlite3.nim
```diff
@@ -161,6 +166,15 @@ template readResult(s: RawStmtPtr, column: cint, T: type): auto =
   else:
     {.fatal: "Please add support for the '" & $(T) & "' type".}
 
+template readResult(s: RawStmtPtr, column: cint, T: type): auto =
+  when T is Option:
+    if sqlite3_column_type(s, column) == SQLITE_NULL:
+      none(typeof(default(T).get()))
+    else:
+      some(readSimpleResult(s, column, typeof(default(T).get())))
```
zah, line 174: Since, `readSimpleResult` is reused, it's worth making it a proc now.

zah (approved):

eth/db/kvstore_sqlite3.nim
arnetheduck: can't, it returns openarray

PR #361, opened 2021-05-26
sqlite: quick exec with result

arnetheduck:


zah:
It would be nice to add a test

zah (approved):

PR #362, opened 2021-06-01
Add (incomplete) implementation of Portal wire protocol

kdeme:
First stab at Portal wire protocol.

Interop with ddht tested and works for ping and findnode message.

Used some local minimum SSZ implementation, grabbed selectively (and altered slightly) from nimbus-eth2 repository, after the (hi)story here: https://github.com/status-im/nim-ssz-serialization/pull/1
Can probably be even further minimized but not the goal right now. Will add to SSZ what is deemed necessary in further development.

Some TODOs still:
- Documentation
- Test interop with Trin
- Probably add some minimum SSZ tests

Also not fully happy with the fact that there is a need for a `TalkProtocol` and `PortalProtocol` which hold a ref to the discovery object, but which is than also stored in a container in that discovery object. But the way the portal protocol requires access to discv5, I didn't have another solution for now (also must be ref objects because of this).

This could be solved however by using an `AsyncQueue`, which additionally allow you to limit the messages in "flight". So I might go that route eventually.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -77,6 +77,7 @@ build_script:
 test_script:
   - nimble test
   - nimble build_dcli
+  - nimble build_portalcli
 
 deploy: off
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -237,4 +237,4 @@ jobs:
           nimble install -y --depsOnly
           nimble test
           nimble build_dcli
-
+          nimble build_portalcli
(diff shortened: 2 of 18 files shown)


kdeme:
Some interop testing with ddht and trin was done.

ddht:
- ping/pong works. Both directions
- findnodes/nodes works. But limited test in sense that no actual nodes are returned (but itself), as the network is a 1 - 1 network.
- findcontent -> Fails currently, ddht doesn't return anything

trin:
- ping/pong works. Both directions
- Others return invalid data

Additional oddity, on base layer, that is discovery v5, empty talkreq messages return some filled talkresp with trin. This should return an empty talkresp.

kdeme:
Update:
- ddht findcontent fixed on their side
- trin findnodes works

PR #367, opened 2021-06-17
Allow discoveryv5 to listen on IPv6

bauerj:
This is my attempt at solving status-im/nimbus-eth2#2210. Currently, it's not possible to supply an IPv6 listen address. It fails like this:

    Error: unhandled exception: /home/pi/nimbus-eth2/vendor/nim-chronos/chronos/transports/datagram.nim(488, 13) 'remote.family == local.family' [AssertionError]

This is due to `newDatagramTransport` setting `remote` to `AnyAddress`, which is `0.0.0.0`.

Not sure if this is the best way or even the right way to fix this so I appreciate any feedback.

Closes: status-im/nimbus-eth2#2210


cheatfate:
This PR is just not enough to fully support IPv6 and IPv4.
We don't need a branch, we need to bind to both IPv6 and IPv6 ports (if IPv6 is enabled). Because you can't send IPv4 datagram using IPv6 socket (transport) as you can't send IPv6 datagram using IPv4 socket. That's why when parsing discovery5 requests you should respond carefully using appropriate transport, so for example discovery5 packet has both IPv4 and IPv6 address is set. Also i expect there more complexity because I'm not familiar with discovery5 specification.

kdeme (commented):
@bauerj Thank you for giving this a try.

However, as @cheatfate already mentioned, it will be a little more complex than only selecting between ipv4 or ipv6 listener.

The goal of IPv6 support would be that we should be able to allow the node to listen & send on both IPv4 and IPv6.

Now if the scope of this PR is to only be able to select between IPv4 OR IPv6, then this is a start, but it still needs work on the data sending (or perhaps we need to filter out IPv4 nodes  when adding them to the routing table, but I think I like this approach less, not sure yet). That is, the ENR of the Node can still hold an IPv4 address, so those have to be filtered out before trying to send on an IPv4 address. Adding some tests for this case would be nice too :).

More interesting of course (for the end user) is the IPv6 support where the combination of IPv4 + IPv6 is possible. This would require two listeners (Although I think on at least Linux you can bind to the IPv6 any-address and still get IPv4 traffic?), so the creation of two transports, and require them both to handle incoming requests and respond. Responses would have to be send to the right address, which would probably be one of the ENR announced addresses for that node instead of the actual source address of the request (I think the source address could be used to select whether we send back on IPv4 or IPv6 in case both are advertised in the ENR of the target).

bauerj:
Thank you for your feedback.

I can confirm that outgoing UDP packets to IPv4 addresses still get sent on my Linux machine like @kdeme said. This was meant as a minimal change to allow use-cases like in status-im/nimbus-eth2#2210 where a user wants to use only IPv6.

But yes, googling suggests that this doesn't work on any other OS so more work needs to be done.

I plan to change this PR to create two sockets as explained by @cheatfate and chose the right transport when sending datagrams.

As i am not yet familiar with this library, is there anything that would have to be done when I implement this?

eth/p2p/discoveryv5/protocol.nim
```diff
@@ -993,11 +1003,20 @@ proc newProtocol*(privKey: PrivateKey,
...
+    bindAddresses = d.bindAddresses
+
+  for bindAddress in d.bindAddresses:
+    let ta = initTAddress(bindAddress.ip, bindAddress.port)
+    if ta.family == AddressFamily.IPv4:
+      d.transports.add(newDatagramTransport(processClient, udata = d, local = ta))
+    elif ta.family == AddressFamily.IPv6:
+      let localSock = AsyncFD(createNativeSocket(ta.getDomain(), SockType.SOCK_DGRAM,
+                                    nativesockets.Protocol.IPPROTO_UDP))
+      discard localSock.setSockOpt(posix.IPPROTO_IPV6, posix.IPV6_V6ONLY, 1)
```
bauerj, line 1015: If this isn't set, binding the socket will fail with "Address already in use", since the IPv6 socket would also work for IPv4 traffic.

Not sure if this is too low-level for this library. Should I instead implement this in nim-chronos?

eth/p2p/discoveryv5/protocol.nim
```diff
@@ -218,22 +219,29 @@ proc updateRecord*(
...
-      # it should not stop the process as that would reset the discovery
-      # progress in case there is even a small window of no connection.
-      # One case that needs this error available upwards is when revalidating
-      # nodes. Else the revalidation might end up clearing the routing tabl
-      # because of ping failures due to own network connection failure.
-      warn "Discovery send failed", msg = f.readError.msg
+  var errors: seq[string]
+  for transp in d.transports:
+    try:
+      await transp.sendTo(ta, data)
```
bauerj, line 227: Since there could be multiple source addresses of the same family (e.g. `10.0.9.10` and `200.1.55.2`), there is no way to know which transport to use, other than trying both.

bauerj:
@kdeme @cheatfate 

I updated the code as per your feedback. It seems to work but the part with `setSockOpt` explained above looks ugly. Please let me know if there's a better way to do it.

I will look into adding tests later.

Also, is there anything else that would have to be changed for IPv6 support in this library apart from discoveryv5?

arnetheduck:
Actually, we can in most cases use an IPv6 socket for IPv4 traffic - the way to do it is to use an dual-stack socket and ipv4-mapped addresses - this works on all major OS's

In short, it works like this:

* first, try to create an ipv6 socket - if this succeeds, set it to dual-stack mode by explicitly disabling IPV6_ONLY with setsockopt (see https://stackoverflow.com/questions/1618240/how-to-support-both-ipv4-and-ipv6-connections)
* if ipv6 socket creation fails, create an ipv4 socket instead

http://long.ccaba.upc.edu/long/045Guidelines/eva/ipv6.html has some examples, though they use `getaddrinfo` which we should probably avoid for now - the general principle holds though - try to bind to ipv6, if that fails fall back to ipv4. 

When sending data, there are a few things that can happen:

* same socket as target address: just send
* ipv4 target, ipv6 socket: send using mapped address
* ipv6 target, ipv4 socket: check if it's a mapped ipv4 address, if yes, unmap and send otherwise fail

Sorry for the ping-pong with one vs many sockets - this is a tricky and annoying problem.

bauerj:
Thank you for weighing in on this.

I'm a bit confused now. Your proposed solution sounds similar to what I initially suggested and directly opposite to what cheatfate and kdeme commented.

A few remarks from my side:

* `IPV6_V6ONLY` seems to be off by default, which is why I had to enable it for this two sockets approach
* Remapping IPv4 addresses to IPv6 seems to happen automatically for sockets where `IPV6_V6ONLY` is off

Did you find any information on OS support for `IPV6_V6ONLY`? I could find references in the Linux and Windows manuals but I'm not sure if it works on FreeBSD and macOS.

Also, should it be possible for the user to supply bind addresses? If so, we would need an approach with multiple sockets anyway. In that case it might be simpler to keep the current approach to prevent yet more if/else branches.

arnetheduck:
> IPV6_V6ONLY seems to be off by default, which is why I had to enable it for this two sockets approach

The default is system and os specific - better set it explicitly.

> support

Wikipedia notes:
> While dual-stack is supported by major operating system and network device vendors, legacy networking hardware and servers don't support IPv6.

Windows for example gained support in Vista: https://docs.microsoft.com/en-us/windows/win32/winsock/dual-stack-sockets
Mac has had dual stack support for a long time afair, as is freebsd (https://people.freebsd.org/~rodrigc/doc/data/ipv6/ipv6only.html)

I think it's safe to assume we can rely on dual stack even if the mode proposed by @cheatfate / @kdeme might be slightly more flexible - the complexity of dealing with two sockets does not outweigh the benefit IMO, but I'm not married to this position.

> Also, should it be possible for the user to supply bind addresses

This is fine - the same thing applies here: try to parse as ipv6, if that doesn't work, parse as ipv4. When done parsing, create an ipv6 socket (in case we want to send), then ipv4 if that doesn't work, then proceed as above.

kdeme:
> Sorry for the ping-pong with one vs many sockets - this is a tricky and annoying problem.

Right, figuring the best way out here is mostly the reason why it wasn't done yet in the first place. Requires a bit of investigation, I only know somewhat the behaviour on Linux.

> Actually, we can in most cases use an IPv6 socket for IPv4 traffic - the way to do it is to use an dual-stack socket and ipv4-mapped addresses - this works on all major OS's

One of the items I wasn't sure of, but yes, apparently Windows has it also since Vista: https://docs.microsoft.com/en-us/windows/win32/winsock/dual-stack-sockets?redirectedfrom=MSDN


> I'm a bit confused now. Your proposed solution sounds similar to what I initially suggested and directly opposite to what cheatfate and kdeme commented.

Yes, sorry for that confusion, development is a process ;-) . What @arnetheduck mentions is indeed similar. However, the initial PR was still missing the proper handling for sending the messages, and also missing the advertisement of both (IPv4 and IPv6) addresses in the ENR. 
Anyhow, as mentioned before, I wasn't too sure of this solution cross platform.
But it is also not clear in this solution on what to do when a user wants to specifically provide certain interfaces to bind to.


> Also, should it be possible for the user to supply bind addresses? If so, we would need an approach with multiple sockets anyway. In that case it might be simpler to keep the current approach to prevent yet more if/else branches.

Right, as mentioned above, that is an issue. Ideally we do all over 1 socket of course. But I think there might be the use case of defining the specific IP addresses of the interfaces users want to use.

bauerj:
> This is fine - the same thing applies here: try to parse as ipv6, if that doesn't work, parse as ipv4. When done parsing, create an ipv6 socket (in case we want to send), then ipv4 if that doesn't work, then proceed as above, binding to whatever the user selected either via mapped address or native.

Okay, but what if the user supplies these bind addresses:

* ::
* 10.0.1.1

Should this work? I assume this will fail because the IPv6 socket already blocks the port on 10.0.1.1.

bauerj:
I have changed it back to using only a single transport. Depending on bindIp it's either:

- IPv4 only if the bindIp is IPv4
- IPv6 only if the bindIp is IPv6
- Dualstack if the bindIp is none. Outgoing datagrams will be sent to IPv6 if the other node has a known IPv6 address, otherwise IPv4

This PR now also:

- adds the public IPv6 address to the ENR
- takes care of IPv4-in-IPv6 wrapped addresses
- explicitly disables IPV6_V6ONLY on the socket
- adds a new instance of IpVote for IPv6 only.
- sets the TCP sockets in discovery.nim and p2p.nim to dualstack mode.

I tested this on two VMs of mine and was able to see IPv6 traffic between the nodes! There are also some IPv4 connections so I guess I didn't break that.

eth/p2p/discoveryv5/protocol.nim
```diff
@@ -236,8 +240,26 @@ proc send(d: Protocol, a: Address, data: seq[byte]) =
...
+  try:
+    localTransportIpv6 = d.transp.localAddress.family == AddressFamily.IPv6
+  except TransportOsError: discard
+  
+  if n.address6.isSome() and localTransportIpv6:
+    # If we have IPv6 and they have it, let's use that.
+    d.send(n.address6.get(), data)
+  else:
+    if localTransportIpv6:
+      let unwrapped = n.address.get().wrapIPv4InIPv6()
```
kdeme, line 255: Question: Can we know here that we have a dual stack socket, and thus that this message can actually be send?

eth/p2p/discoveryv5/protocol.nim
```diff
@@ -236,8 +240,26 @@ proc send(d: Protocol, a: Address, data: seq[byte]) =
       warn "Discovery send failed", msg = f.readError.msg
 
 proc send(d: Protocol, n: Node, data: seq[byte]) =
-  doAssert(n.address.isSome())
-  d.send(n.address.get(), data)
+  doAssert(n.address.isSome() or n.address6.isSome())
+
+  var localTransportIpv6: bool
+  try:
+    localTransportIpv6 = d.transp.localAddress.family == AddressFamily.IPv6
```
kdeme, line 247: I guess it would be nicer if we had the information if we are dealing with IPv6+IPv4, IPv4 only or IPv6 only socket since the transport was created.

eth/net/utils.nim
```diff
@@ -65,3 +63,65 @@ proc getRouteIpv4*(): Result[ValidIpAddress, cstring] =
...
+  if ta.family != AddressFamily.IPv6:
+    return false
+  for i in 0..9:
+   if ta.address_v6[i] != 0x0:
+     return false
+  if ta.address_v6[10] != 0xff or ta.address_v6[11] != 0xff:
+    return false
+  return true
+
+proc unwrapIPv4InIPv6*(ta: TransportAddress): TransportAddress =
```
kdeme, line 94: similar functionality already exists in chronos: https://github.com/status-im/nim-chronos/blob/aab1e30a726bb47c5d3f4a75a826981836cde9e2/chronos/transports/ipnet.nim#L195

eth/net/utils.nim
```diff
@@ -65,3 +63,65 @@ proc getRouteIpv4*(): Result[ValidIpAddress, cstring] =
...
+  address_v4[1] = ta.address_v6[13]
+  address_v4[2] = ta.address_v6[14]
+  address_v4[3] = ta.address_v6[15]
+  return TransportAddress(
+    family: AddressFamily.IPv4,
+    address_v4: address_v4,
+    port: ta.port
+  )
+
+proc wrapIPv4InIPv6*(ta: TransportAddress): TransportAddress =
```
kdeme, line 107: similar functionality already exists in chronos: https://github.com/status-im/nim-chronos/blob/aab1e30a726bb47c5d3f4a75a826981836cde9e2/chronos/transports/ipnet.nim#L168

kdeme:
@bauerj Sorry I lost track of this and didn't get back to your last question. Thanks for continuing on it.

> * IPv4 only if the `bindIp` is IPv4
> 
> * IPv6 only if the `bindIp` is IPv6
> 
> * Dualstack if the `bindIp` is none. Outgoing datagrams will be sent to IPv6 if the other node has a known IPv6 address, otherwise IPv4

Yeah, that practise looks fine to me.
I've added some comments though in the review:
a. it should be cross platform
b. can we store this information, so it can be used for example at the send call.

If you'd like, you can also make already a PR with some of the supporting parts (like enr, node, etc. changes) to get those in already.


>     * takes care of IPv4-in-IPv6 wrapped addresses
I'd prefer if we use the existing chronos calls here.

Then something else to note is that the `routing_table` code does not support the `node.address6` part. Practically this means no ipv6 only nodes will ever end up in the routing table. This is no concern for this PR.

eth/p2p/discoveryv5/protocol.nim
```diff
@@ -981,24 +1024,48 @@ proc newProtocol*(privKey: PrivateKey,
...
+
+  if d.bindIp.isNone:
+    ## The default strategy is to use an IPv6 socket in dual stack mode which
+    ## "can be used to send and receive packets to and from an IPv6 address or
+    ## an IPv4-mapped IPv6 address."
+    ta = initTAddress(IPv6_any(), d.bindPort)
+    let nativeSock = createNativeSocket(ta.getDomain(), SockType.SOCK_DGRAM,
+                                  nativesockets.Protocol.IPPROTO_UDP)
+    let asyncSock = AsyncFD(nativeSock)
+    let dualstack = asyncSock.setSockOpt(posix.IPPROTO_IPV6, posix.IPV6_V6ONLY, 0)
```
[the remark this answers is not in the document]
bauerj: > Could the newDatagramTransport6 be used, and catch the error if it fails?

I'm not exactly sure what you mean. `setsockopt` is used to make the socket dualstack. Otherwise it would depend on the OS preset which may or may not be what we want.

There may be a high-level option in nim-chronos that also does this but I was unable to find it.

What makes you think `setsockopt` is not cross-platform? I was able to find the same options in [Windows](https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-ipv6-socket-options) and [FreeBSD](https://www.freebsd.org/cgi/man.cgi?query=ip6&sektion=4&manpath=FreeBSD+9.0-RELEASE) docs, so at least the most important OSs support it 😀

eth/p2p/discoveryv5/protocol.nim
bauerj: Not yet but I will add a flag to the `Protocol` as you suggested. It should probably fail here if it's impossible to reach the target.

eth/net/utils.nim
bauerj: 🥲

bauerj:
Yes, I will definitely change it to use the functions integrated in nim-chronos. I'll consider my redundant attempt a valuable lesson in Nim 😅

I'm not sure if this is really platform-dependent and would appreciate further information on that. Worst that could happen on an exotic OS is it would use IPv4 only, if I understand this right.

I'll have a look at the routing_table code.

eth/p2p/discoveryv5/protocol.nim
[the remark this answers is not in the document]
kdeme: > What makes you think `setsockopt` is not cross-platform? 
I was referring to the use of posix. Imports will have to differ depending on OS, see for example in chronos: https://github.com/status-im/nim-chronos/blob/d270dba8a3b8f91cc263fb4a8d92b8e729761876/chronos/transports/datagram.nim#L16
That's why ideally we can just use directly chronos functionality.






> There may be a high-level option in nim-chronos that also does this but I was unable to find it.

But yes, I haven't checked yet if this is available.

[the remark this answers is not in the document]
bauerj: Okay, I'll check that. I'm not exactly sure how to do that with `winlean`, since it doesn't expose the constants needed but I'll read through the nim-chronos code.

bauerj:
Quick status update:

I have incorporated most of your feedback already.

The code is now working on Windows as well, although I had to implement it without winlean: https://github.com/status-im/nim-eth/pull/367/commits/9181afd9537d1d67dfbfb0c9600027c24ceef956. Since this is so low-level now, I will look into moving this into chronos' `newDatagramTransport6`.

I have no idea how to solve the issue with `routing_table` yet. Any advice?

kdeme:
> I have no idea how to solve the issue with `routing_table` yet. Any advice?

I think it is a matter of:
1. Allowing [here](https://github.com/status-im/nim-eth/blob/master/eth/p2p/discoveryv5/routing_table.nim#L306) for both address options.
2. Reworking `ipLimitInc` and `ipLimitDec` to add either IPv4, IPv6 or both, depending on what the node has: https://github.com/status-im/nim-eth/blob/master/eth/p2p/discoveryv5/routing_table.nim#L137

I do realize that a user with IPv4 and IPv6 capabilities can suddenly double its routing table presence. But I don't think there is much to do against that.

bauerj:
Hi,

thanks for the heads-up. I currently don't plan to implement this.

PR #368, opened 2021-06-25
implement EIP-1559 Transaction and BlockHeader

jangko:


eth/common/eth_types.nim
```diff
@@ -113,6 +130,10 @@ type
     mixDigest*:     Hash256
     nonce*:         BlockNonce
 
+  BlockHeader* = object
+    header*:        LegacyHeader
+    baseFee*:       Option[UInt256]
+
```
jlokier, line 136: I don't think this representation is useful or helpful here - same for the long list of accessor templates that goes with it.  The extra indirections seem like complication without benefit here. Is there a reason for it that I have missed?

In the transaction types we have `isContractCreate: bool` representing different transaction encodings.  That's fine even though it's not a field in the raw RLP-encoded transaction.

The same applies here: I prefer we add `baseFee` as an option field to `BlockHeader`, no need for another header type. 

 If there was a different meaning between the three cases `LegacyHeader`, `BlockHeader` with none for `baseFee` and `BlockHeader` with some `baseFee` I could see the rationale but there don't seem to be 3 meaningful states. The RLP encoder shows only the last two are meaningful.
 
 (This reminds me of the comments at https://github.com/status-im/nim-quic/pull/27#discussion_r651648773, about multiple states that mean the same thing but get processed by different code paths, which creates more opportunities for problems. If there was a meaningful distinction between `LegacyHeader` and `BlockHeader` in this PR, I would expect `baseFee` to be `UInt256` not `Option[UIint256]` here.)
  
The Nim objects exist to provide simple, easy to understand, Nim-friendly and _uniform_ access to the fields in Eth objects.  Ideally fast, direct access.  The Nim objects don't need to exactly match the detailed RLP-encoding variants for the same Eth type, when the variants are really just the same Eth object, with the same fields (except one or two options), with all those fields used the same way almost everywhere. Imagine if `isContractCreate` were implemented as variant types `CreateContractTx` and `CallTx` instead. Then you would have lots of run-time checking accessors called in a lot of code instead of direct field access, and you would have _twelve_ transaction type variants by now in EIP1559, all very similar to each other. It's a combinatorial explosion, which doesn't make the code any clearer or simpler.

Small differences in size don't matter. (When in-memory database size matters (as it does in Geth/OpenEthereum for their pruning algorithms) RLP is better than variants.)  Simplicity and avoiding code paths that access the same named fields slightly differently are important, because nearly identical variants creates new branches that aren't really being exercised, and compiles to slower and larger run-time code for no benefit that I can see.  For example, the RLP decoder for lists of EIP2718 transactions was not tested when it was added, because it just didn't get used.

It's a judgement call of coures.  I would make the opposite argument if the types were significantly different, or if it was about saving bytes. But here I think the difference between pre-EIP1559 and post-EIP-1559 blocks is one optional field, not different kinds of object. Almost every part of the code doesn't care about the difference, and this is reflected in the way `LegacyHeader` won't be used by any of it, and needing a long list of field accessor templates that don't do anything. I think the same is true of the EIP2930/EIP1559 transaction and receipt variants: Almost every part of the code wants to use these objects the same way, and the differences are very small, just one or two optional fields.

jlokier:
I prefer that we hold off on merging this at the moment, unless you depend specifically on its functionality to move something else forward.  However I don't actively object because it looks like it works and I don't see any new bugs.

- The RLP changes are in an area that's most likely the reason why real blocks from the network (mainnet, Goerli) results in RLP parse errors at the moment. There's a bug probably in the EIP2718 decoding. Still under investigation.

- The proliferation of duplicate object types with very slightly different fields complicates transactions, receipts and now blocks. I've given my thoughts with an inline comment at the new definition of `BlockHeader`, but it applies to the new transaction variants as well.  When there was only `AccessListTx` added, it solved a problem (though some code was added that never got used or tested) and that didn't seem so large, but it's starting to resemble a combinatorial explosition by now.

On the second point, although I don't agree with the direction, I understand you may be building things on top that are making progress, and I don't object to merging if you still think it's useful after reading my review.  I don't see any (new) bugs.  If you do decide to merge, I'll be looking to make a PR that combines the RLP-encoding-variant types back together again to their original typers :) to reduce the combinatorial explosition, using option fields or flags like `isContractCreate`.

jlokier (commented):
I prefer that we hold off on merging this at the moment, unless you depend specifically on its functionality to move something else forward.  However I don't actively object because it looks like it works and I don't see any new bugs.

- The RLP changes are in an area that's most likely the reason why real blocks from the network (mainnet, Goerli) results in RLP parse errors at the moment. There's a bug probably in the EIP2718 decoding. Still under investigation.

- The proliferation of duplicate object types with very slightly different fields complicates transactions, receipts and now blocks. I've given my thoughts with an inline comment at the new definition of `BlockHeader`, but it applies to the new transaction variants as well.  When there was only `AccessListTx` added, it solved a problem (though some code was added that never got used or tested) and that didn't seem so large, but it's starting to resemble a combinatorial explosition by now.

On the second point, although I don't agree with the direction, I understand you may be building things on top that are making progress, and I don't object to merging if you still think it's useful after reading my review.  I don't see any (new) bugs.  If you do decide to merge, I'll be looking to make a PR that combines the RLP-encoding-variant types back together again to their original typers :) to reduce the combinatorial explosition, using option fields or flags like `isContractCreate`.

eth/common/eth_types.nim
jangko: I agree this has become a horrible mess. When i try to avoid custom rlp de/serialization, it only make things worse.
And because we already using custom serialization for tx/receipt/blockHeader, further customization actually doesn't  matter.

I'll rewrite them back to simpler types instead of this combinatorial explosion types.  And in the process we can write more controllable rlp encoder-decoder with more meaningful error message.

Reminder for myself: EIP-658, EIP-2718, EIP-2930, and EIP-1559.

PR #369, opened 2021-07-02
metrics: initialise lock for custom collector

stefantalpalaru:


jakubgs:
So this is the fix for the Windows Segmentation fault described in https://github.com/status-im/nimbus-eth2/issues/2666?

Damn, such a tiny oversight. But why would it work fine on other platforms?

stefantalpalaru:
Yeah, this is the fix.

Linux locks work just fine without being initialised, because the Nim GC zeroes allocated memory and `pthread_mutex_init()` happens to do the same thing.

On Windows it's different, probably because `InitializeCriticalSection()` needs to create some file handles to put in that data structure.

jakubgs:
Very peculiar. Thanks for explaining.

jakubgs:
Are you going to request any reviews for this?

eth/db/kvstore_sqlite3.nim
```diff
@@ -561,6 +561,7 @@ when defined(metrics):
                         help: help,
                         typ: "gauge",
                         creationThreadId: getThreadId())
+    result.lock.initLock()
```
arnetheduck, line 564: wouldn't this make more sense in .. `register`? having to initialize an implementation detail field at the point of usage sounds like a bug source that will keep on giving

stefantalpalaru: Sure, but a collector can be registered and unregistered more than once, so a new flag field is needed in the object.

arnetheduck: seems like a win regardless? it's an incredibly dumb bug to repeat, ever

zah: This looks like a problem stemming from Nim's poor support for inheritance (namely, the missing provisions for enforcing that a "constructor" of a base object will be executed while constructing a derived object). The `register` work-around sounds like something that can be used to make the API less error-prone, but it is also possible to create a general solution for this problem by defining some new syntax for constructing derived objects (stew/objects can define a macro for this):

```nim
result = construct SqliteInfo(Gauge.init(name, help), derivedField = "foo")
```

arnetheduck: well, because of said .. lack .. my preferred way for nim objects is to work out-of-the box without any construction - ie a Table shouldn't require `newTable` and ideally a lock (at the Nim level) shouldn't require initLock or whatever it's called - that said, happy to see something done about construction as well

PR #374, opened 2021-07-15
Add top level push raises Defect to p2p code

kdeme:
@oskarth This seems to work.

But it can't be merged as it will currently break nimbus-eth2


oskarth:
Thanks @kdeme!

PR #376, opened 2021-07-19
Port merkleization off ssz

KonradStaniec:
Merkelization of ssz is needed for https://github.com/status-im/nimbus-eth1/issues/756.

There is an effort to port whole ssz to separate library https://github.com/status-im/nim-ssz-serialization/pull/1, but to finish this it would require rethinking design of the library. For now this pr follows suit of the rest of ssz to make progress on fluffy front.


kdeme (approved):
If we are sticking with this solution for a while to come, we better add some tests too in the near future

KonradStaniec:
Created separate issue for this - https://github.com/status-im/nim-eth/issues/377

PR #378, opened 2021-07-20
Add merkleization tests

KonradStaniec:
Add merkleization test cases from - https://github.com/ethereum/eth2.0-specs/blob/dev/tests/core/pyspec/eth2spec/utils/test_merkle_minimal.py


eth.nimble
```diff
@@ -66,6 +66,9 @@ task test_trie, "Run trie tests":
 task test_db, "Run db tests":
   runTest("tests/db/all_tests")
 
+task test_ssz, "Run ssz tests":
+  runTest("tests/ssz/merkleization")
```
kdeme, line 70: I think we have the unwritten rule to prefix the filename of tests with `test_`

tests/ssz/merkleization.nim
```diff
@@ -0,0 +1,75 @@
...
+  newSeqWith(28, elem) & v.toBytesLe().toSeq()
+
+proc z(i: int): seq[byte] =
+  zerohashes[i].data.toSeq()
+
+type TestData[limit: static int64] = object
+  count: uint32
+  expectedRoot: seq[byte]
+
+# only happy cases as our merkleizer will happy accept more chunks than limit
```
kdeme, line 23: What does this mean in practise? It continues hashing those chunks or something? What should happen, ignore those or abort?
Can this be abused?

kdeme (approved):

tests/ssz/merkleization.nim
KonradStaniec: it continues hashing those chunks and produces invalid final hash as limit influences its computation. So if you create merkelizer with `limit = 2` and add `3` chunks, you will be able to get final hash but it will be wrong.

I am not sure it can be abused, but it is a rather brittle api which can produce silent incorrect result. Imo two things could happen:
- either return error when adding chunks, if added chunks is over the limit
- or return error when generating final hash

My slight preference would to be to return `Result` from final hash computation and in case of such condition return error to the caller.

PR #379, opened 2021-07-23
add merkle proof verification

KonradStaniec:
Merkle proof verification will be need, to validate proof received with block headers in fluffy history network.

### DIFF
--- a/eth/ssz/merkleization.nim
+++ b/eth/ssz/merkleization.nim
@@ -12,6 +12,7 @@
 {.push raises: [Defect].}
 
 import
+  math,
   stew/[bitops2, endians2, ptrops],
   stew/ranges/ptr_arith, nimcrypto/[hash, sha2],
   serialization/testing/tracing,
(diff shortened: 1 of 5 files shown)


eth/ssz/merkleization.nim
```diff
@@ -609,3 +610,34 @@ func hash_tree_root*(x: auto): Digest {.raises: [Defect].} =
...
+func isValidProof*(leaf: Digest, proof: openArray[Digest],
+                             index: uint64, root: Digest): bool =
+  if len(proof) == getGeneralizedIndexLength(index):
+    var
+      value = leaf
+      buf: array[64, byte]
+
+    for i, digest in proof:
+      if getGeneralizedIndexBit(index, uint64 i):
+        buf[0..31] = digest.data
```
arnetheduck, line 632: this copy looks redundant, why not `h.update` straight from `digest.data`?

KonradStaniec: yep you are right, I can use `mergeBranches`  which does exactly that :+1:

eth/ssz/merkleization.nim
```diff
@@ -618,25 +618,19 @@ func getGeneralizedIndexLength(x: uint64): int =
...
   if len(proof) == getGeneralizedIndexLength(index):
     var
       value = leaf
-      buf: array[64, byte]
 
     for i, digest in proof:
       if getGeneralizedIndexBit(index, uint64 i):
-        buf[0..31] = digest.data
-        buf[32..63] = value.data
+        value = mergeBranches(digest, value)
```
arnetheduck, line 631: this can further be simplified to `value = if ... : merge.. ; else: merge ...`

arnetheduck (approved):

PR #385, opened 2021-08-01
Allow anonymous data proc for kvstore get

etan-k:
Allow anonymous data proc for kvstore get

Nim 1.2.12 has trouble compiling kvstore clients that attempt to fetch
data by passing an anonymous proc to `get`, i.e.,
```
    db.get(key, proc(data: openArray[byte]) = v = @data)[]
```
leads to a compilation error such as:
```
    internal error: expr: param not init data_34695004
```

Note that defining the proc in a named fashion, then passing the name,
works fine. Nim 1.4.8 also succeeds in the case of an anonymous proc.

The error seems to show up due to an interaction with templates taking
`openArray` procs. As the kvstore functions do not need to use templates
this patch replaces them with inline procs. Furthermore, an additional
test checks that passing an anonymous proc works as expected.

A minimal standalone program triggering the same compilation problem:

```
    type
      X* = ref object
        bar: proc (p: proc(v: openArray[byte]))

    template foo(x: X, p: proc(v: openArray[byte])) =
      x.bar(p)

    X().foo(proc(v: openArray[byte]) = discard @v)
```

See https://github.com/nim-lang/Nim/issues/18648

### DIFF
--- a/eth/db/kvstore.nim
+++ b/eth/db/kvstore.nim
@@ -50,22 +50,23 @@ template put*(dbParam: KvStoreRef, key, val: openArray[byte]): KvResult[void] =
   let db = dbParam
   db.putProc(db.obj, key, val)
 
-template get*(dbParam: KvStoreRef, key: openArray[byte], onData: untyped): KvResult[bool] =
+# template would be preferred, but cannot be called with an anonymous proc on Nim 1.2.12.
+# See https://github.com/nim-lang/Nim/issues/18648
+proc get*(db: KvStoreRef, key: openArray[byte], onData: DataProc): KvResult[bool] =
   ## Retrive value at ``key`` and call ``onData`` with the value. The data is
   ## valid for the duration of the callback.
   ## ``onData``: ``proc(data: openArray[byte])``
   ## returns true if found and false otherwise.
-  let db = dbParam
   db.getProc(db.obj, key, onData)
 
-template find*(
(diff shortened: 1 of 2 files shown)


eth/db/kvstore.nim
```diff
@@ -45,42 +45,35 @@ type
     containsProc: ContainsProc
     closeProc: CloseProc
 
-template put*(dbParam: KvStoreRef, key, val: openArray[byte]): KvResult[void] =
+proc put*(db: KvStoreRef, key, val: openArray[byte]): KvResult[void] {.inline.} =
```
arnetheduck, line 48: we tend to avoid `inline`, which in Nim simply is a slow and noisy subset of LTO

arnetheduck:
Is there an upstream bug reference for this?

eth/db/kvstore.nim
etan-k: Would you prefer just `proc` without `inline`?
Or to stick with `template` for the functions not taking the `openArray` callbacks?

arnetheduck: I think it might be reasoonable to use `proc` without inline for the openArray callbacks and `template` for the rest - what we do in such cases is to link to an upstream bug to explain the inconsistency / constraint.

etan-k:
> Is there an upstream bug reference for this?

Reported upstream as https://github.com/nim-lang/Nim/issues/18648

arnetheduck:
seems like we got by without this, and since we're on our way to 1.6, seems obsolete?

PR #386, opened 2021-08-02
Delete Whisper remains (moved to nim-waku)

oskarth:
Types, protocols, tests, hardcoded nodes, references.

Now part of nim-waku:
- https://github.com/status-im/nim-waku/pull/663
- https://github.com/status-im/nim-waku/pull/678


jm-clius (approved):

jlokier:
What luck, I was about to delete Whisper from `nim-eth` :-)  I hadn't realised you'd copied it over to `nim-waku`, I was just going to nuke it.

I removed another part of Whisper implementation in https://github.com/status-im/nimbus-eth1/pull/705, and as you can see from @arnetheduck's comment, it's ok to drop Whisper generally, it's obsolete and hasn't been used in a couple of years.

Aside from the types file (which can probably be trimmed if current Waku doesn't use all of it), is Waku still using Whisper/`shh` at all?  If not, rather than keeping the `shh` protocol alive in `nim-waku`, I suggest it's time to just delete it.  It will always be in the repo if someone wants to dig it out again.  Each fewer thing that depends on RLPx makes it that bit easier to update RLPx/p2pProtocol.

oskarth:
>  Each fewer thing that depends on RLPx makes it that bit easier to update RLPx/p2pProtocol.

Agree. Unfortunately Waku v1 (https://rfc.vac.dev/spec/6/) is still using RLPx, and it is also needed for the bridge between Waku v1 and Waku v2. Once the whole network has migrated to Waku v2 and Waku v1 is nuked completely and spec is deprecated (will probably take a while, judging by Waku v0 lifecycle, basically Whisper, https://rfc.vac.dev/spec/5/) we can remove it completely :)

There is a lot of duplication between Whisper and Waku v1, but for now we can leave it be. Then refactor as necessary in nim-waku repo.

tests/p2p/test_shh.nim
```diff
@@ -1,382 +0,0 @@
-#
```
kdeme, line 1: This test should get activated in nim-waku, I think it all whisper tests are deactivated there for CI time, but this code is actually common and should thus get tested.

tests/fuzzing/rlpx/thunk.nim
```diff
@@ -1,7 +1,7 @@
 import
   testutils/fuzzing, chronos,
   eth/p2p, eth/p2p/rlpx, eth/p2p/private/p2p_types,
-  eth/p2p/rlpx_protocols/[whisper_protocol, eth_protocol],
```
kdeme, line 4: While I agree that this needs to be removed here, the fuzzing tests would be useful for Waku also, but no concern of this PR of course.

kdeme (approved):
Agree on removal of this code. Just the remark that it would be wise to activate that one whisper types test in nim-waku

oskarth:
Rebased this old PR per request, @arnetheduck @kdeme 

CI seems extremely slow

@jm-clius haven't moved over fuzzing stuff, might be of interest to nwaku

kdeme:
> @zah @oskarth anything outstanding here in terms of the cleanup?

I think we are all good now.

PR #389, opened 2021-08-03
Add option to `connectToNetwork` return without waiting for peers

jlokier:
The new sync code wants to start without waiting.  We can `discard` the async result but there is no need for a background task polling and running a timer for no clear benefit.


kdeme (approved):

zah:
Merged manually by cherry-picking.

PR #394, opened 2021-09-07
[WIP] Implementing uTP sub-protocol

austinlhx:
This is a very very rough draft of adding uTP to the fluffy client. This is a very rough draft PR where I am looking to get recommendations on how I should implement certain parts and if there are any recommended modules that I should use. Kim from the team contacted me a couple days ago and I am just providing with what I have completed so far. I am still learning Nim so please let me know if there are any things that I should fix! As of now I am currently still working on how I should go about encoding the packets that are sent through uTP. I am using the stew/endian2 package to convert them, if there are any better options, please let me know as I am looking for any suggestions!

### DIFF
--- a/eth/p2p/discoveryv5/utp.nim
+++ b/eth/p2p/discoveryv5/utp.nim
@@ -1,7 +1,9 @@
-import times
+import times, math
+import nim-stew/stew/endians2
+import endians
 
 const
-  UTP_PROTOCOL = "utp"
+  uTP_PROTOCOL = "utp"
   HEADER_SIZE = 20
   MAX_DISCV5_PACKET_SIZE = 1280
   MIN_PACKET_SIZE = 150
@@ -13,13 +15,13 @@ const
   VERSION = 1
 
 type
-  UTPStates = enum
-    UTP_STATE_CONNECT = 1,
+  UTPStates* = enum #* just means its public
+    UTP_STATE_CONNECT = 1, 
     UTP_STATE_WRITABLE = 2,
     UTP_STATE_EOF = 3,
     UTP_STATE_DESTROYING = 4
 
-  ConnectionState = enum
+  ConnectionState* = enum
     ST_DATA = 0
     ST_FIN = 1
     ST_STATE = 2
@@ -27,19 +29,17 @@ type
     ST_SYN = 4
     ST_NUM_STATES
 
-  PacketHeader = object
-    packet_type: ConnectionState
+  PacketHeader = ref object
     type_ver: uint8
+    extensions: uint8
     connection_id: uint16
     timestamp: uint32
     timestamp_diff: uint32
     wnd_size: uint32
     seq_nr: uint16
     ack_nr: uint16
-    extensions: string
-    data: string
 
-  UTPProtocol = object
+  UTPProtocol = ref object
     state: ConnectionState
     seq_nr: uint16
     ack_nr: uint16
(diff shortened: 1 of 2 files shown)


eth/p2p/discoveryv5/utp.nim
```diff
@@ -0,0 +1,234 @@
...
+  ConnectionState* = enum
+    ST_DATA = 0
+    ST_FIN = 1
+    ST_STATE = 2
+    ST_RESET = 3
+    ST_SYN = 4
+    ST_NUM_STATES
+
+  PacketHeader = ref object
+    type_ver: uint8
```
kdeme, line 33: In general we follow NEP-1 style, see also our style guide: https://status-im.github.io/nim-style-guide/formatting.style.html

However, there are cases where we deviate from this. So we can discuss what are the benefits of sticking with another style are. I know the spec uses some snake case, but I don't think that is compelling enough to follow that style. However the c/c++ libs probably do too, so if we are following/porting some implementation closely, that might be good enough reason.

eth/p2p/discoveryv5/utp.nim
```diff
@@ -0,0 +1,234 @@
...
+proc set_version(data: PacketHeader, new_type: uint8) =
+  data.type_ver = new_type
+
+proc get_type(data: PacketHeader) =
+  return data.type_ver
+
+proc set_type(data: PacketHeader, i: uint8) = #need to pass var uint8 if we want to mutate it
+  data.type_ver = (data.type_ver & 0xf) | (i << 4)
+
+proc set_connection_id(data: PacketHeader, connection_id: uint16) =
```
kdeme, line 153: In Nim we usually don't create setters and getters for every field, but instead just access those fields directly.

And in this specific case, I don't think you need to set the connection_id, or any other fields, of the PacketHeader. You'd probably create a full new header each time with for example an init or new call

eth/p2p/discoveryv5/utp.nim
```diff
@@ -0,0 +1,234 @@
...
+
+  ConnectionState* = enum
+    ST_DATA = 0
+    ST_FIN = 1
+    ST_STATE = 2
+    ST_RESET = 3
+    ST_SYN = 4
+    ST_NUM_STATES
+
+  PacketHeader = ref object
```
kdeme, line 32: It doesn't look like this needs to be a `ref` type. See also guideline: https://status-im.github.io/nim-style-guide/language.refobject.html

eth/p2p/discoveryv5/utp.nim
```diff
@@ -0,0 +1,234 @@
...
+                      seq_nr: toBytes(dcur  ata.seq_nr, bigEndian),
+                      ack_nr: toBytes(data.ack_nr, bigEndian))]#
+
+
+
+proc get_timestamp(): uint32 =
+  return (getTime() * 10000000) & 0xffffffff
+
+proc get_version(data: PacketHeader): uint8 = 
+  return data.type_ver
```
kdeme, line 142: You'll have to bitmask the `type_ver` value here to get the version

kdeme (commented):
@austinlhx Thanks for your interest in working on the uTP implementation in Nim and starting with this PR.

As I've mentioned to you in PM, @KonradStaniec has done some preliminary work past week on the uTP implementation and will get this in a draft PR. From what I understood, most of that work has been regarding serialization of packets, to be able to have a quick interaction with the reference implementation. But no work yet regarding congestion control logic.
I believe that the biggest bulk of the work will actually be the latter (+ testing).

I'd suggest to wait for that PR, and then we can assess better the different items of work for this protocol to work on still, and see who can work on what.

This PR will most likely collide with some of the work of @KonradStaniec, thus we will probably close it. Nevertheless, I gave it a small review, touching mostly Nim specific items that I spotted.
In general I advice you also to have a look at our Nim style guide, which holds a lot of useful information of how we (= the Nimbus team) develop in Nim: https://status-im.github.io/nim-style-guide/00_introduction.html

kdeme:
I will close this PR in favour of https://github.com/status-im/nim-eth/pull/397

Lets continue the development from there.

PR #396, opened 2021-09-07
Build fuzzing tests in CI and fix current fuzzing tests

kdeme:
Test in CI if fuzzer code builds to avoid bit rot.


kdeme:
CI is quite a bit slower due to the separate builds of the fuzzing tests now.

PR #397, opened 2021-09-08
Add initial skeleton of utp protocol

KonradStaniec:
- Serialization/deserialization of packets
- initial skeleton of utp protocol over udp
- test app to interact with reference impl

### DIFF
--- a/eth/utp/packets.nim
+++ b/eth/utp/packets.nim
@@ -1,3 +1,11 @@
+# Copyright (c) 2020-2021 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [Defect].}
+
 import
   std/[monotimes, random],
   faststreams,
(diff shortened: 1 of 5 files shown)


eth/utp/packets.nim
```diff
@@ -0,0 +1,153 @@
...
+
+proc encodePacket*(p: Packet): seq[byte] =
+  var mem = memoryOutput().s
+  try:
+    mem.write(encodeHeader(p.header))
+    if (len(p.payload) > 0):
+      mem.write(p.payload)
+    mem.getOutput()
+  except IOError:
+    # TODO not sure how writing to memory buffer could throw. Just swallow it for now
```
kdeme, line 95: idem ^

eth/utp/packets.nim
```diff
@@ -0,0 +1,153 @@
...
+    mem.write(h.extension)
+    mem.write(h.connectionId.toBytesBE())
+    mem.write(h.timestamp.toBytesBE())
+    mem.write(h.timestampDiff.toBytesBE())
+    mem.write(h.wndSize.toBytesBE())
+    mem.write(h.seqNr.toBytesBE())
+    mem.write(h.ackNr.toBytesBE())
+    return mem.getOutput()
+  except IOError:
+    # TODO not sure how writing to memory buffer could throw. Just swallow it for now
```
kdeme, line 84: Perhaps doAssert then for now?

eth/utp/packets.nim
```diff
@@ -0,0 +1,153 @@
+# Copyright (c) 2020-2021 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [Defect].}
+
+import
+  std/[monotimes, random],
```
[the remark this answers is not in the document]
KonradStaniec: hmm the problem with `Moment.now()` I had, is that I did not see any api to convert `Moment` type to `uint32`. Ideally we would have `fastEpochTimeNano` exposed as public function in `chronos/timer.nim`

As for `random2`, sure will do it 👍

kdeme (approved):

PR #398, opened 2021-09-10
Modify nodes verification

KonradStaniec:
Some times it can be valuable to disable distances validation, for example when we asked only for closest nodes. 
Until now when passing empty distances `verifyNodesRecords` would always return empty list. This pr, changes this so that empty distances list mean that distance is disabled.

### DIFF
--- a/eth/p2p/discoveryv5/nodes_verification.nim
+++ b/eth/p2p/discoveryv5/nodes_verification.nim
@@ -1,0 +1,85 @@
+{.push raises: [Defect].}
+
+import
+  std/[sets, options],
+  stew/results, stew/shims/net, chronicles, chronos, 
+  "."/[node, enr, routing_table]
+
+logScope:
+  topics = "nodes-verification"
+
+proc validIp(sender, address: IpAddress): bool =
+  let
+    s = initTAddress(sender, Port(0))
+    a = initTAddress(address, Port(0))
+  if a.isAnyLocal():
+    return false
+  if a.isMulticast():
+    return false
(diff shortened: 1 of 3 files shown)


kdeme (commented):
Not sure I understand the reasoning here.

The idea in discv5 is that you have to provide the distances (to the target node) for which you are searching nodes. Not to let the target node return its closest nodes (put that work on your own side basically, not the peer). This is different from default Kademlia.

Thus, normally a node will respond with an already empty response (or no response at all? The behaviour is not explicitly defined in spec afaik). So a better check would be to see if the response is actually empty when the list of distances was empty (but we should never send such request in the first place)

KonradStaniec:
@kdeme  Main reasoning was that in portal network, for find content queries node responds with either content or nodes closer to content, then the distances validation does not have much sense (but other do, like ips, and duplicates) so we should either 
1. re implement this validation for portal network
2. or disable log distances validation

I wanted avoid another code duplication, so I chose the latter, but I agree that this function being in `discv5` suggests that distances will always be validated.

Maybe it would be good to move this function to some kinda util module so it can be better re-used between different discovery protocols ?

kdeme:
> I wanted avoid another code duplication, so I chose the latter, but I agree that this function being in `discv5` suggests that distances will always be validated.

OK, I understand the reasoning now. Perhaps it would be more clear then to either have the distances as an `Option`, or to separate it in to two `procs`, one calling the other and only doing the distances check extra.



> Maybe it would be good to move this function to some kinda util module so it can be better re-used between different discovery protocols ?

I'm fine with that.

PR #408, opened 2021-10-18
Enr rlp lists

kdeme:
Noticed that there are ENRs going around that hold RLP lists as value. This is not explicitly forbidden in the ENR spec, however enr decoding fails on that.

Although this is not commonly used, it makes us drop full discv5 `Nodes` messages that still might contain some other ENRs that don't even contain any RLP lists (they get dropped as it is assumed the peer is sending invalid ENRs in general).

### DIFF
--- a/eth/p2p/discoveryv5/enr.nim
+++ b/eth/p2p/discoveryv5/enr.nim
@@ -46,7 +46,8 @@ type
   FieldKind = enum
     kString,
     kNum,
-    kBytes
+    kBytes,
+    kList
 
   Field = object
     case kind: FieldKind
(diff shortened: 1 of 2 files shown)


kdeme:
Such ENR will now print as something like:
```
(9, eth: (Raw RLP list) 0xC7C684B8C6299D80, id: "v4", ip: 0xA7472221, secp256k1: 0x026D13DD4A1CFF9F3AEA234D545254C006812DDBDCA99FF7AE388EB0EF1FFCFF9D, snap: (Raw RLP list) 0xC0, tcp: 30303, udp: 30303)
```

PR #412, opened 2021-10-20
fix big-endian merkleization for UintN arrays

etan-status:
UintN arrays were incorrectly merkleized on big-endian. This was fixed
by making sure to use the correct buffer to store the final chunk.


etan-status:
https://github.com/status-im/nimbus-eth2/pull/3008

PR #414, opened 2021-10-20
allow `ElemType(XXX)` on inputs not called `T`

etan-status:
Currently, `ElemType` can only be called on `List` types when they are
first assigned to a variable called `T`. The template has been adjusted
so that different variable names may be used. The template can now also
be applied to in-line computed types, e.g., `ElemType(typeof(x))`.


etan-status:
https://github.com/status-im/nimbus-eth2/pull/3010

PR #415, opened 2021-10-21
homogenize merkleization code

etan-status:
There were various implementation styles for merkleization of different
data types. This patch homogenizes them for improved maintainability.

### DIFF
--- a/eth/ssz/merkleization.nim
+++ b/eth/ssz/merkleization.nim
@@ -7,7 +7,7 @@
 
 # This module contains the parts necessary to create a merkle hash from the core
 # SSZ types outlined in the spec:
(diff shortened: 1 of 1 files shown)


etan-status:
Assuming that this will also switch over to the new `ssz_serialization` repo. Closing here and re-opening there.

kdeme:
> Assuming that this will also switch over to the new `ssz_serialization` repo. Closing here and re-opening there.

Yes, this code is not to be used anymore and shouldn't get updated. I've not removed it because of this: https://github.com/status-im/nim-ssz-serialization/issues/4
But perhaps I can temporarily place that under fluffy repository and delete all from nim-eth

PR #417, opened 2021-10-27
Utp code cleanup

KonradStaniec:
Mostly kicking code around to have some separate building blocks.

Major things:
- Split `utp_protocol` module, which was responsible for transport, routing packet to correct sockets, and the socket implementation into 3 separate modules `utp_socket`, `utp_router`, and `utp_protocol` (which acutually serves as transport)
- Abstract away transport from `utp_socket`, `utp_router` by using generic address type. That way router and socket can be used in both udp and discoveryv5 context.

### DIFF
--- a/eth/utp/utp_router.nim
+++ b/eth/utp/utp_router.nim
@@ -114,8 +114,8 @@ proc processIncomingBytes*[A](r: UtpRouter[A], bytes: seq[byte], sender: A) {.as
 # Reference implementation: https://github.com/bittorrent/libutp/blob/master/utp_internal.cpp#L2732
 proc connectTo*[A](r: UtpRouter[A], address: A): Future[UtpSocket[A]] {.async.}=
   let socket = initOutgoingSocket[A](address, r.sendCb, r.socketConfig, r.rng[])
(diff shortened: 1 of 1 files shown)


eth/utp/utp_discov5_protocol.nim
```diff
@@ -0,0 +1,75 @@
...
+  var h = 0
+  h = h !& x.remoteAddress.hash
+  h = h !& x.rcvId.hash
+  !$h
+
+proc initSendCallback(t: protocol.Protocol, subProtocolName: seq[byte]): SendCallback[Node] =
+  return (
+    proc (to: Node, data: seq[byte]): Future[void] = 
+      let fut = newFuture[void]()
+      # TODO In discvoveryv5 each talkreq wait for talkresp, but here we would really
```
kdeme, line 22: Yeah, in fact, we don't really care about the result as the response can also not be sent (well not according to discv5 specs right now, but the idea is that in portal utp usage at least).

As long as that is not cleared out, discarding here is best we can do (or add a talkreq which doesn't look/wait for response basically)

PR #419, opened 2021-11-04
Adapt to changes in Chronicles v0.12.0

zah:


tersec:
Chronicles v0.12.0 (and later) have long since been adapted to.

PR #421, opened 2021-11-08
Handling of fin packet

KonradStaniec:
- add handling of fin packet
- add different closing methods

Some arbitrary decisions taken:
- for now sending fin and receiving ack for it result in connection being destroyed by library (which is similar to C reference impl), other choice would be to let user of library known than fin was acked and leave destroying in user hands. This would enable user to have socket open for reading and closed for writing.
Both approaches has their pros and cons, if we would like to have other semantics we could revisit it later if necessary
- for not we have two ways of closing socket - just kill it without any notification to remote or go though fin-ack sequence. Another possibility which would be nice to have, is having method `shutdown` which would first sent reset packet and only then kill socket. (although one could say that it is similar to sending fin and not waiting for ack, but reset is more forceful)

### DIFF
--- a/tests/utp/test_utp_socket.nim
+++ b/tests/utp/test_utp_socket.nim
@@ -7,7 +7,7 @@
 {.used.}
 
 import
-  std/[algorithm, random],
+  std/[algorithm, random, sequtils],
   chronos, bearssl, chronicles,
   testutils/unittests,
   ./test_utils,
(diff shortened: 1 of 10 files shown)


tests/utp/test_utp_socket.nim
```diff
@@ -326,3 +327,171 @@ procSuite "Utp socket unit test":
...
+    await outgoingSocket.processPacket(dataP)
+
+    let bytes = await readF
+
+    check:
+      readF.finished()
+      outgoingSocket.atEof()
+      bytes == concat(data, data1)
+
+  asyncTest "Processing out of order fin should buffer it until receiving all remaining packets":
```
kdeme, line 382: Perhaps rename this test so it is different from the above test.

kdeme (approved):

PR #423, opened 2021-11-10
Advertise correct rcv buffer size

KonradStaniec:
Advertise correct rcv buffer size in packet we send

### DIFF
--- a/tests/utp/test_utp_socket.nim
+++ b/tests/utp/test_utp_socket.nim
@@ -57,8 +57,11 @@ procSuite "Utp socket unit test":
       resultBytes.add(p.payload)
     return resultBytes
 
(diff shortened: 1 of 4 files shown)


tests/utp/test_utp_socket.nim
```diff
@@ -495,3 +499,72 @@ procSuite "Utp socket unit test":
 
     check:
       error.kind == FinSent
+
+  asyncTest "Processing data packet should update window size accordingly and use it in all send packets":
+    let q = newAsyncQueue[Packet]()
+    let initalRemoteSeqNr = 10'u16
```
kdeme, line 505: typo
```suggestion
    let initialRemoteSeqNr = 10'u16
```

kdeme (approved):

PR #424, opened 2021-11-12
Add ackNr validation

KonradStaniec:
Validates that ack number of received packet is in correct range


KonradStaniec:
Merging as pr is pretty small and self contained

PR #425, opened 2021-11-16
Add possibility to connect with requested conneciton id

KonradStaniec:
- Improve error handling when connecting to remote peer.
- Add api to specify connection id of made connections.
- Add possibility to configure a callback which will only accept some of the incoming connecitons.

### DIFF
--- a/eth/utp/utp.nim
+++ b/eth/utp/utp.nim
@@ -7,7 +7,7 @@
 {.push raises: [Defect].}
 
 import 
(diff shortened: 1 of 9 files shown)


KonradStaniec:
After some thought I have also added here possibility to configure accepting only some of the incoming peers. It makes it possible to test our whole flow with discv5.

kdeme (approved):

PR #426, opened 2021-11-17
Fix nim 1.4

markspanbroek:
Fixes compilation and tests for Nim 1.4.

* fixes capitalization of chronicles log level (`ERROR` instead of `error`)
* ~~removes unused `NodeId.toBytes` function~~
   ~~(it wasn't seen by Nim 1.2.x, and leads to failing tests when it is seen by Nim 1.4.x)~~
   Fixed in #427
* removes restriction to Nim 1.2.x in nimble file

By the way: the fact that the `NodeId.toBytes` function was not seen by Nim 1.2.x might mean that there's a bug lurking in the [`logDistance`](https://github.com/status-im/nim-eth/blob/fe1bb4c4e75c9d20698e767427effa5fc29bbbc7/eth/p2p/discoveryv5/routing_table.nim#L107) function. In reality it is calculating the distance based on the platform endianness, but the existence of the NodeId.toBytes function seems to suggest that it should have used big endian instead.

### DIFF
--- a/eth.nimble
+++ b/eth.nimble
@@ -24,13 +24,13 @@ proc runTest(path: string, release: bool = true, chronosStrict = true) =
   let chronosMode =
     if chronosStrict: "-d:chronosStrictException" else: ""
   exec "nim c -r " & releaseMode & " " & chronosMode &
-    " -d:chronicles_log_level=error --verbosity:0 --hints:off " & path
+    " -d:chronicles_log_level=ERROR --verbosity:0 --hints:off " & path
   rmFile path
 
 proc buildBinary(path: string) =
   echo "\nBuilding: ", path
   exec "nim c -d:release -d:chronosStrictException " &
-    "-d:chronicles_log_level=trace --verbosity:0 --hints:off --threads:on " &
+    "-d:chronicles_log_level=TRACE --verbosity:0 --hints:off --threads:on " &
     "--warning[CaseTransition]:off --warning[ObservableStores]:off " &
     path
 
--- a/eth.nimble
+++ b/eth.nimble
@@ -4,7 +4,7 @@ description   = "Ethereum Common library"
 license       = "MIT"
 skipDirs      = @["tests"]
 
-requires "nim >= 1.2.0 & <= 1.2.14",
+requires "nim >= 1.2.0",
          "nimcrypto",
          "stint",
          "secp256k1",


eth.nimble
```diff
@@ -4,7 +4,7 @@ description   = "Ethereum Common library"
 license       = "MIT"
 skipDirs      = @["tests"]
 
-requires "nim >= 1.2.0 & <= 1.2.14",
```
arnetheduck, line 7: before we change this constraint, we need to enable testing on versions we claim to support, as is being done for each library step by step - the plan is to first support multiple versions in CI, then work towards increasing version support in each library - once that is done, we can freely move between nim versions without the risk of one project breaking things for another - see https://discuss.status.im/t/the-road-to-nim-1-6/2305 for a general overview of the migration / upgrade plan

arnetheduck:
> the fact that the NodeId.toBytes function was not seen by Nim 1.2.x might mean that there's a bug lurking in the logDistance function.

this sounds like something  in dire need of a test case (cc @kdeme) and verification that the removal doesn't break something else

eth.nimble
markspanbroek: I agree, I'll see whether I can add the support for multiple Nim versions to the CI script.

arnetheduck: https://github.com/status-im/nim-libp2p/pull/624 should be a useful model - cc @stefantalpalaru is on top of this

kdeme (commented):
> By the way: the fact that the `NodeId.toBytes` function was not seen by Nim 1.2.x might mean that there's a bug lurking in the [`logDistance`](https://github.com/status-im/nim-eth/blob/fe1bb4c4e75c9d20698e767427effa5fc29bbbc7/eth/p2p/discoveryv5/routing_table.nim#L107) function. In reality it is calculating the distance based on the platform endianness, but the existence of the NodeId.toBytes function seems to suggest that it should have used big endian instead.

Oops, I had already hit the issue some time back, but seems I lost track of adding the fix. I've opened a PR which should properly fix this (correct it for BE architecture): https://github.com/status-im/nim-eth/pull/427 , thanks for reporting!

Test cases for this do exist by the way (which I assumed are the ones that failed). It is just not being tested on BE architecture, which is why this remained unnoticed. Some of the test cases were taken from go-ethereum, and I've now also added more of those form there.

markspanbroek:
Great, thanks for the quick fix @kdeme!

markspanbroek:
CI is fixed in #429, which also removes the restriction on Nim version.

PR #428, opened 2021-11-18
Make initial state of socket configurable

KonradStaniec:
Makes initial state of incoming socket configurable. 

By default in reference implementation, incoming socket is created in SynRecv state, and it stays in that state until it receives  data from remote peer or it timeouts. In that state socket cannot send data. This is a way to avoid traffic amplification attacks.

When working over different transport than udp (lets say discoveryv5, as this is our use case)  this measure may be not necessary. Therefore this pr makes it configurable by library user.

### DIFF
--- a/eth/utp/utp_socket.nim
+++ b/eth/utp/utp_socket.nim
@@ -57,6 +57,13 @@ type
     # Maximnal size of receive buffer in bytes
     optRcvBuffer*: uint32
 
+    # If set to some(`duration`), incoming socket will be createded in not connected
+    # state and remote peer will have `duration` to transfer data to make socket
+    # connected
+    # If set to none, incoming socket will be immediatly in connected state and will
+    # be able to transfer data.
+    incomingSocketReceiveTimeout*: Option[Duration]
+
   UtpSocket*[A] = ref object
     remoteAddress*: A
     state: ConnectionState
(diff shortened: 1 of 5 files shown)


eth/utp/utp_socket.nim
```diff
@@ -57,6 +57,13 @@ type
     # Maximnal size of receive buffer in bytes
     optRcvBuffer*: uint32
 
+    # If set to some(`duration`), incoming socket will be createded in not connected
+    # state and remote peer will have `duration` to transfer data to make socket
+    # connected
+    # If set to none, incoming socket will be immediatly in connected state and will
+    # be able to transfer data.
```
kdeme, line 64:
```suggestion
    # If set to some(`Duration`), the incoming socket will be initialized in
    # `SynRecv` state and the remote peer will have `Duration` to transfer data
    # to move the socket in `Connected` state.
    # If set to none, the incoming socket will immediately be set to `Connected`
    # state and will be able to transfer data.
```

tests/utp/test_discv5_protocol.nim
```diff
@@ -81,11 +81,16 @@ procSuite "Utp protocol over discovery v5 tests":
 
     let clientSocketResult = await utp1.connectTo(node2.localNode)
     let clientSocket = clientSocketResult.get()
+    let serverSocket = await queue.get()
 
     check:
       clientSocket.isConnected()
+      # in this test we do not configure socket to be connected just affter
```
kdeme, line 88:
```suggestion
      # in this test we do not configure the socket to be connected just after
```

kdeme (approved):
some nitpicks/typos

eth/utp/utp_socket.nim
```diff
@@ -441,7 +443,8 @@ proc initOutgoingSocket*[A](
     sndConnectionId,
     initialSeqNr,
     # Initialy ack nr is 0, as we do not know remote inital seqnr
-    0
+    0,
+    cfg.initialSynTimeout
   )
 
 proc initIncomingSocket*[A](
```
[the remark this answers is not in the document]
KonradStaniec: yea, you are probably right. The same applies to the outgoing one 👍

PR #431, opened 2021-11-22
Make queryRandom async and add exports

kdeme:


eth/p2p/discoveryv5/protocol.nim
```diff
@@ -716,9 +718,9 @@ proc query*(d: Protocol, target: NodeId, k = BUCKET_SIZE): Future[seq[Node]]
   d.lastLookup = now(chronos.Moment)
   return queryBuffer
 
-proc queryRandom*(d: Protocol): Future[seq[Node]] =
+proc queryRandom*(d: Protocol): Future[seq[Node]] {.async.} =
   ## Perform a query for a random target, return all nodes discovered.
-  d.query(NodeId.random(d.rng[]))
+  return await d.query(NodeId.random(d.rng[]))
```
arnetheduck, line 723: this await is a no-op essentially, except for slowing down the processing by an event loop

kdeme: hmm, right. I'll revert this, I don't actually need it, I was sleeping.

PR #433, opened 2021-11-23
avoid allocation in `hash(ValidIpAddress)`

arnetheduck:
While casually browsing the profiler output, to my great surprise I
found that an allocating string conversion function (gasp!) in the hash
function for ip addresses - this PR carefully excises this evil
construct from the codebase.

### DIFF
--- a/eth.nimble
+++ b/eth.nimble
@@ -4,7 +4,7 @@ description   = "Ethereum Common library"
 license       = "MIT"
 skipDirs      = @["tests"]
 
-requires "nim >= 1.2.0 & <= 1.2.12",
+requires "nim >= 1.2.0 & <= 1.2.14",
          "nimcrypto",
          "stint",
          "secp256k1",
--- a/eth/net/utils.nim
+++ b/eth/net/utils.nim
@@ -16,7 +16,10 @@ type
     limit*: uint
     ips: Table[ValidIpAddress, uint]
 
-func hash(ip: ValidIpAddress): Hash = hash($ip)
+func hash(ip: ValidIpAddress): Hash =
+  case ip.family
+  of IpAddressFamily.IPv6: hash(ip.address_v6)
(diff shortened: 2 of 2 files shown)


zah:
It's a bit odd that this is defined here. I think it belongs to the module of the `ValidIpAddress` type in stew.

arnetheduck:
presumably whoever added it here disagreed, so I considered it outside of the scope of this PR - ie I guess requires importing/exporting the `hashes` module which has its own downsides and issues and easily could balloon into a bigger change.

In general, `ValidIpAddress` is quite odd as a construct given that .. er .. all possible values for IP addresses are valid, in some sense - it's unclear what the "validity" refers to, except perhaps as a testbed for `requiresInit` - what's funniest is that the parser it uses "validity" itself rejects correctly formatted ip addresses :)

kdeme (approved):
I'm fine with keeping it in nim-eth.
And the original idea of using `ValidIpAddress` here for me was indeed the `requiresInit` part so that one could avoid having a Node without an IP initialized from the ENR to end up in the routing table (e.g. an 0.0.0.0). Not sure if that really holds up though ... as it will accept also such IP. And because of that this type of check is done elsewhere anyhow.

And yes, oops! Looks like leftover from first quick dev that was forgotten to clean-up before PR.

eth.nimble
```diff
@@ -4,7 +4,7 @@ description   = "Ethereum Common library"
 license       = "MIT"
 skipDirs      = @["tests"]
 
-requires "nim >= 1.2.0 & <= 1.2.12",
+requires "nim >= 1.2.0 & <= 1.2.14",
```
kdeme, line 7: This shouldn't be needed as this was already adjusted some weeks ago. 

It's strange that Github doesn't notice any more differences while the branch is not based from latest..

arnetheduck: diff vs `git merge-base` I'm guessing

PR #435, opened 2021-11-25
Add optional node validation to routing table

jm-clius:
This PR addresses https://github.com/status-im/nim-waku/issues/770

It adds an optional node validator to the discv5 routing table, useful for client-defined rules to prevent certain nodes from being added for discovery v5. This can be used i.a. to create a separate discv5 network for Waku v2 based on the [Waku v2 ENR key](https://rfc.vac.dev/spec/31/#waku2-enr-key).

@kdeme, assuming that we reach consensus on this PR's approach, I'm not sure what the suggested way would be to proceed:
a) merge this to nim-eth `master` if it's "safe enough", or
b) rather run Waku v2 discovery v2 on a forked implementation for a while until we see how well it works.

cc @richard-ramos for future reference

### DIFF
--- a/eth/p2p/discoveryv5/protocol.nim
+++ b/eth/p2p/discoveryv5/protocol.nim
@@ -116,6 +116,8 @@ const
   ## call
 
 type
+  NodeValidator* = proc(node: Node): bool {.gcsafe, raises: [Defect].}
+
   Protocol* = ref object
     transp: DatagramTransport
     localNode*: Node
@@ -135,6 +137,7 @@ type
     talkProtocols*: Table[seq[byte], TalkProtocol] # TODO: Table is a bit of
     # overkill here, use sequence
     rng*: ref BrHmacDrbgContext
+    nodeValidator*: Option[NodeValidator]
 
   PendingRequest = object
     node: Node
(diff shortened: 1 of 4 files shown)


oskarth (commented):
Seems reasonable to me.

Question: Are we "validating" a node or are we "filtering"? E.g. what's the proper model to use here? For validating, it seems like we'd kick off those nodes completely. For filtering, it seems like we'd mostly hide from view, and one could imagine this filter growing into something more dynamic (say capability based). But perhaps that should happen at a higher level too, not sure.

kdeme (commented):
> @kdeme, assuming that we reach consensus on this PR's approach, I'm not sure what the suggested way would be to proceed:
> a) merge this to nim-eth `master` if it's "safe enough", or
> b) rather run Waku v2 discovery v2 on a forked implementation for a while until we see how well it works.

I do think it is safe enough, and for that I'd be fine with having it merged.

I'm however not fully sure if it is sufficient, see my answer to your question here: https://github.com/status-im/nim-waku/issues/770#issuecomment-979792158

Perhaps it should first be tried out because of this? We could leave this branch alive and it can be used in nim-waku. WDYT?

jm-clius:
Thanks, @kdeme and also for your through answer in https://github.com/status-im/nim-waku/issues/770

> Perhaps it should first be tried out because of this? We could leave this branch alive and it can be used in nim-waku. WDYT?

I agree, and think it's relatively cheap to test on this feature branch for a while.

> Question: Are we "validating" a node or are we "filtering"?

@oskarth fair point. In my mind, preventing a node from being added to the local routing table is slightly closer to "validating" than "filtering". Maybe this is only because discv5 already allows for ["filtering" on enr field](https://github.com/status-im/nim-eth/blob/master/eth/p2p/discoveryv5/protocol.nim#L725) when querying for nodes, and I wanted to keep this a separate concept. I'd say capability filtering is closer to the latter. No strong opinion though.

jm-clius:
Closing, as we've gone a different way. cc @kaiserd

PR #437, opened 2021-11-30
Track send buffer and properly handle back pressoure when window is to small to process data

KonradStaniec:
Pr add several improvements necessary to properly handle changing size of sending window. 

Main design requirement here is that writing data larger than current send window should asynchronously block the caller while the write is progressing. That way caller can just call `write` proc and the whole flow control will be under control of the library.

To that end it is necessary to introduce writeLoop to properly track all caller writes, and progress them one by one

Current window send window is sized only by the remote peer window.

### DIFF
--- a/eth/utp/utp_socket.nim
+++ b/eth/utp/utp_socket.nim
@@ -66,6 +66,23 @@ type
     # state and will be able to transfer data.
     incomingSocketReceiveTimeout*: Option[Duration]
 
+  WriteErrorType* = enum
+    SocketNotWriteable, 
+    FinSent
+
+  WriteError* = object
+    case kind*: WriteErrorType
+    of SocketNotWriteable:
+      currentState*: ConnectionState
+    of FinSent:
+      discard
+
+  WriteResult* = Result[int, WriteError]
+
+  WriteRequest = object
+    data: seq[byte]
+    writer: Future[WriteResult]
+
   UtpSocket*[A] = ref object
     remoteAddress*: A
     state: ConnectionState
@@ -148,6 +165,10 @@ type
 
     sendBufferTracker: SendBufferTracker
 
+    writeQueue: AsyncQueue[WriteRequest]
+
+    writeLoop: Future[void]
+
     # socket identifier
     socketKey*: UtpSocketKey[A]
 
@@ -158,19 +179,6 @@ type
   SocketCloseCallback* = proc (): void {.gcsafe, raises: [Defect].}
 
   ConnectionError* = object of CatchableError
-
-  WriteErrorType* = enum
-    SocketNotWriteable, 
-    FinSent
-
-  WriteError* = object
-    case kind*: WriteErrorType
-    of SocketNotWriteable:
-      currentState*: ConnectionState
-    of FinSent:
-      discard
-
-  WriteResult* = Result[int, WriteError]
 
   OutgoingConnectionErrorType* = enum
     SocketAlreadyExists, ConnectionTimedOut, ErrorWhileSendingSyn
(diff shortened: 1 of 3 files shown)


eth/utp/utp_socket.nim
```diff
@@ -399,6 +437,93 @@ proc checkTimeoutsLoop(s: UtpSocket) {.async.} =
...
+          doAssert(socket.state == Destroy)
+          if (not writeFut.finished()):
+             let res = Result[int, WriteError].err(WriteError(kind: SocketNotWriteable, currentState: socket.state))
+             writeFut.complete(res)
+          # we need to re-raise exception so the outer loop will be properly cancelled too
+          raise exc
+        bytesWritten = bytesWritten + len(dataSlice)
+        i = lastOrEnd + 1
+
+      # in case future got cancelled
```
kdeme, line 488: Don't you mean: "In case data got send"?

eth/utp/utp_socket.nim
```diff
@@ -305,29 +331,34 @@ proc flushPackets(socket: UtpSocket) {.async.} =
...
   while i < s.curWindowPackets:
 
     let packetSeqNr = s.seqNr - 1 - i
-    if (s.outBuffer.exists(packetSeqNr, (p: OutgoingPacket) => p. transmissions > 0 and p.needResend == false)):
+    if (s.outBuffer.exists(packetSeqNr, (p: OutgoingPacket) => p.transmissions > 0 and p.needResend == false)):
       s.outBuffer[packetSeqNr].needResend = true
       let packetPayloadLength = s.outBuffer[packetSeqNr].payloadLength
-      doAssert(s.currentWindow >= packetPayloadLength)
-      s.currentWindow = s.currentWindow - packetPayloadLength
+      # lack of waiters notification in case of timeout effectivly means that
```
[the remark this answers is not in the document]
KonradStaniec: nope, it is the case that if we have some write in progress which is waiting for free bytes in buffer, we do not want it to progress in case of this decrease, as then user of library would be able to grow the send buffer infinitely i.e writes would always progress  when there would be timeouts of some packets.

eth/utp/utp_socket.nim
KonradStaniec: no really, as this is the future returned to the user of the library, and can only by completed in our write loop (unless user will do something weird and complete this future himself), but user of the library could at this point also cancel this. 

I will change it to ` in case future got cancelled or finished`

eth/utp/utp_socket.nim
```diff
@@ -399,6 +437,93 @@ proc checkTimeoutsLoop(s: UtpSocket) {.async.} =
 proc startTimeoutLoop(s: UtpSocket) =
   s.checkTimeoutsLoop = checkTimeoutsLoop(s)
 
+proc getPacketSize*(socket: UtpSocket): int =
+  # TODO currently returning constant, ultimatly it should be bases on mtu estimates
```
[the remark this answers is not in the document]
KonradStaniec: yea probably, I kinda imagine that packet size can be in principle estimated in different ways in Discv5 and in UDP.

eth/utp/utp_socket.nim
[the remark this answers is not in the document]
kdeme: Yeah, that's what I meant with "end up still getting send", as in still being processed after the free

eth/utp/utp_socket.nim
kdeme: Yes, but my point is, you are completing the future (successfully) because all data got send, right?

KonradStaniec: oh yep! the comment only appliaes to the if part (` if (not writeFut.finished()):`)

KonradStaniec: I have reformulated it a little bit

PR #444, opened 2021-12-09
Update timestamp and ackNr on packet resend

KonradStaniec:
Currently timestamp and ackNr of packet were filled only during first send of a packet. This mean that in case of packet re-sends (in case of failue) those values could be stale and could lead for example to incorrect delays calculation

### DIFF
--- a/eth/utp/packets.nim
+++ b/eth/utp/packets.nim
@@ -9,7 +9,7 @@
 import
   std/[monotimes],
   faststreams,
-  stew/[endians2, results, objects], bearssl,
+  stew/[endians2, results, objects, arrayops], bearssl,
   ../p2p/discoveryv5/random2
 
 export results
@@ -56,9 +56,11 @@ type
 # https://github.com/bittorrent/libutp/blob/master/utp_utils.cpp, to check all the
 # timing assumptions on different platforms
 proc getMonoTimeTimeStamp*(): uint32 = 
+  # TODO
   # this value is equivalent of:
   # uint32((Moment.now() - Moment.init(0, Microseconds)).microseconds())
   # on macOs
+  # so we we can use moment here and return (Moment, uint32) tuple from this function
   let time = getMonoTime()
   cast[uint32](time.ticks() div 1000)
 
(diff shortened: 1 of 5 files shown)


eth/utp/utp_socket.nim
```diff
@@ -357,9 +357,16 @@ proc sendAck(socket: UtpSocket): Future[void] =
 
 # Should be called before sending packet
 proc setSend(s: UtpSocket, p: var OutgoingPacket): seq[byte] =
+  let currentMoment = Moment.now()
```
kdeme, line 360: We should be able to reduce these two calls to one.

I remember in the earlier PRs we discussed this but I forgot the reason why `getMonoTimeTimeStamp` was still needed. Or was it just because of the conversion?

Anyway, perhaps a TODO to be added.

eth/utp/packets.nim
```diff
@@ -136,6 +136,23 @@ proc decodePacket*(bytes: openArray[byte]): Result[Packet, string] =
...
 
+proc modifyTimeStampAndAckNr*(packetBytes: var seq[byte], newTimestamp: uint32, newAckNr: uint16) =
+  ## Modifies timestamp and ack nr of already encoded packets. Those fiels should be
+  ## filled right before sending, so in case of re-sent packet we would like to update
+  ## them withound decoding and re-encoding packet once again
+
+  doAssert(len(packetBytes) >= minimalHeaderSize)
+  let timestampBytes = toBytesBE(newTimestamp)
+  let ackBytes = toBytesBE(newAckNr)
+  var timeStampStart = 4
```
kdeme, line 147: Perhaps better call this timeStampLoc or Index.

Side note: I'm wondering if we don't have some nicer utility calls available to do this (aside from directly doing memCopy), perhaps: https://github.com/status-im/nim-stew/blob/master/stew/arrayops.nim#L71 ? Not sure if there are any caveats with that, and it probably doesn't really matter considering that the fields are small.

eth/utp/packets.nim
```diff
@@ -136,6 +136,23 @@ proc decodePacket*(bytes: openArray[byte]): Result[Packet, string] =
 
     ok(Packet(header: header, payload: payload))
 
+proc modifyTimeStampAndAckNr*(packetBytes: var seq[byte], newTimestamp: uint32, newAckNr: uint16) =
+  ## Modifies timestamp and ack nr of already encoded packets. Those fiels should be
+  ## filled right before sending, so in case of re-sent packet we would like to update
+  ## them withound decoding and re-encoding packet once again
```
kdeme, line 142:
```suggestion
  ## Modifies timestamp and ack nr of already encoded packets. Those fields should be
  ## filled right before sending, so when re-sending the packet we would like to update
  ## it without decoding and re-encoding the packet once again
```

kdeme (approved):

eth/utp/utp_socket.nim
KonradStaniec: I totally agree. The problem is that there is no easy way to get numeric value of time stamp from `Moment` object.
But I recently thought that we can get around this by using: `(Moment.now() - ZeroMoment).microseceons())` as subtraction of two moments give as `Duration` which has procs to retrieve numeric values 
I will leave todo here and fix it in next pr.

eth/utp/packets.nim
KonradStaniec: thanks for `arrayops` it looks much better with it ! 👍

PR #445, opened 2021-12-09
Utp top level raises & some clean-up

kdeme:
Had to adjust some unneeded `CatchableError`s and make the `messageHandler` private. Did some additional clean-up while skimming over the code.


KonradStaniec (approved):
Thanks for fixing all of my typos 😅

PR #446, opened 2021-12-10
Improve handling of timestamps

KonradStaniec:
Refactor timestamps so now `Moment` and numeric timestamps are always consistent with each other

### DIFF
--- a/eth/utp/packets.nim
+++ b/eth/utp/packets.nim
@@ -61,6 +61,15 @@ type
 # every system, and warnings when monotonic clock is not avaialable.
 proc getMonoTimestamp*(): TimeStampInfo = 
   let currentMoment = Moment.now()
+
+  # Casting this value from int64 to uin32, my lead to some sudden spikes in
+  # timestamp numeric values i.e it is possible that timestamp can suddenly change
+  # from 4294967296 to for example 10, this may lead to sudden spikes in
+  # calculated delays
+  # uTP implementation is resistant to those spikes are as it keeps history of
+  # few last delays on uses smallest one for calculating ledbat window.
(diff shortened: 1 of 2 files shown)


eth/utp/packets.nim
```diff
@@ -47,22 +50,19 @@ type
...
-  # TODO
-  # this value is equivalent of:
-  # uint32((Moment.now() - Moment.init(0, Microseconds)).microseconds())
-  # on macOs
-  # so we we can use moment here and return (Moment, uint32) tuple from this function
-  let time = getMonoTime()
-  cast[uint32](time.ticks() div 1000)
+proc getMonoTimestamp*(): TimeStampInfo = 
+  let currentMoment = Moment.now()
+  let timestamp = uint32((currentMoment - zeroMoment).microseconds())
```
kdeme, line 64: This relies on wrapping around, right? I guess this mostly isn't an issue the left over value is still fairly long in this context. But couldn't it give a little trouble/edge case when its at the moment of before and after the max?

kdeme (approved):

eth/utp/utp_socket.nim
```diff
@@ -939,14 +934,14 @@ proc processPacket*(socket: UtpSocket, p: Packet) {.async.} =
     if (sentTimeRemote == 0):
       0'u32
     else:
-      receiptTimestamp - sentTimeRemote
+      timestampInfo.timestamp - sentTimeRemote
```
kdeme, line 937: I guess this is one of those occasions where my first comment can have an impact. But it already needs to be resistant against it anyway as the remote clock can be off also as you mention in the comment.

PR #447, opened 2021-12-10
remove outdated and incorrect SSZ code

etan-status:
This removes the outdated copy of the SSZ code. It became incorrect
over time (e.g., empty SSZ list elements), and is no longer in use by
GitHub projects: https://github.com/search?q=extension%3Anim+eth%2Fssz

The canonical SSZ implementation resides at `nim-ssz-serialization`.
Compared to `nim-eth`, these changes were made meanwhile:
- `bitseqs` was extended with JSON serialization support,
  and with the new functions `isZero` and `countOnes`.
- `bytes_reader` was renamed to `codec`, extended with a few additional
  SSZ type conversions as well as support for `SingleMemberUnion`.
- The simplified merkle tree implementation in `merkle_tree.nim`
  was removed. It was not used by other projects.
- `merkleization` was extended with support for `HashArray`, `HashList`
  and `SingleMemberUnion`. The `isValidProof` functionality has been
  moved to `nimbus-eth2` and replaced with the EF defined function
  `is_valid_merkle_branch`. The test was also moved to `nimbus-eth2`.
  There are no other GitHub projects using `isValidProof`:
  https://github.com/search?q=extension%3Anim+isValidProof
  Furthermore, a definition for `GeneralizedIndex` was added.
- `ssz_serialization` was moved one directory up, and improved with
  bug fixes and `HashArray`, `HashList` and `SingleMemberUnion` support.
- `types` was extended with JSON serialization and new type support for
  `Uint128`, `Uint256`, `HashArray`, `HashList` and `SingleMemberUnion`.
  There is also a new `getBit` function for `BitList`.


mratsim (approved):
LGTM

kdeme:
This code was only here still, after the move of most code to nim-ssz-serialization, because of the older/simpler merkle proof building and the `isValidProof` code and its tests, as I didn't have the time to check if anything needed to be salvaged of that code since the new development in nimbus-eth2. I see the latter got covered now completely after https://github.com/status-im/nimbus-eth2/pull/3182. The first seems to be covered also by the `build_proof` code in nimbus_eth2 (cc @KonradStaniec might want to have a check also).


> * It was not used by other projects.

True, but not **yet** used. It will become eventually needed for the Fluffy client in nimbus-eth1. Hence, we need to move the parts that are in nimbus-eth2 in the (near) future.

etan-status:
Yep, I'll move the proof validation over once the other pending SSZ work is done.

etan-status:
Pending SSZ work that touches the same file: https://github.com/status-im/nim-ssz-serialization/pull/16
After that one is in, I'll port the proof validation over (`is_valid_merkle_branch` EF api).

PR #450, opened 2021-12-11
reject WHOAREYOU packets with non-empty message

etan-status:
This changes the `discv5` parser to reject malformed WHOAREYOU packets
that have a non-0 message length. The extra data used to be ignored.
The `message` part of WHOAREYOU packets is always empty.
See https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md

### DIFF
--- a/eth/p2p/discoveryv5/encoding.nim
+++ b/eth/p2p/discoveryv5/encoding.nim
@@ -441,12 +441,16 @@ proc decodeMessagePacket(c: var Codec, fromAddr: Address, nonce: AESGCMNonce,
     messageOpt: some(message), requestNonce: nonce, srcId: srcId))
 
 proc decodeWhoareyouPacket(c: var Codec, nonce: AESGCMNonce,
-    iv, header: openArray[byte]): DecodeResult[Packet] =
+    iv, header, ct: openArray[byte]): DecodeResult[Packet] =
   # TODO improve this
   let authdata = header[staticHeaderSize..header.high()]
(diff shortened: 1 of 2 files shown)


eth/p2p/discoveryv5/encoding.nim
```diff
@@ -441,13 +441,17 @@ proc decodeMessagePacket(c: var Codec, fromAddr: Address, nonce: AESGCMNonce,
...
-    iv, header: openArray[byte]): DecodeResult[Packet] =
+    iv, header, ct: openArray[byte]): DecodeResult[Packet] =
   # TODO improve this
   let authdata = header[staticHeaderSize..header.high()]
   # We now know the exact size that the authdata should be
   if authdata.len != idNonceSize + sizeof(uint64):
     return err("Invalid header length for whoareyou packet")
 
+  # The `message` part of WHOAREYOU packets is always empty.
+  if ct.len != 0:
```
kdeme, line 452: It's a little bit weird to pass in an usually empty array and check for its size, but I understand its done like this to remain consistent with having the size checks inside these decode packet procs.

kdeme (approved):

eth/p2p/discoveryv5/encoding.nim
etan-status: Yeah, it is just a consistency thing, so that the responsibility of checking message length is always in the handler. It's not an inner loop so should not affect performance.

PR #455, opened 2021-12-27
Minor improvements from pr comments

KonradStaniec:


arnetheduck:
In general, a more helpful commit message would list the actual improvements (without the context of the PR, it's hard to tell when looking at git history)

PR #461, opened 2022-01-14
Add simple utp discovery test app

KonradStaniec:
Aimple utp over discv5 test app. It mirrors utp interface with json-rpc, so it is easy to check all common utp operations.

Draft for now as it is not yet settled if we will use it

### DIFF
--- a/eth/utp/discv5_test_app/discv5_test_app.nim
+++ b/eth/utp/discv5_test_app/discv5_test_app.nim
@@ -1,0 +1,201 @@
+# Copyright (c) 2021 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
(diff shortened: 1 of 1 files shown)


kdeme (commented):
If we were to add this to the CI to build (which we should to avoid bit rot), then it would mean we would have to add the nim-json-rpc dependency to the nim-eth Nimble file, which is then probably annoying for other projects / modules.

This is actually the main reason that I've put the discv5 json-rpc API used by Fluffy in the nimbus-eth1 repo and not the nim-eth repo. The API could be useful for other projects too, but I didn't want to force it upon users.

So perhaps this tool should also go into nimbus-eth1, more specifically the fluffly/tools folder? Might also be easier then to build into docker images for testing cross client?

KonradStaniec:
you are probably right.

I was a little unsure where this should live, as it is not necessarily fluffy specific thing, but also not really library thing. After I wil have something workable,I will close this pr and open it in fluffy repo.

KonradStaniec:
closing in favour: https://github.com/status-im/nimbus-eth1/pull/931

PR #462, opened 2022-01-19
Re-send acks for old duplicated packets

KonradStaniec:
Fix one bug when testing with 10% packet loss. In case where
1. `A` sends `DATA` packet
2. `B` receives it and send `STATE` packet
3. `STATE` packet is lost
4. `A` re-sends `DATA` packet
we did not re-sent `STATE` for this duplicated  `DATA` packet

### DIFF
--- a/eth/utp/utp_socket.nim
+++ b/eth/utp/utp_socket.nim
@@ -1019,6 +1019,8 @@ proc processPacket*(socket: UtpSocket, p: Packet) {.async.} =
 
   debug "Process packet",
     socketKey = socket.socketKey,
+    socketAckNr= socket.ackNr,
+    socketSeqNr= socket.seqNr,
     packetType = p.header.pType,
     seqNr = p.header.seqNr,
     ackNr = p.header.ackNr,
@@ -1059,6 +1061,19 @@ proc processPacket*(socket: UtpSocket, p: Packet) {.async.} =
 
   # If packet is totally of the mark short circout the processing
   if pastExpected >= reorderBufferMaxSize:
+
+    # if `pastExpected` is really big number (for example: uint16.high) then most
+    # probably we are receiving packet which we already received
+    # example: we already received packet with `seqNr = 10` so our `socket.ackNr = 10`
+    # if we receive this packet once again then `pastExpected = 10 - 10 - 1` which 
+    # equals (due to wrapping) 65535
+    # this means that remote most probably did not receive our ack, so we need to resend
+    # it. We are doing it for last `reorderBufferMaxSize` packets
+    let isPossilbeDuplicatedOldPacket = pastExpected >= (int(uint16.high) + 1) - reorderBufferMaxSize
+
+    if (isPossilbeDuplicatedOldPacket and p.header.pType != ST_STATE):
+      asyncSpawn socket.sendAck()
+
     debug "Got an invalid packet sequence number, too far off",
       pastExpected = pastExpected
     return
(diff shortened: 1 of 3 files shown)


eth/utp/utp_socket.nim
```diff
@@ -1019,6 +1019,8 @@ proc processPacket*(socket: UtpSocket, p: Packet) {.async.} =
 
   debug "Process packet",
     socketKey = socket.socketKey,
+    socketAckNr= socket.ackNr,
+    socketSeqNr= socket.seqNr,
```
kdeme, line 1023:
```suggestion
    socketAckNr = socket.ackNr,
    socketSeqNr = socket.seqNr,
```

eth/utp/utp_socket.nim
```diff
@@ -1059,6 +1061,19 @@ proc processPacket*(socket: UtpSocket, p: Packet) {.async.} =
 
   # If packet is totally of the mark short circout the processing
   if pastExpected >= reorderBufferMaxSize:
+
+    # if `pastExpected` is really big number (for example: uint16.high) then most
```
kdeme, line 1065: Wouldn't it be ok to just check before/at line 1050 if pkSeqNr <= socket.ackNr or something alike?

KonradStaniec: hmm I am pretty sure it would lead to some issues when wrapping around i.e something like
1. `socket.ackNr = 0`
2. We receive packet with `pkSeqNr = 65535` i.e the one before. 
3. with `pkSeqNr <= socket.ackNr`  we do not send ack even though we should

From what I saw in reference impl it is better to stick with this relative values like `pastExpected` and `acks` as they correctly account for this scenarios.

eth/utp/utp_socket.nim
```diff
@@ -1059,6 +1061,19 @@ proc processPacket*(socket: UtpSocket, p: Packet) {.async.} =
...
+    # probably we are receiving packet which we already received
+    # example: we already received packet with `seqNr = 10` so our `socket.ackNr = 10`
+    # if we receive this packet once again then `pastExpected = 10 - 10 - 1` which 
+    # equals (due to wrapping) 65535
+    # this means that remote most probably did not receive our ack, so we need to resend
+    # it. We are doing it for last `reorderBufferMaxSize` packets
+    let isPossilbeDuplicatedOldPacket = pastExpected >= (int(uint16.high) + 1) - reorderBufferMaxSize
+
+    if (isPossilbeDuplicatedOldPacket and p.header.pType != ST_STATE):
```
kdeme, line 1074:
```suggestion
    let isPossibleDuplicatedOldPacket = pastExpected >= (int(uint16.high) + 1) - reorderBufferMaxSize

    if (isPossibleDuplicatedOldPacket and p.header.pType != ST_STATE):
```

eth/utp/utp_socket.nim
kdeme: Ok, I hadn't considered the scenario where we run out of pkSeqNr . However, wouldn't that scenario have to same (invalid) effect on the `isPossibleDuplicatedOldPacket` bool?

KonradStaniec: not really as in this scenario:
`pastExpected = 65535 - 0 - 1 => pastExpected = 65534`
and `65534 >= 65536 - 1024` so `isPossibleDuplicatedOldPacket` evaluates to `true` therefore we will send ack as necessary in that case.

kdeme: Well yes, perhaps I'm not understanding your example well, but in my original suggestion of `pkSeqNr <= socket.ackNr`, it would also be seen as a possible duplcated old packet and thus acked again.

PR #474, opened 2022-02-08
Add more tests stressing conccurent reading and writing on utp socket

KonradStaniec:


eth/utp/utp_socket.nim
```diff
@@ -287,7 +288,7 @@ const
   allowedAckWindow*: uint16 = 3
 
   # Timeout after which the send window will be reset to its minimal value after it dropped
-  # to zero. i.e when we received a packet from remote peer with `wndSize` set to 0.
+  # belwo our current packet size. i.e when we received a packet from remote peer with `wndSize` set to number <= current packet size
```
kdeme, line 291:
```suggestion
  # a value lower than our current packet size. i.e when we received a packet from remote peer with `wndSize` set to number <= current packet size
```

tests/utp/test_protocol_integration.nim
```diff
@@ -0,0 +1,231 @@
+# Copyright (c) 2020-2021 Status Research & Development GmbH
```
kdeme, line 1:
```suggestion
# Copyright (c) 2022 Status Research & Development GmbH
```

eth/utp/utp_socket.nim
```diff
@@ -566,57 +572,59 @@ proc resetSendTimeout(socket: UtpSocket) =
...
+          dataSlice,
+          socket.replayMicro
+        )
+      let outgoingPacket = OutgoingPacket.init(encodePacket(dataPacket), 1, false, payloadLength)
+      socket.registerOutgoingPacket(outgoingPacket)
+      await socket.sendData(outgoingPacket.packetBytes)
+    except CancelledError as exc:
+      # write loop has been cancelled in the middle of processing due to the
+      # socket closing
+      # this approach can create partial write in case destroyin socket in the
```
kdeme, line 613:
```suggestion
      # this approach can create partial write in when destroying the socket in the
```

eth/utp/utp_socket.nim
```diff
@@ -566,57 +572,59 @@ proc resetSendTimeout(socket: UtpSocket) =
...
       if (not writeFut.finished()):
-        writeFut.complete(Result[int, WriteError].ok(bytesWritten))
+          let res = Result[int, WriteError].err(WriteError(kind: SocketNotWriteable, currentState: socket.state))
+          writeFut.complete(res)
+      # we need to re-raise exception so the outer loop will be properly cancelled too
+      raise exc
+    bytesWritten = bytesWritten + len(dataSlice)
+    i = lastOrEnd + 1
+
+  # Before completeing future with success (as all data was sent sucessfuly)
```
kdeme, line 624:
```suggestion
  # Before completing the future with success (as all data was sent successfully)
```

eth/utp/utp_protocol.nim
```diff
@@ -78,6 +80,7 @@ proc new*(
   address: TransportAddress,
   socketConfig: SocketConfig = SocketConfig.init(),
   allowConnectionCb: AllowConnectionCallback[TransportAddress] = nil,
+  sendCallbackBuilder: SendCallbackBuilder = nil,
```
kdeme, line 83: Why is this `sendCallbackBuilder` needed? Is it just to have access to `maxDelay: int, packetDropRate: int` in the test cases? Would that not work also when build outside of the utp code?

kdeme (approved):

eth/utp/utp_protocol.nim
KonradStaniec: yep it is here so user of the library can augment send callback. Currently it is used only in tests to pass `maxDelay: int, packetDropRate: int` so that we can simulate delays and dropping packets but in theory one could add some application specific logic like logging or metrics for each send.

I think there could be the way to generalize it a bit and make it possible to build it outside utp and pass it as param, but this approach was a little bit simpler to implement.

PR #475, opened 2022-02-16
Add event loop to socket

KonradStaniec:
Proposition to add event loop to socket.

Main thing this improves is that before `processPacket` was public async procedure, which means that in theory there could be many `processPacket` procedures running concurrently racing between each other. Not to mention timeout `checkTimeout` was also async procedure running concurrently.

With addition of eventLoop there can be always at most one `processPacket` procedure running, also `processPacket` does not race with `checkTimeout` anymore. 

This approach could be easily extended to other types of events.

Full list of changes:
- add eventLoop to control all incoming events
- change semantic of write to asynchronously block only when send buffer is full, and not when bytes do not fit into send window
- change handling of receive buffer, to start dropping packets if the reorder buffer and receive buffer are full. Old behaviour was to asyncblock unless there is space which could lead to resource exhaustion attacks


kdeme (approved):
I didn't review into all the details, but the mechanism of the eventloop does sound like a better / controlled way of dealing with the incoming data and other events

Issue #255, opened 2020-06-18
RLP list decoding too lax for objects


kdeme:
There seems to be no list size check done when decoding an object (possible also in general for lists?).

e.g. when decoding `ef8c000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000000080` to a `WhoAreYouObj`, that works as expected, as the rlp is valid.

However, decoding `c08c000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000000080` to a `WhoAreYouObj`, also works, while clearly the list size (c0 -> size 0) is not correct.

Only spotted this by adding a re-encode step and comparing with original payload in a fuzzing test.

Perhaps related with https://github.com/status-im/nim-eth/issues/254

mjfh:
Just ran into that issue when debugging wire protocol datagrams as processed by `checkedRlpRead()`. In particular, `r.read(MsgType)` succeeds when `r` is a non-empty list and `MsgType` is `EmptyList` and also when `r` is a list of more than one element and  `MsgType` is `DisconnectionReasonList`, a single element object. The latter one coincides with the reported issue, above.

PR #476, opened 2022-02-21
Jordan/sync test node tuning

mjfh:


eth/p2p.nim
```diff
@@ -81,8 +81,8 @@ proc listeningAddress*(node: EthereumNode): ENode =
   node.toENode()
 
 proc startListening*(node: EthereumNode) {.raises: [CatchableError, Defect].} =
-  # TODO allow binding to specific IP / IPv6 / etc
-  let ta = initTAddress(IPv4_any(), node.address.tcpPort)
+  # Bind to un- or specific IP / IPv6 / etc
+  let ta = initTAddress(node.address.ip, node.address.tcpPort)
```
kdeme, line 85: `node.address` is the external (reachable) address that should be used in the enode string.
The bind address is not necessarily the same, and while I'm not sure if this is consistently used correctly in the code base, it should be a seperate configurable parameter which gets defaulted to `IPv4_any()` (or the IPv6 version when that gets supported).

eth/p2p/discovery.nim
```diff
@@ -281,8 +281,8 @@ proc processClient(transp: DatagramTransport, raddr: TransportAddress):
     debug "Receive failed", exc = e.name, err = e.msg
 
 proc open*(d: DiscoveryProtocol) {.raises: [Defect, CatchableError].} =
-  # TODO allow binding to specific IP / IPv6 / etc
-  let ta = initTAddress(IPv4_any(), d.address.udpPort)
+  # Bind to un- or specific IP / IPv6 / etc
+  let ta = initTAddress(d.address.ip, d.address.udpPort)
```
kdeme, line 285: Same as above remark

eth/p2p/rlpx.nim
```diff
@@ -207,7 +207,8 @@ proc getDispatcher(node: EthereumNode,
 
 proc getMsgName*(peer: Peer, msgId: int): string =
   if not peer.dispatcher.isNil and
-     msgId < peer.dispatcher.messages.len:
+     msgId < peer.dispatcher.messages.len and
+     not peer.dispatcher.messages[msgId].isNil:
```
kdeme, line 211: When does this end up being `nil`?

eth/p2p/rlpx.nim
```diff
@@ -506,20 +507,53 @@ proc recvMsg*(peer: Peer): Future[tuple[msgId: int, msgData: Rlp]] {.async.} =
     await peer.disconnectAndRaise(BreachOfProtocol,
                                   "Cannot read RLPx message id")
 
+template checkedRlpReadFringeCases(r: var Rlp, MsgType: type): untyped =
+  when MsgType is EmptyList:
+    # No list at all
+    if not r.hasData:
```
kdeme, line 513: So some client was sending no data in the case where it should be an empty list, e.g. the devp2p ping message?
Do the other clients accept this?

eth/p2p/rlpx.nim
```diff
@@ -506,20 +507,53 @@ proc recvMsg*(peer: Peer): Future[tuple[msgId: int, msgData: Rlp]] {.async.} =
...
+      return EmptyList()
+    # Non empty list would be accepted as `EmptyList` by rlp parser
```
kdeme, line 515: Not sure, but perhaps it can be made stricter by making EmptyList and empty array or such.

The whole `EmptyList` and `DisconnectionReasonList` was kind of a workaround in the first place though.

eth/p2p.nim
mjfh: Hmm, currently it is initialised in Nimbus as the listener address by the corresponding options. Its default is `IPv4_any()` but I am not sure if this is the whole story, though.

eth/p2p/rlpx.nim
```diff
@@ -506,20 +507,53 @@ proc recvMsg*(peer: Peer): Future[tuple[msgId: int, msgData: Rlp]] {.async.} =
...
+    if r.isList and 0 < r.listlen:
+      raise newException(RlpTypeMismatch, "Empty list expected")
+  when MsgType is DisconnectionReasonList:
+    # Also accepted: a single byte int, or a blob of an int list
```
kdeme, line 519: What are other clients sending here. What are we accepting?

In theory it should just be a list of a single item (which is a bit weird yes)?

eth/p2p/rlpx.nim
```diff
@@ -506,20 +507,53 @@ proc recvMsg*(peer: Peer): Future[tuple[msgId: int, msgData: Rlp]] {.async.} =
...
-            # rlpData = r.inspect
-
+            # r.inspect would crash here
+      raise e
+    except rlp.RlpError as e:
+      debug "Failed rlp.read",
+            peer = peer,
+            dataType = MsgType.name,
+            exception = e.msg,
+            rlpData = r.inspect
```
kdeme, line 556: I wouldn't add this in a debug build even, it gets pretty verbose and slow.

Perhaps at trace level, but even then.

eth/p2p/rlpx.nim
```diff
@@ -506,20 +507,53 @@ proc recvMsg*(peer: Peer): Future[tuple[msgId: int, msgData: Rlp]] {.async.} =
...
       return r.read(MsgType)
-    except rlp.RlpError as e:
+    except rlp.MalformedRlpError as e:
       debug "Failed rlp.read",
             peer = peer,
             dataType = MsgType.name,
             exception = e.msg
-            # rlpData = r.inspect
-
+            # r.inspect would crash here
```
kdeme, line 549: Did you identify on what part?

tests/p2p/test_rlpx_thunk.json
```diff
@@ -1,6 +1,6 @@
 {
-  "Invalid list when decoding for object": {
```
kdeme, line 2: Both tests are useful, don't replace the original one.

eth/p2p.nim
mjfh: Would function prototypes do that indicate setup functionality as `openIpv4Any()` and `openStrict()` or so (I know this is ugly)?

tests/p2p/test_rlpx_thunk.json
```diff
@@ -49,10 +49,15 @@
     "error": "MalformedRlpError",
     "description": "listElem to error on invalid size encoding"
   },
-  "Listing elements when not a list": {
```
kdeme, line 52: Same remark as above

eth/p2p/rlpx.nim
```diff
@@ -506,20 +507,53 @@ proc recvMsg*(peer: Peer): Future[tuple[msgId: int, msgData: Rlp]] {.async.} =
     await peer.disconnectAndRaise(BreachOfProtocol,
                                   "Cannot read RLPx message id")
 
+template checkedRlpReadFringeCases(r: var Rlp, MsgType: type): untyped =
```
kdeme, line 510: In general I'm not to found of having this call, and would hope things could get fixed by using the right types and perhaps some fixes in rlp. But if that is not directly feasible we can go with this route temporary. 
However, I'd like to understand a bit better which client is sending this different message encodings, and if it shouldn't be fixed on their end (and us just dropping these). And thus basically also knowing what other clients do with it (say geth).

eth/p2p.nim
mjfh: But until then we should at least hide the `listenAddress`  option in Nimbus as it is not used. I find it annoying having a blind feature :)

eth/p2p/rlpx.nim
mjfh: I had instances when the seq[] entry was NULL which caused a SEGFAULT (see PR comment.) This happened for a while and then vanished, again.

eth/p2p/rlpx.nim
mjfh: I know, still handy for debugging as it uncovers a lot of information. I could prob. log the peer if I enter this case so I could _grep_ the ident from the log file-- might make sense.

eth/p2p/rlpx.nim
mjfh: I noticed that, see https://github.com/status-im/nim-eth/issues/255#issuecomment-1047096223

eth/p2p/rlpx.nim
kdeme: Well, right now it is not just informative, but actually alters the received data from what I understand? So it is more than just "handy too debug"?

So I'm trying to understand in this PR:
- What rlp serialisation cases we are "too strict" on and should get passed although we error on them.
- What rlp serialization cases we are "too lax" on and should get the messages dropped.

And this ideally per specification, unless all/most clients deviate from this in which case the implementation becomes spec (Yes, annoying, I know).

This PR seems to do both? But it is not immediately clear to me based on what grounds what is altered here, without myself digging into it.

eth/p2p/rlpx.nim
mjfh: makes sense though :)

tests/p2p/test_rlpx_thunk.json
mjfh: They would break the rlp parser changes as it relaxes strictness. In particular the disconnect reason list is returnd by geth 1.10.6 and 1.2.0 as a blobbed list on devnet4 (but should probably confirmed at mainnet.)

eth/p2p/rlpx.nim
mjfh: neither am i, but at least it is recorded somewhere -- it is part of the debugging PR cited in the patch list above

eth/p2p/rlpx.nim
kdeme: Ok, I specifically ask because i think there are other locations in the code that call messages[id]

PR #484, opened 2022-03-10
Jordan/nethermind sync

mjfh:
Support for syncing against _Nethermind_ on _Devnet4_, _Devnet5_, _Kiln_. Here is a list of client ids syncing against `kiln` up to block _#100217_:
```
 1 besu/v22.1.3
 1 erigon/v2022.99.99
60 Geth/v1.10.17
96 Nethermind/v1.12.4 chunked
 4 Nethermind/v1.12.4
```
So up to now, the _chunked_ protocol version of _Nethermind_ still seems to play a role.

As can be shown in a test set up, _Geth_  rejects the chunked messages when detecting truncated RLP data. Many  messages are smaller than 1k (typical chunk size) so that this behaviour is observed only for larger ones.

### DIFF
--- a/eth/p2p/rlpx.nim
+++ b/eth/p2p/rlpx.nim
@@ -13,6 +13,17 @@ import
   ".."/[rlp, common, keys, async_utils],
   ./private/p2p_types, "."/[kademlia, auth, rlpxcrypt, enode, p2p_protocol_dsl]
 
+const
+  # Insane kludge for suppporting chunked messages when syncing against clients
+  # like Nethermind.
+  #
+  # The original specs which are now obsoleted can be found here:
+  # github.com/ethereum/devp2p/commit/6504d410bc4b8dda2b43941e1cb48c804b90cf22.
+  #
+  # The current requirement is stated at
+  # github.com/ethereum/devp2p/blob/master/rlpx.md#framing
+  allowObsoletedChunkedMessages = defined(chunked_rlpx_enabled)
+
 when useSnappy:
   import snappy
   const devp2pSnappyVersion* = 5
@@ -455,8 +466,8 @@ proc resolveResponseFuture(peer: Peer, msgId: int, msg: pointer, reqId: int) =
 
 proc getRlpxHeaderData(header: RlpxHeader): (int,int,int) =
   ## Helper for `recvMsg()`
-  # This is insane. Some clients like Nethermid use the now obsoleted
-  # chunked frame protocol, see
+  # This is insane. Some clients like Nethermind use the now obsoleted
+  # chunked message frame protocol, see
   # github.com/ethereum/devp2p/commit/6504d410bc4b8dda2b43941e1cb48c804b90cf22.
   result = (-1, -1, 0)
   proc datagramSize: int =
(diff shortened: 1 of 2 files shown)


mratsim:
How does Geth sync against nethermind btw? From last week discussion, at first glance they seemed to support incompatible mode of syncing?

mjfh:
> How does Geth sync against nethermind btw? From last week discussion, at first glance they seemed to support incompatible mode of syncing?

Unfortunately, Nethermind is the only node type that works with Nimbus-eth1 on the test networks. It returns (obsoleted) chunked RLPx data when returning data blocks. Geth currently answers with a sub-protocol error right after sending a status message.

eth/p2p/rlpx.nim
```diff
@@ -452,15 +464,53 @@ proc resolveResponseFuture(peer: Peer, msgId: int, msg: pointer, reqId: int) =
 
     debug "late or duplicate reply for a RLPx request"
 
+proc getRlpxHeaderData(header: RlpxHeader): (int,int,int) =
+  ## Helper for `recvMsg()`
+  # This is insane. Some clients like Nethermind use the now obsoleted
+  # chunked message frame protocol, see
+  # github.com/ethereum/devp2p/commit/6504d410bc4b8dda2b43941e1cb48c804b90cf22.
+  result = (-1, -1, 0)
```
kdeme, line 472: Avoid using result in new code: https://status-im.github.io/nim-style-guide/language.result.html

eth/p2p/rlpx.nim
```diff
@@ -496,16 +546,104 @@ proc recvMsg*(peer: Peer): Future[tuple[msgId: int, msgData: Rlp]] {.async.} =
...
+
   var rlp = rlpFromBytes(decryptedBytes)
 
+  var msgId: int32
   try:
     # int32 as this seems more than big enough for the amount of msgIds
-    let msgId = rlp.read(int32)
-    return (msgId.int, rlp)
+    msgId = rlp.read(int32)
+    result = (msgId.int, rlp)
```
kdeme, line 571: Same about result usage as above.

eth/p2p/rlpxcrypt.nim
```diff
@@ -200,6 +200,14 @@ proc decryptHeader*(c: var SecretState, data: openArray[byte],
     c.aesdec.decrypt(toa(data, 0, RlpHeaderLength), output)
     result = ok()
 
+proc decryptHeaderAndGetMsgSize*(c: var SecretState,
+                                 encryptedHeader: openArray[byte],
+                                 outSize: var int,
+                                 outHeader: var RlpxHeader): RlpxResult[void] =
+  result = decryptHeader(c, encryptedHeader, outHeader)
```
kdeme, line 207: Same about result usage as above.

eth/p2p/rlpxcrypt.nim
```diff
@@ -200,6 +200,14 @@ proc decryptHeader*(c: var SecretState, data: openArray[byte],
     c.aesdec.decrypt(toa(data, 0, RlpHeaderLength), output)
     result = ok()
 
+proc decryptHeaderAndGetMsgSize*(c: var SecretState,
+                                 encryptedHeader: openArray[byte],
+                                 outSize: var int,
+                                 outHeader: var RlpxHeader): RlpxResult[void] =
```
kdeme, line 206: It's a bit strange having a version of this call that both outputs the outSize and the header, as it seems that then later that header is again parsed for its size + other fields

PR #491, opened 2022-03-18
Try to debut chronos defect

KonradStaniec:


KonradStaniec:
fixed in: https://github.com/status-im/nim-eth/pull/493

PR #492, opened 2022-03-23
Change closing proc

KonradStaniec:
So as seen from CI run this pr does not solve Defect which is sometimes thrown from tests. But it fixes weird issue, that when both produced and consumer loops are canceled from consumer loops context only producer is cancelled and consumer is left hanging in weird state.

Most simple minimizer:
```
import 
  std/[deques, sequtils],
  chronos

type Foo = ref object
  q: AsyncQueue[int]
  pLoop: Future[void]
  cLoop: Future[void]

proc producerLoop(f: Foo): Future[void] {.async.} =
  try:
    var i = 0
    while true:
      await sleepAsync(milliseconds(500))
      echo "puting i = " & $i
      f.q.putNoWait(i)
      inc i
  except CancelledError as exc:
    echo "Producer Cancelled"
    raise exc


proc consumerLoop(f: Foo): Future[void] {.async.} =
  try:
    while true:
      let i = await f.q.get()
      echo "Hello i is " &  $i
      if i == 2:
        f.cLoop.cancel()
        f.pLoop.cancel()
  except CancelledError as exc:
    echo "Consumer Cancelled"
    raise exc


proc new(T: type Foo): T =
  let f = Foo(q: newAsyncQueue[int]())
  f.pLoop = producerLoop(f)
  f.cLoop = consumerLoop(f)
  return f

let s = Foo.new()

runForever()

```

Now I would expect to see in logs:
```
Consumer Cancelled
Producer Cancelled
```
but in only 
```
Producer Cancelled
```
shows up

tbh I am not sure if its intended behaviour for such case. Wdyt @kdeme ?

### DIFF
--- a/eth/utp/utp_socket.nim
+++ b/eth/utp/utp_socket.nim
@@ -612,9 +612,13 @@ proc checkTimeoutsLoop(s: UtpSocket) {.async.} =
   try:
     while true:
       await sleepAsync(checkTimeoutsLoopInterval)
-      await s.eventQueue.put(SocketEvent(kind: CheckTimeouts))
-  except CancelledError:
+      s.eventQueue.putNoWait(SocketEvent(kind: CheckTimeouts))
+  except CancelledError as exc:
+    # check timeouts loop is last running future managed by socket, if its
+    # cancelled we can fire closeEvent
+    s.closeEvent.fire()
     trace "checkTimeoutsLoop canceled"
+    raise exc
 
 proc startTimeoutLoop(s: UtpSocket) =
   s.checkTimeoutsLoop = checkTimeoutsLoop(s)
@@ -682,6 +686,10 @@ proc isConnected*(socket: UtpSocket): bool =
 proc isClosed*(socket: UtpSocket): bool =
   socket.state == Destroy and socket.closeEvent.isSet()
 
+proc isClosedAndCleanedUpAllResources*(socket: UtpSocket): bool =
+  ## Test Api to check that all resources are properly cleaned up 
+  socket.isClosed() and socket.eventLoop.cancelled() and socket.checkTimeoutsLoop.cancelled()
+
 proc destroy*(s: UtpSocket) =
   info "Destroying socket",
     to = s.socketKey
@@ -689,8 +697,12 @@ proc destroy*(s: UtpSocket) =
   ## Remote is not notified in any way about socket end of life
   s.state = Destroy
   s.eventLoop.cancel()
-  s.checkTimeoutsLoop.cancel()
-  s.closeEvent.fire()
+  # This procedure initiate cleanup process which goes like:
+  # Cancel EventLoop -> Cancel timeoutsLoop -> Fire closeEvent
+  # This is necessary due to how evenLoop look like i.e it has only one await
+  # point on `eventQueue.get` which trigger cancellation excepion only when
+  # someone will try run `eventQueue.put`. Without `eventQueue.put` , eventLoop
+  # future shows as cancelled, but handler for CancelledError is not run
 
 proc destroyWait*(s: UtpSocket) {.async.} =
   ## Moves socket to destroy state and clean all reasources and wait for all registered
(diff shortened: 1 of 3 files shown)


kdeme (commented):
I've tested your minimized test case and noticed that both loops do get canceled, just the message is not printed, so it doesn't get into the except clause (However, how come it does get canceled then? Confusing.).

I'd have to dig in the code / generated code to try to figure this out, but I did notice the following already:
- Adding an `await sleepAsync` in the in the consumerLoop() makes "Consumer Cancelled" also print.
- Using `cancelAndWait` instead of cancel seems cleaner and works. You'd have to change the order of course:
```
  await f.pLoop.cancelAndWait()
  await f.cLoop.cancelAndWait()
```

Canceling it outside of either loop also works, I don't know if it is/was intended to be supported the way you are using it here.

eth/utp/utp_socket.nim
```diff
@@ -682,15 +686,23 @@ proc isConnected*(socket: UtpSocket): bool =
...
+  socket.isClosed() and socket.eventLoop.cancelled() and socket.checkTimeoutsLoop.cancelled()
+
 proc destroy*(s: UtpSocket) =
   info "Destroying socket",
     to = s.socketKey
   ## Moves socket to destroy state and clean all reasources.
   ## Remote is not notified in any way about socket end of life
   s.state = Destroy
   s.eventLoop.cancel()
-  s.checkTimeoutsLoop.cancel()
```
kdeme, line 692: I'm not fully understanding this versus your minimized example: It seems that in this case both loops are canceled from "the outside", but I assume `checkTimeoutsLoop` gives issues, similar as in the minimized example?

KonradStaniec: from what I gathered it does not, so the  chain of closing looks like:
- we are canceling consumer loop (event loop) from the inside, and it works as producer (check timeouts) loop still produces items and putting item onto cancelled queue triggers proper `CancelledException` when we are awaiting for `get`
- form consumer loop cancel handler we trigger cancellation of producer loop, it also properly tirgger `CancelledException`

It works and every handler is properly triggered, although I agree it is not super pretty.

eth/utp/utp_socket.nim
```diff
@@ -612,9 +612,13 @@ proc checkTimeoutsLoop(s: UtpSocket) {.async.} =
...
       await sleepAsync(checkTimeoutsLoopInterval)
-      await s.eventQueue.put(SocketEvent(kind: CheckTimeouts))
-  except CancelledError:
+      s.eventQueue.putNoWait(SocketEvent(kind: CheckTimeouts))
+  except CancelledError as exc:
+    # check timeouts loop is last running future managed by socket, if its
+    # cancelled we can fire closeEvent
+    s.closeEvent.fire()
     trace "checkTimeoutsLoop canceled"
+    raise exc
```
kdeme, line 621: Are the re-raises here and in EventLoop still needed?

KonradStaniec: nope, will remove them

KonradStaniec: actually after second thought, I would keep those re-raises  as with them both futures `eventLoop` and `checkTimeoutsLoop` are properly returning that they are cancelled (i.e. `future.cancelled` returns `true`) without them those future could indicate that they are just finished.

KonradStaniec:
I digged  a bit why it happens, and it mostly have to do with how chronos and and `await` work together. So in the example from minimisation:
```
proc consumerLoop(f: Foo): Future[void] {.async.} =
  try:
    while true:
      let i = await f.q.get()
      echo "Hello i is " &  $i
      if i == 2:
        f.cLoop.cancel()
        f.pLoop.cancel()
  except CancelledError as exc:
    echo "Consumer Cancelled"
    raise exc
```

when we cancel consumer loop from within, it does not have at this point any child futures so it directly goes to `cancelAndSchedule` ([cancelAndSchedule](https://github.com/status-im/nim-chronos/blob/master/chronos/asyncfutures2.nim#L257))  proc, so it does set status of future as `Cancelled` and `mustCancel` flag to `true`, but it does not throw any exception so the loop goes to next iteration and it will wait for another element from queue. Now `await f.q.get()` will only throw exception when it will finish and it will finish only when some one will call `put`, and this is due to `await` checking parent future status only after future finishing [checkParentStatus](https://github.com/status-im/nim-chronos/blob/master/chronos/asyncmacro2.nim#L273)

So the approach from this pr works due to the fact, that `checkTimeoutsLoop` is cancelled after `eventLoop` so there will be new events which will trigger exception in eventLoop `await`

PR #494, opened 2022-03-24
Mitigating RLP annoyances

mjfh:
why:
  Rlp errors throw exceptions which cause the dispatcher loop to
  terminate the current session immediately.

details:
  The DisconnectionReasonList message requires a single entry list.
  Observed and now accepted deviations are:

  Geth: single byte number

  bor(a Geth fork): blobbed single entry list containing a number

### DIFF
--- a/eth/p2p/rlpx.nim
+++ b/eth/p2p/rlpx.nim
@@ -50,7 +50,7 @@ proc read(rlp: var Rlp; T: type DisconnectionReasonList): T
     # Be strict here: The expression `rlp.read(DisconnectionReasonList)`
     # accepts lists with at least one item. The array expression wants
     # exactly one item.
-    let a = rlp.read(array[1,DisconnectionReason])
+    let a = rlp.read(array[1, DisconnectionReason])
     return DisconnectionReasonList(value: a[0])
 
   # Also accepted: a single byte reason code. Is is typically used
--- a/eth/p2p/rlpx.nim
+++ b/eth/p2p/rlpx.nim
@@ -54,7 +54,7 @@ proc read(rlp: var Rlp; T: type DisconnectionReasonList): T
     return DisconnectionReasonList(value: a[0])
 
   # Also accepted: a single byte reason code. Is is typically used
-  # by variants of the reference implentation `Geth`
+  # by variants of the reference implementation `Geth`
   if rlp.blobLen <= 1:
     let n = rlp.read(int)
     return DisconnectionReasonList(value: n.DisconnectionReason)
--- a/eth/p2p/rlpx.nim
+++ b/eth/p2p/rlpx.nim
@@ -43,8 +43,8 @@ type
   DisconnectionReasonList = object
     value: DisconnectionReason
 
-proc read(rlp: var Rlp; Q: type DisconnectionReasonList): Q
-    {.raises: [Defect,RlpError].} =
+proc read(rlp: var Rlp; T: type DisconnectionReasonList): T
+    {.raises: [Defect, RlpError].} =
   ## Rlp mixin: `DisconnectionReasonList` parser
   if rlp.isList:
     # Be strict here: The expression `rlp.read(DisconnectionReasonList)`
--- a/eth/p2p/rlpx.nim
+++ b/eth/p2p/rlpx.nim
@@ -59,7 +59,7 @@ proc read(rlp: var Rlp; T: type DisconnectionReasonList): T
     let n = rlp.read(int)
     return DisconnectionReasonList(value: n.DisconnectionReason)
 
-  # Also accepted: a blob of a list (aka object) of reason code. Is is
+  # Also accepted: a blob of a list (aka object) of reason code. It is
   # used by `bor`, a `geth` fork
   var subList = rlp.toBytes.rlpFromBytes
   if subList.isList:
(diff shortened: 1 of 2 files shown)


mratsim:
How big of a change would it be to replace exceptions with results for RLP?

eth/p2p/rlpx.nim
```diff
@@ -43,6 +43,32 @@ type
   DisconnectionReasonList = object
     value: DisconnectionReason
 
+proc read(rlp: var Rlp; Q: type DisconnectionReasonList): Q
+    {.raises: [Defect,RlpError].} =
```
kdeme, line 47:
```suggestion
proc read(rlp: var Rlp; T: type DisconnectionReasonList): T
    {.raises: [Defect, RlpError].} =
```

eth/p2p/rlpx.nim
```diff
@@ -43,6 +43,32 @@ type
...
+  ## Rlp mixin: `DisconnectionReasonList` parser
+  if rlp.isList:
+    # Be strict here: The expression `rlp.read(DisconnectionReasonList)`
+    # accepts lists with at least one item. The array expression wants
+    # exactly one item.
+    let a = rlp.read(array[1,DisconnectionReason])
```
kdeme, line 53:
```suggestion
    let a = rlp.read(array[1, DisconnectionReason])
```

eth/p2p/rlpx.nim
```diff
@@ -43,6 +43,32 @@ type
...
+    return DisconnectionReasonList(value: a[0])
+
+  # Also accepted: a single byte reason code. Is is typically used
+  # by variants of the reference implentation `Geth`
```
kdeme, line 57:
```suggestion
  # by variants of the reference implementation `Geth`
```

eth/p2p/rlpx.nim
```diff
@@ -43,6 +43,32 @@ type
...
+  if rlp.blobLen <= 1:
+    let n = rlp.read(int)
+    return DisconnectionReasonList(value: n.DisconnectionReason)
+
+  # Also accepted: a blob of a list (aka object) of reason code. Is is
```
kdeme, line 62:
```suggestion
  # Also accepted: a blob of a list (aka object) of reason code. It is
```

eth/p2p/rlpx.nim
```diff
@@ -43,6 +43,32 @@ type
...
+  # used by `bor`, a `geth` fork
+  var subList = rlp.toBytes.rlpFromBytes
+  if subList.isList:
+    # Diddo, see above.
```
kdeme, line 66:
```suggestion
    # Ditto, see above.
```

tests/p2p/test_rlpx_thunk.json
```diff
@@ -49,10 +49,15 @@
     "error": "MalformedRlpError",
     "description": "listElem to error on invalid size encoding"
   },
-  "Listing elements when not a list": {
```
kdeme, line 52: I wouldn't remove this test, it is still useful

kdeme (commented):
lgtm, just some style / typo nitpicks.

Not sure if we should really care about this `bor` behaviour, based on client stats: https://www.ethernodes.org/
Unless some other clients also do this (it is rather a weird serialisation...)

kdeme:
> How big of a change would it be to replace exceptions with results for RLP?

I wouldn't go there for now. I am all for usage of Results, but I believe there are more important items to work on first.
And when we want to start addressing this we should first aim to contain the exceptions within rlp, and for example getting first rid of them bit by bit in the networking code.

Be aware that discv5 also uses rlp, so it would be something that we have to rework very carefully. And discv5 is pretty much exception free except for rlp.
Side note: SSZ in eth2 also relies heavily on exceptions (or nim-serialization in general)

tests/p2p/test_rlpx_thunk.json
mjfh: It would fail, commenting out like /*..*/ in json works but are non-standard.

mjfh:
On the other hand, this is a real annoyance. The later we wait, the more we work around exceptions. For practical reasons, the assumption that data RLP are correct does not hold. So exceptions become the rule (rathner than the exception.)

kdeme:
> On the other hand, RLP exceptions are a real annoyance. The later we wait, the more we work around exceptions. For practical reasons, the assumption that RLP data are correct does not hold. So exceptions become the rule (rather than the exception.)

I understand that exceptions give even extra issues (performance) when they become the "common path". 
What I'm saying is to approach this from the top to the bottom. If you would rework rlp first to use for example results instead of exceptions, a lot of code on top of it will have to be reworked, which means a big endeavor/PR.
You could of course make the rlp code raise at the public API and use results internally only, but then you kind of loose the benefit of why you are removing the exceptions imo.

FYI, I've canceled some CI builds as each little commit was doing a CI build. I've just kept the last one running.

tests/p2p/test_rlpx_thunk.json
kdeme: hmm, right, because you accept a blob now.

kdeme:
@mjfh When having a commit history like the one above (Small commits with not so meaningful commit messages, and changes that belong to the same functionality/feature), please squash them first. This keeps our history cleaner and easier to assess for changes.

Perhaps merge commit should be disabled on nim-eth repo.

PR #495, opened 2022-03-30
disc: updateExternalIp()

stefantalpalaru:
New public proc that can be used to inform the discovery subsystem about
a changed external IP (as reported by UPnP/NAT-PMP in some other module).

### DIFF
--- a/eth/p2p/discoveryv5/protocol.nim
+++ b/eth/p2p/discoveryv5/protocol.nim
@@ -842,7 +842,8 @@ proc refreshLoop(d: Protocol) {.async.} =
   except CancelledError:
     trace "refreshLoop canceled"
 
-proc updateExternalIp*(d: Protocol, extIp: ValidIpAddress, udpPort: Port) =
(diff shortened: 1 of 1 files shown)


eth/p2p/discoveryv5/protocol.nim
```diff
@@ -842,6 +842,20 @@ proc refreshLoop(d: Protocol) {.async.} =
...
+  let
+    previous = d.localNode.address
+    res = d.localNode.update(d.privateKey,
+      ip = some(extIp), udpPort = some(udpPort))
+
+  if res.isErr:
+    warn "Failed updating ENR with newly discovered external address",
+      previous, newExtIp = extIp, newUdpPort = udpPort, error = res.error
+  else:
+    discovery_enr_auto_update.inc()
```
kdeme, line 855: As it becomes a public function, this metric does not hold any more necessarily.

I assume you want to use this for uPnP based IP updates, which is also some sort of `enr_auto_update`, but I think that should be a different metric?

So perhaps this `proc` should either use a result, returning the error message, or a bool, and then based on the return valuethe metric can be incremented or not?

PR #496, opened 2022-03-30
nat: getPublicRoutePrefSrcOrExternalIP()

stefantalpalaru:
And some ~~public gateway~~ *preferred routing source* refactoring to minimise code duplication.

### DIFF
--- a/eth/net/nat.nim
+++ b/eth/net/nat.nim
@@ -23,6 +23,13 @@ type
     NatPmp
     NatNone
 
+  DefaultGatewayStatus = enum
+    NoRoutingInfo
+    DefGwIsPublic
+    DefGwIsPrivate
+    BindAddressIsPublic
+    BindAddressIsPrivate
+
 const
   UPNP_TIMEOUT = 200 # ms
   PORT_MAPPING_INTERVAL = 20 * 60 # seconds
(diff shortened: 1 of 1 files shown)


stefantalpalaru:
OK, so it's about the preferred source for a route to a public IP, not the default gateway. Naming fixed.

eth/net/nat.nim
```diff
@@ -102,6 +109,42 @@ proc getExternalIP*(natStrategy: NatStrategy, quiet = false): Option[IpAddress]
           error "parseIpAddress() exception", err = e.msg
           return
 
+# This queries the routing table to get the "preferred source" attribute and
+# checks if it's a public IP. If so, then it's our public IP.
+#
+# Further more, we check if the bind address (user provided, or a "0.0.0.0"
+# default) is a public IP. That's a long shot, because code paths involving a
+# user-provided bind address are not supposed to get here.
```
kdeme, line 117: Not sure I get the comment here. In nimbus-eth2 the bind-address (listen address) does get passed up until here?

kdeme (approved):

eth/net/nat.nim
stefantalpalaru: Not if `--listen-address=1.2.3.4` is combined with `--nat=extip:1.2.3.4`, which a knowledgeable user would do.

kdeme: Okay, that's what you mean. Yes, you're right that in this (more sensible) scenario it is not useful, but it can be still useful for those not knowing/understanding the nat:extip option. After all they are not behind a NAT in that case, so why should they care (this is more an issue of bad UX for this cli option but that aside).

PR #507, opened 2022-05-25
Dedicated RLP reader for `DisconnectionReason`

mjfh:
Currently, a program communicating via RLPX will crash when receiving out of bound reason codes disconnect message.

This happens in the RLP message parser. Out of bound value assignments to an enum causes a `RangeError`defect and consequently the program to terminate. This `RangeError` is avoided here and a catchable error raised.

### DIFF
--- a/eth/p2p/rlpx.nim
+++ b/eth/p2p/rlpx.nim
@@ -61,7 +61,7 @@ proc read[T: DisconnectionReason](rlp: var Rlp; W: type array[1,T]): T
   # fixed-size array (aka blob), see function `eth/rlp.readImpl()`.
   let rc = rlp.read(array[1,int])[0].to(T)
   if rc.isErr:
-    raise newException(MalformedRlpError, rc.error)
+    raise newException(RlpTypeMismatch, rc.error)
   rc.value
 
 proc read(rlp: var Rlp; T: type DisconnectionReasonList): T
--- a/eth/p2p/rlpx.nim
+++ b/eth/p2p/rlpx.nim
@@ -54,16 +54,6 @@ type
   DisconnectionReasonList = object
     value: DisconnectionReason
 
-proc read[T: DisconnectionReason](rlp: var Rlp; W: type array[1,T]): T
-    {.gcsafe, raises: [RlpError, Defect].} =
-  ## Rlp mixin: `array[1,DisconnectionReason]` parser
-  # Need to parse for an `int` type as a `byte` (or a `char`) would expect a
-  # fixed-size array (aka blob), see function `eth/rlp.readImpl()`.
-  let rc = rlp.read(array[1,int])[0].to(T)
-  if rc.isErr:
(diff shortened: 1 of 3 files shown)


eth/p2p/rlpx.nim
```diff
@@ -54,29 +54,41 @@ type
...
+    {.gcsafe, raises: [RlpError, Defect].} =
   ## Rlp mixin: `DisconnectionReasonList` parser
+
   if rlp.isList:
     # Be strict here: The expression `rlp.read(DisconnectionReasonList)`
     # accepts lists with at least one item. The array expression wants
     # exactly one item.
-    let a = rlp.read(array[1, DisconnectionReason])
-    return DisconnectionReasonList(value: a[0])
+    return DisconnectionReasonList(
```
kdeme, line 75: I think this is perhaps being made a bit more complex than it should be, several items:
* I gave a quick try at `return DisconnectionReasonList(value: rlp.read(array[1, DisconnectionReason])[0])` and that does work for me. What exactly doesn't work with that for you with that array usage? As else all other custom rlp parsing code can be dropped I think.
* Could also go with something like `rlp.enterList` and then `return DisconnectionReasonList(value: rlp.read(DisconnectionReason))`. This will not check if there is more than 1 item in the rlp list. However, from what I remember, this sometimes may not be checked (e.g. for the hello message this is specifically mentioned in the spec as a must), and I think that's why perhaps it also doesn't get checked for our rlp object encoding, but I might be wrong on this. (Anyhow, this would practically be similar as just running a read on `DisconnectionReasonList`).
* If one does want a check on the above, be aware that the array read implementation uses listLen for this, which iterates over all the items before throwing an exception if there are too many. This is probably worse from a security point of view than just ignoring the rest. For that reason, if you really do want to restrict it to single item list, its probably just better to do a `rlp.blobLen == 2` check in this already custom rlp code.
* Less important, but perhaps a `DisconnectionReasonListOrValue` can be created, which then has a custom read that then reads `DisconnectionReasonList` or `DisconnectionReason`. (Change List with Array if we go for that solution).

mjfh: > * I gave a quick try at `return DisconnectionReasonList(value: rlp.read(array[1, DisconnectionReason])[0])` and that does work for me. What exactly doesn't work with that for you with that array usage? As else all other custom rlp parsing code can be dropped I think.

will double check (again), maybe I made another mistake

mjfh: > > * I gave a quick try at `return DisconnectionReasonList(value: rlp.read(array[1, DisconnectionReason])[0])` and that does work for me. What exactly doesn't work with that for you with that array usage? As else all other custom rlp parsing code can be dropped I think.
> 
> will double check (again), maybe I made another mistake

That was it: 
 ```
 DisconnectionReasonList(value: rlp.read(array[1,DisconnectionReason])[0])
```

PR #511, opened 2022-06-15
Increase tests speed to avoid hangs

KonradStaniec:
Decreases drop rates and delays to avoid hangs in CI.
Also avoid using `yield`

### DIFF
--- a/tests/utp/test_protocol_integration.nim
+++ b/tests/utp/test_protocol_integration.nim
@@ -16,7 +16,6 @@ import
   ../../eth/keys,
   ../../eth/p2p/discoveryv5/random2
 
-
 proc connectTillSuccess(p: UtpProtocol, to: TransportAddress, maxTries: int = 20): Future[UtpSocket[TransportAddress]] {.async.} =
   var i = 0
   while true:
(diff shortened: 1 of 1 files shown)


kdeme (commented):
Why does it end up hanging with higher drop rates and delays?

KonradStaniec:
My working theory is that it is mainly due to troubles with establishing utp connection when there are high drop rates i.e

- when Syn is lost it leads to 3,6s,12s delay before each resend 
- if Syn-Ack is lost, then every re-sent Syn will be ignored by receiver, so it immediately forces scenario with all re-sends on sender side.

So establishing connection in those conditions can take a lot of time, and after that there is still whole test scenario to through.  Also some scenarios have really small buffers which reduces throughput significantly which increases test times significantly.

kdeme:
But did you have a look at the timestamps provided in the issue?
There is a huge difference between a passing test and a failing one (which gets canceled due to 1h timeout). This would indicate an actual hang and not just a test that takes very long.

KonradStaniec:
Yep I saw it, it is about 40min and it is plausible that it could happen with larger drop rates and longer delays and the facts about utp connections establishment I mentioned earlier.

1 failed connection attempt is about 21s of timeouts only, so to simplify lets say whole failed connection take 30s. So 10 failed connections attempts would take 5min (we know it does not take more than 20 as this is the limit in `connectTillSuccess` loop). And those are only connections attempts for one scenario, without the running scenario itself. And each scenario also can take some time due to packet loss and delays. So taking into account that each test have 3-4 scenarios, imo this is plausible that sometimes we will have much longer run.

Of course due to probabilistic nature of this tests, this is only a theory. What worries me a bit more is that it happens only on CI, and I was not able to reproduce it locally despite many tries. So there is definitly a possibility of different explanation  lie bug in uTP which causes tests to hang.

kdeme:
> What worries me a bit more is that it happens only on CI, and I was not able to reproduce it locally despite many tries.

This and the fact that there is such a big timestamp difference makes me think it is more likely that the test really is stuck.

Failed connection attempts would happen on each test run, sure some might have few more than others.

KonradStaniec:
> This and the fact that there is such a big timestamp difference makes me think it is more likely that the test really is stuck.
> 
> Failed connection attempts would happen on each test run, sure some might have few more than others.

As for me, both explanations are plausible for now (test getting stuck due to some bug in utp or test taking too long due to failed connections) and decreasing drop-rates and therefore decreasing likelihood of failed connection establishment is much easier hypothesis to test. If the test will still spuriously fail with new values then I will have more certanity that there is something wrong with utp.

PR #512, opened 2022-06-16
Fix stability issues

mjfh:
Employing the new `snap` protocol there might be different views on what
 the messages really contain (currently specs are more a hint.)

TBC

### DIFF
--- a/eth/common/eth_types.nim
+++ b/eth/common/eth_types.nim
@@ -454,21 +454,16 @@ proc readTxTyped(rlp: var Rlp, tx: var Transaction) {.inline.} =
   let txType = rlp.getByteValue
   rlp.position += 1
 
-  try:
-    case TxType(txType):
-      of TxEip2930:
-        rlp.readTxEip2930(tx)
-        return
-      of TxEip1559:
-        rlp.readTxEip1559(tx)
-        return
-      else:
-        discard
-  except RangeError:
-    discard
-
-  raise newException(UnsupportedRlpError,
-    "TypedTransaction type must be 1 or 2 in this version, got " & $txType)
+  case txType.int:
+    of TxEip2930.ord:
+      rlp.readTxEip2930(tx)
+      return
+    of TxEip1559.ord:
+      rlp.readTxEip1559(tx)
+      return
+    else:
+      raise newException(UnsupportedRlpError,
+        "TypedTransaction type must be 1 or 2 in this version, got " & $txType)
 
 
 proc read*(rlp: var Rlp, T: type Transaction): T =
--- a/eth/common/eth_types.nim
+++ b/eth/common/eth_types.nim
@@ -3,6 +3,9 @@ import
   stew/[endians2, byteutils], chronicles, stint, nimcrypto/[keccak, hash],
   ../rlp, ../trie/[trie_defs, db]
 
+from stew/objects
+  import checkedEnumAssign
+
 export
   stint, read, append, KeccakHash, rlp, options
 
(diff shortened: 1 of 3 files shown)


eth/common/eth_types.nim
```diff
@@ -454,14 +454,22 @@ proc readTxTyped(rlp: var Rlp, tx: var Transaction) {.inline.} =
...
+    case TxType(txType):
+      of TxEip2930:
+        rlp.readTxEip2930(tx)
+        return
+      of TxEip1559:
+        rlp.readTxEip1559(tx)
+        return
+      else:
+        discard
+  except RangeError:
```
zah, line 467: `RangeError` is a Defect. You are not supposed to handle it, but rather you should prevent it from happening in the first place. 

Defects like that depend on compilation settings that control whether they will be detected and how they will be signaled (through an exception or through direct program termination). Thus, if you allow the code to produce a defect, under different compilation settings it can produce a more seriuous problem like a memory corruption or even a security vulnerability.

zah (changes requested):

eth/common/eth_types.nim
mjfh: I believe that the error handling belongs here and is an extension of the default (aka `else:` case() of the switch.

Ok, what I can do is comparing numerical values of the enum type to avoid the de-serialisation range error.

A similar issue happens within the function `auth.decodeAuthMessageEIP8()`. I had cases where the `eciesDecryptedLength(int(size))` directive bailed out with an integer range error.

zah: Do you know where is the origin of the `RangeError`? You just need to add an explicit check there and raise an exception derived from `CatchableError` instead of relying on the `RangeError` detection.

mjfh: Which case? the `eth_types.readTxTyped()` function gets tainted data and should parse de-seitalise it. If the data are wrong, is must detected here and not outside (which would contradict the whole purpose of the logic of this function).

As for the other case, the exception happened in `ecies.eciesDecryptedLength(int(size))` as stated above.

mjfh: Actually, the construction that led to the `RangeErrorI` was reported by the stack dump) was
```
   let size = uint16.fromBytesBE(m)
   ...
   var buffer = newSeq[byte](eciesDecryptedLength(int(size)))
```

zah: > let size = uint16.fromBytesBE(m)
   ...
   var buffer = newSeq[byte](eciesDecryptedLength(int(size)))
   
It's not clear here which line can throw a range error. A range error is usually something like out-of-bounds indexing within an array or a conversion between int and range types where the converted value is outside of the range of the destination type.

mjfh: My solution was useless anyway -- still the problem remains. Will wait for the next crash :)

eth/common/eth_types.nim
```diff
@@ -454,15 +454,18 @@ proc readTxTyped(rlp: var Rlp, tx: var Transaction) {.inline.} =
   let txType = rlp.getByteValue
   rlp.position += 1
 
-  case TxType(txType):
```
zah, line 457: [`checkedEnumAssign`](https://github.com/status-im/nim-stew/blob/master/stew/objects.nim#L94) is a potentially safer construct that you can use here.

The main benefit is that you'll be able to use a regular case statement over the loaded enum value, ensuring that when new values are added to the enum, the compiler will remind you to update the case statement.

mjfh: ok

zah (approved):

eth/common/eth_types.nim
```diff
@@ -454,15 +454,18 @@ proc readTxTyped(rlp: var Rlp, tx: var Transaction) {.inline.} =
...
-    of TxEip2930:
+  case txType.int:
```
kdeme, line 457: Could also use `checkedEnumAssign` from nim-stew here for a safe conversion.

eth/common/eth_types.nim
mjfh: Bingo, that is what I need :)

eth/common/eth_types.nim
mjfh: see above, thanks

kdeme (approved):

PR #522, opened 2022-08-04
Add connection limits to utp

KonradStaniec:


eth/utp/utp_router.nim
```diff
@@ -82,6 +82,7 @@ proc getUtpSocket[A](s: UtpRouter[A], k: UtpSocketKey[A]): Option[UtpSocket[A]]
 proc deRegisterUtpSocket[A](s: UtpRouter[A], socket: UtpSocket[A]) =
   s.sockets.del(socket.socketKey)
   utp_established_connections.set(int64(len(s.sockets)))
+  debug "Removed utp socket", dst = socket.socketKey, lenSockets = len(s.sockets)
```
kdeme, line 85: I'd stick with consistent naming for registering and deregistering.

eth/utp/utp_router.nim
```diff
@@ -181,6 +183,14 @@ proc processPacket[A](r: UtpRouter[A], p: Packet, sender: A) {.async.}=
     if (maybeSocket.isSome()):
       debug "Ignoring SYN for already existing connection"
     else:
+      if (len(r.sockets) >= r.socketConfig.maxNumberOfOpenConnections):
```
kdeme, line 186: How does the flow work exactly in case of failure due to limits?
- Is there a FIN send to the remote end?
- How does the application side know the socket (it is listening on) is closed or needs to be closed?

KonradStaniec: In general I stick to reference implementation (and libtorrent) behaviour i.e
- when limit is hit, SYN remains unanswered so the remote side will eventually timeout. (after 2 retries)
- socket is started/allocated only if the limits are not exceeded, so the application is not aware that there was connection try and does not need to do anything.

eth/utp/utp_router.nim
KonradStaniec: you are probably right. I will improve naming with some small cleanup pr.

PR #523, opened 2022-08-05
Make discv5 timeouts configurable

KonradStaniec:
Sometimes it maybe useful to increase those timeouts for example in bad network environment or in tests on weaker machines.


eth/p2p/discoveryv5/protocol.nim
```diff
@@ -914,13 +921,32 @@ func init*(
...
     tableIpLimits: TableIpLimits(
       tableIpLimit: tableIpLimit,
       bucketIpLimit: bucketIpLimit),
-    bitsPerHop: bitsPerHop
+    bitsPerHop: bitsPerHop,
+    handshakeTimeout: handshakeTimeout,
+    responseTimeout: responseTimeout
+  )
+
+func init*(
```
kdeme, line 938: Could avoid this `func` by having the first init call set those defaults.

PR #524, opened 2022-08-10
Allow direct access to the DB driver object ref

mjfh:


arnetheduck:
the (better) alternative here is to capture the concrete reference you pass to kvstore "outside" and also pass it to the user that needs a concrete instance, instead of exposing a typeless interface here.

mjfh:
I did that first but found it ugly, then. Maybe I can do better now .. let's see

arnetheduck:
this is what we do elsewhere, where we need a specific instance - it ensures that kvstore-based code remains pure which simplifies reasoning about what will break if we change DB engine - code that requires a specific backend should be explicit in this requirement

mjfh:
yes, can do better -- works now all in nimbus-eth1 :)

mjfh:
Ahhh, that was that unit test easy part.

In real life we work with `TrieDatabaseRef` which is in nim-eth and just one level above `ChainDB` (from nimbus-eth1) which can hold the back end information. This is why it gets ugly now.

arnetheduck:
nonetheless, there exists a master location in the "outer" wrapper that creates the database and configures it (presumably in nimbus-eth1) - it is from here that you should pass the instance to whoever else needs it, in addition to passing it to KvStore. The kvstore is not meant to leak its database this way.

mjfh:
Ok then :)

PR #531, opened 2022-09-02
clean up eth_types

arnetheduck:
`eth_types` is being imported from many projects and ends up causing
long build times due to its extensive import lists - this PR starts
cleaning some of that up by moving the chain DB and RLP to their own
modules.

this PR also moves `keccakHash` to its own module and uses it in many
places.


arnetheduck:
Failing `devel` build is a std lib regression unrelated to this PR

kdeme (approved):

PR #534, opened 2022-09-06
fix discovery v4 ping pong handler

jangko:
The goal of this PR is to fix missing additional ping in [ping handler](https://github.com/ethereum/devp2p/blob/master/discv4.md#ping-packet-0x01). 

The spec says after recv ping, should respond with pong and additional ping if last pong received exceeds BOND_EXPIRATION(12h). This PR add the missing additional ping.

The additional ping also paired with a pong matcher to prevent invalid pong hash amplification attack. 

With this PR, the disc v4 test suite in hive is cleared.

### DIFF
--- a/eth/p2p/kademlia.nim
+++ b/eth/p2p/kademlia.nim
@@ -19,6 +19,9 @@ logScope:
   topics = "kademlia"
 
 type
+  # 32 bytes NodeId | 16 bytes ip | 1 byte mode
+  TimeKey = array[49, byte]
+
   KademliaProtocol* [Wire] = ref object
     wire: Wire
     thisNode: Node
@@ -27,6 +30,7 @@ type
     pingFutures: Table[Node, Future[bool]]
     neighboursCallbacks: Table[Node, proc(n: seq[Node]) {.gcsafe, raises: [Defect].}]
     rng: ref HmacDrbgContext
+    pingPongTime: OrderedTable[TimeKey, int64] # int64 -> unix time
 
   NodeId* = UInt256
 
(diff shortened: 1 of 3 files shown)


eth/p2p/kademlia.nim
```diff
@@ -81,6 +94,68 @@ proc hash*(n: Node): hashes.Hash = hash(n.node.pubkey.toRaw)
...
+
+  if k.pingPongTime.len < MaxEntries:
+    return
+
+  k.pingPongTime.sort(cmp, order = SortOrder.Descending)
+  var
+    num = 0
+    rc: array[MaxRC, TimeKey]
+
+  for v in keys(k.pingPongTime):
```
kdeme, line 139: You don't have to iterate here over all the keys right?

eth/p2p/kademlia.nim
```diff
@@ -81,6 +94,68 @@ proc hash*(n: Node): hashes.Hash = hash(n.node.pubkey.toRaw)
...
+  k.pingPongTime.getOrDefault(n.timeKeyPong, 0'i64).fromUnix
+
+proc cmp(x, y: (TimeKey, int64)): int =
+  if x[1] < y[1]: return -1
+  if x[1] > y[1]: return 1
+  0
+
+proc removeTooOldPingPongTime(k: KademliaProtocol) =
+  const
+    MaxEntries = 128
```
kdeme, line 128: Would this be enough? How did you get to this value?

eth/p2p/kademlia.nim
```diff
@@ -496,16 +571,37 @@ proc recvPong*(k: KademliaProtocol, n: Node, token: seq[byte]) =
   var future: Future[bool]
   if k.pongFutures.take(pingid, future):
     future.complete(true)
+  k.updateLastPongReceived(n, getTime())
```
kdeme, line 574: Should this also be added for other message types? Not sure, but the devp2p spec talks about the general term "communication"

jangko: not sure either

eth/p2p/kademlia.nim
kdeme: Making it configurable would help for testing values in the application that uses it (nimbus-eth1), but it is not necessarily needed.

I was mostly curious if 128 is enough or not and thus if many new ping/pongs would be triggered still (perhaps a metric could be of use here).
On the other hand, worst thing that can happen for `MaxEntries` being too low, is that the node would do more ping/pongs. Additionally, the bound expiration of 12h is rather long, so maybe the smaller MaxEntries and more ping/pongs because of it is actually a good thing.

jangko: I've modified the algorithm. Instead of using hard coded limit or user supplied limit, I choose to use dynamic limit with lower bound=128 and upper bound=2*routing.len. max number of removal is 25% from current limit.

kdeme (approved):

PR #536, opened 2022-09-14
Safer mpt proof verification

KonradStaniec:
Adds safer alternative to `isValidBranch` for  mpt proof verification.

Improvements:
- No unexpected exceptions and assertions. There are still some assertions in rlp methods (like `doAssert self.isList()`) but there are proper checks that they won't happen. (and if they happen then our `rlp` may be broken
- Proper handling of missing keys
- stricter parsing of mpt nodes.

This changes makes this procedure more geared to verifying proofs coming from untrusted sources.

### DIFF
--- a/eth/trie/hexary_proof_verification.nim
+++ b/eth/trie/hexary_proof_verification.nim
@@ -94,14 +94,14 @@ proc getNextNode(nodeRlp: Rlp, key: NibblesSeq): Result[NextNodeResult, string] 
       return err("invalid reference")
 
     if nextRef.isList:
-      let rawBytes = ?nextRef.getRawRlpBytes()
+      let rawBytes = ? nextRef.getRawRlpBytes()
       if len(rawBytes) > 32:
-        return err("Embededed node longer that 32 bytes")
+        return err("Embedded node longer than 32 bytes")
       else:
         currNode = nextRef
         restKey = restKey.slice(keyLen)
     else:
-      let nodeBytes = ?nextRef.blobBytes()
+      let nodeBytes = ? nextRef.blobBytes()
       if len(nodeBytes) == 32:
         return ok(
           NextNodeResult(
@@ -113,16 +113,16 @@ proc getNextNode(nodeRlp: Rlp, key: NibblesSeq): Result[NextNodeResult, string] 
       elif len(nodeBytes) == 0:
         return ok(NextNodeResult(kind: EmptyValue))
       else:
-        return err("referernce rlp blob should have 0 or 32 bytes")
+        return err("reference rlp blob should have 0 or 32 bytes")
 
   while true:
-    let listLen = ?currNode.getListLen()
+    let listLen = ? currNode.getListLen()
 
     case listLen
     of 2:
       let
-        firstElem = ?currNode.getListElem(0)
-        blobBytes = ?firstElem.blobBytes()
+        firstElem = ? currNode.getListElem(0)
+        blobBytes = ? firstElem.blobBytes()
 
       let (isLeaf, k) = hexPrefixDecode blobBytes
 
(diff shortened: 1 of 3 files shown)


kdeme (commented):
from https://github.com/status-im/nimbus-eth1/pull/1223#issuecomment-1246377960

> 
> Imo trie `isValidBranch` maybe useful to validate consistency of proofs generated by trie `getBranch`. I would just not use it to validate proofs which come from the network, as it has some assertions in it and its parsing of mpt nodes is a bit to lax.

Perhaps that piece of info can be added in comment of the function call, to make it clear that these addtions are supposed to be safe for untrusted data and the old one is not (it probably was never intented to be).

eth/trie/hexary_proof_verification.nim
```diff
@@ -0,0 +1,228 @@
...
+  var restKey = key
+
+  template handleNextRef(nextRef: Rlp, keyLen: int) =
+    if not nextRef.hasData:
+      return err("invalid reference")
+
+    if nextRef.isList:
+      let rawBytes = ?nextRef.getRawRlpBytes()
+      if len(rawBytes) > 32:
+        return err("Embededed node longer that 32 bytes")
```
kdeme, line 99:
```suggestion
        return err("Embedded node longer than 32 bytes")
```

eth/trie/hexary_proof_verification.nim
```diff
@@ -0,0 +1,228 @@
...
+          NextNodeResult(
+            kind: HashNode,
+            nextNodeHash: nodeBytes,
+            restOfTheKey: restKey.slice(keyLen)
+          )
+        )
+      elif len(nodeBytes) == 0:
+        return ok(NextNodeResult(kind: EmptyValue))
+      else:
+        return err("referernce rlp blob should have 0 or 32 bytes")
```
kdeme, line 116:
```suggestion
        return err("reference rlp blob should have 0 or 32 bytes")
```

eth/trie/hexary_proof_verification.nim
```diff
@@ -0,0 +1,228 @@
...
+proc getNextNode(nodeRlp: Rlp, key: NibblesSeq): Result[NextNodeResult, string] =
+  var currNode = nodeRlp
+  var restKey = key
+
+  template handleNextRef(nextRef: Rlp, keyLen: int) =
+    if not nextRef.hasData:
+      return err("invalid reference")
+
+    if nextRef.isList:
+      let rawBytes = ?nextRef.getRawRlpBytes()
```
kdeme, line 97: We typically space the `?`
```suggestion
      let rawBytes = ? nextRef.getRawRlpBytes()
```

eth/trie/hexary_proof_verification.nim
```diff
@@ -0,0 +1,228 @@
...
+    key: seq[byte],
+    value: seq[byte]): MptProofVerificationResult =
+  ## Verifies provided proof of inclusion (trie branch) against provided trie
+  ## root hash.
+  ## Distinguishes 3 possible results:
+  ## - proof is valid but key is not part of the trie
+  ## - proof is invalid
+  ## - proof is valid
+  ## In case of valid proof, value is extracted from the leaf node and compared
+  ## agains provided value
```
kdeme, line 199:
```suggestion
  ## against provided value
```

eth/trie/hexary_proof_verification.nim
```diff
@@ -0,0 +1,228 @@
...
+
+  if maybeProofValue.isNone():
+    return missingKey()
+
+  let proofValue = maybeProofValue.unsafeGet()
+
+  if proofValue == value:
+    return validProof(proofValue)
+  else:
+    return invalidProof("proof does not contrain expected value")
```
kdeme, line 228:
```suggestion
    return invalidProof("proof does not contain expected value")
```

kdeme (approved):

PR #543, opened 2022-10-11
move eth1 specific code to where it belongs

jangko:
turn out Nim compiler devel branch introduces breaking changes:
- shallowCopy is removed
- default gc is ORC

both changes requires refactor. that's why I need to touch a lot of files.


eth/p2p/discoveryv5/routing_table.nim
```diff
@@ -482,10 +482,10 @@ proc len*(r: RoutingTable): int =
 proc moveRight[T](arr: var openArray[T], a, b: int) =
   ## In `arr` move elements in range [a, b] right by 1.
   var t: T
-  shallowCopy(t, arr[b + 1])
+  t = system.move(arr[b + 1])
```
kdeme, line 485: What does this do without ORC / Does it work without ORC? Considering ORC is only defaulted recently, I'm a bit wary of using this, not sure on the details of it though, and how it applies to older versions.

This is why I asked last week what the plan is on this and if there isn't yet, perhaps it is better to just default disable ORC for now.

@zah @arnetheduck ?

eth/p2p/p2p_protocol_dsl.nim
```diff
@@ -1,7 +1,7 @@
 {.push raises: [Defect].}
 
 import
-  std/[options, sequtils],
+  std/[options, sequtils, macrocache],
```
kdeme, line 4: Just wanted to note that `p2p_protocol_dsl.nim` is also used in nimbus-eth2, so be sure that it works there also.

@zah you might want to have a look at this too for this reason

eth/p2p/rlpx.nim
```diff
@@ -131,6 +131,7 @@ proc requestResolver[MsgType](msg: pointer, future: FutureBase) {.gcsafe.} =
           doAssert false, "trying to resolve a timed out request with a value"
       except CatchableError as e:
         debug "Exception in requestResolver()", exc = e.name, err = e.msg
+        discard e
```
kdeme, line 134: Why is this needed?

tests/fuzzing/rlpx/thunk.nim
```diff
@@ -1,7 +1,7 @@
 import
  testutils/fuzzing, chronos,
   ../../../eth/p2p, ../../../eth/p2p/rlpx, ../../../eth/p2p/private/p2p_types,
-  ../../../eth/p2p/rlpx_protocols/eth_protocol,
+  ../../p2p/eth_protocol,
```
kdeme, line 4: The Fuzzing target becomes much less useful without an implemented protocol. Eventually these would be more useful to have versions in nimbus-eth1

kdeme (commented):
I kind of had in my mind of keeping everything devp2p (https://github.com/ethereum/devp2p/) related inside nim-eth as it is quite contained as spec and thus in theory good for a library. In theory, as I do realize that the actual implementation of these protocols require access to database, so perhaps having it in nimbus-eth1 is just more practical...? Unsure myself.

I do realize that we already have duplicates of eth protocol inside nimbus-eth1, so I am rather for removing the duplicates, no matter where the final code lives.

I would have liked this PR rather split in two: one PR which removes old, unused & duplicate code and another PR with the code fixes for nim-devel. That way if the latter one has impact on existing applications, it can be more easily rolled back.

eth/p2p/discoveryv5/routing_table.nim
jangko: not related to ORC. it's related to nim version >= 1.7.0 which remove `shallowCopy`.

eth/p2p/rlpx.nim
jangko: to silence declared but not used compiler warning.

eth/p2p/discoveryv5/routing_table.nim
kdeme: Yes, I realize that, but it is related to ORC in the sense that because of the default switch to ORC, shallowCopy is removed.
My question is, what does move do here in an non ORC enabled Nim?

eth/p2p/rlpx.nim
kdeme: In the tests I assume? That's probably because it is build for the tests with `-d:chronicles_log_level=ERROR`. Perhaps we should alter that instead.

eth/p2p/discoveryv5/routing_table.nim
jangko: `move` will behave like `shallowCopy` + `reset source` both in ORC enabled and non ORC Nim. the behaviour will not change whether ORC enabled or not.

jangko:
I've split the PR into two part, and as expected, it will fail the test in CI for nim-devel

kdeme (approved):

kdeme:
> I've split the PR into two part, and as expected, it will fail the test in CI for nim-devel

Yeah, that is fine. As long as we understand why this can be merged. Could even make PRs not fail on nim-devel CI temporarily in such instances (as long as nothing in production depends on nim-devel, which it never should :) )

PR #544, opened 2022-10-14
Fix related to nim devel branch

jangko:
I've changed the way to silence unused warning based on chronicles log level.

### DIFF
--- a/eth/keyfile/keyfile.nim
+++ b/eth/keyfile/keyfile.nim
@@ -93,7 +93,10 @@ type
   KfResult*[T] = Result[T, KeyFileError]
 
 proc mapErrTo[T, E](r: Result[T, E], v: static KeyFileError): KfResult[T] =
-  r.mapErr(proc (e: E): KeyFileError = v)
+  if r.isOk:
+    ok(r.value)
+  else:
+    err(v)
 
 const
   SupportedHashes = [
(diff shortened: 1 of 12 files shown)


kdeme (approved):

zah:
If we rely on a compile-time counter, we can go all the way and store the table of protocols in a constant.

Otherwise, the changes here should be tested in nimbus-eth2 as well.

jangko:
we can experiment on it. const can store simple function pointer. as long as we don't put too much nim-ish features into the protocol table.

arnetheduck:
I think it's time to move the protocol dsl macro into each of nimbus-eth1/nimbus-eth2 and specialize them for devp2p / libp2p respectively - trying to serve both creates unnecessary complexity because the two are not really _that_ similar and we never want to use the same code to support both libp2p and devp2p.

kdeme:
@jangko Was this intentionally merged?

It was clearly not tested for nimbus-eth2 as I just tried bumping it for other changes I want to do and hit several issues: https://github.com/status-im/nim-eth/issues/550

Perhaps my approval was not fully clear here: There was still discussion ongoing (even though with low action rate, hence my latest comment) so merging th PR was not fully appropriate imo. My approval was meant as "I don't see issues with it, but it needs to be tested and work with nimbus-eth2 (stated several times in comments here and in the previous PR that was holding this work) and others need to agree too."
I also would expect a follow up PR in nimbus-eth2 after this PR (Or during, in case of tricky changes that should get tested a bit up front) so other don't get hit with the fallout.

jangko:
oops, sorry.
how can I revert a protected branch? make a new PR?

PR #545, opened 2022-10-14
allow EhereumNode network state override

jangko:
This is useful in situation where we are using interface-like for protocol network state and then have multiple implementation. e.g. Eth67 can switch to post merge protocol handler during _runtime_ from pre merge protocol handler. or another scenario, switching between sync algorithm.

I know we can also modify  networkstate using `node.protocolState(Eth).blabla = something`, but this changes reduce indirection access.

### DIFF
--- a/tests/p2p/test_protocol_handlers.nim
+++ b/tests/p2p/test_protocol_handlers.nim
@@ -86,3 +86,9 @@ suite "Testing protocol handlers":
       peer.isNil == true
       # To check if the disconnection handler did not run
       node1.protocolState(hah).count == 0
+
+  test "Override network state":
+    let rng = newRng()
+    var node = setupTestNode(rng, hah)
+    node.addCapability(hah, network())
+    node.replaceNetworkState(hah, network())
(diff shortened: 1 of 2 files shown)


tests/p2p/test_protocol_handlers.nim
```diff
@@ -86,3 +86,9 @@ suite "Testing protocol handlers":
       peer.isNil == true
       # To check if the disconnection handler did not run
       node1.protocolState(hah).count == 0
+
+  test "Override network state":
+    let rng = newRng()
+    var node = setupTestNode(rng, hah)
+    node.addCapability(hah, network())
+    node.replaceNetworkState(hah, network())
```
kdeme, line 94: Perhaps add some check of a value in the state that has changed?

kdeme (approved):

tests/p2p/test_protocol_handlers.nim
jangko: okay, done

PR #546, opened 2022-10-14
more descriptive silence unused warning

jangko:


kdeme (commented):
Sorry, I should have explained better in your previous PR about what I disliked exactly and how I preferred the fix.
The issue was not that much about the fact that the `discard` was not clear enough.
What I disliked was the need of adding extra clutter code to **each** of these cases to dismiss a compiler hint that is invalid (well, it is valid for that compile time setting). This solution, while more clear of what it does, adds actually more clutter.

So what I should have clarified better is that we could fix it by simply always compiling with `-d:chronicles_log_level=TRACE`. (side note: I noticed that in nimble we actually already remove all hints so you won't see them currently). 

For the tests this will show then of course lots of logs, but this could be fixed by adding `-d:chronicles_runtime_filtering=TRUE` and adding a ` setLogLevel(INFO)` call. 
nim-libp2p for example does this in a stub file: https://github.com/status-im/nim-libp2p/blob/fa5d102370cff4e68b47ac633153b22609b8aa41/tests/stublogger.nim#L5

If you prefer, I can also make this change.

jangko:
okay, I give up on this one, it's not too important anyway.

PR #549, opened 2022-11-07
Default compile all tests and binaries with TRACE log level

kdeme:
In order to avoid unused warning and more importantly to make sure all log statements work.

Other solution to https://github.com/status-im/nim-eth/pull/546

### DIFF
--- a/tests/stubloglevel.nim
+++ b/tests/stubloglevel.nim
@@ -1,0 +1,6 @@
+import chronicles
+
+{.used.}
+
+when defined(chronicles_runtime_filtering):
+  setLogLevel(ERROR)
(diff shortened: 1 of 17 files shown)


tests/stubloglevel.nim
```diff
@@ -0,0 +1,6 @@
+import chronicles
+
+{.used.}
+
+when defined(chronicles_runtime_filtering):
+  setLogLevel(ERROR)
```
jangko, line 6: the title says `TRACE`, but why here is `ERROR`?

kdeme: `TRACE` at compile time, to make sure all log statements are tested.
At run time we set tests to `ERROR` so that we avoid log spam in the test output.

jangko (approved):

PR #552, opened 2022-11-09
Improve the discv5 initialization logs

kdeme:
- Unify the debug log and info log into one. Removing the redundant information
- Log the custom ENR fields more pretty
- Make the JSON format logging more pretty for several types

### DIFF
--- a/eth/p2p/discoveryv5/enr.nim
+++ b/eth/p2p/discoveryv5/enr.nim
@@ -13,6 +13,7 @@
 import
   std/[strutils, macros, algorithm, options],
   nimcrypto/[keccak, utils], stew/shims/net, stew/[base64, results],
+  chronicles,
   ".."/../[rlp, keys]
 
 export options, results, keys
@@ -489,6 +490,9 @@ proc `$`(f: Field): string =
   of kList:
     "(Raw RLP list) " & "0x" & f.listRaw.toHex
 
+func `$`*(fp: FieldPair): string =
+  fp[0] & ":" & $fp[1]
+
 proc `$`*(r: Record): string =
   result = "("
   result &= $r.seqNum
@@ -526,3 +530,5 @@ proc read*(rlp: var Rlp, T: typedesc[Record]):
 
 proc append*(rlpWriter: var RlpWriter, value: Record) =
   rlpWriter.appendRawBytes(value.raw)
+
+chronicles.formatIt(seq[FieldPair]): $it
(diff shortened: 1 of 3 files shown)


eth/p2p/discoveryv5/protocol.nim
```diff
@@ -173,6 +173,10 @@ const
     responseTimeout: defaultResponseTimeout
   )
 
+chronicles.formatIt(seq[FieldPair]): $it
+chronicles.formatIt(Option[Port]): $it
+chronicles.formatIt(Option[ValidIpAddress]): $it
```
arnetheduck, line 178: hm.. I can never tell where these formatters should go - ie I'm not sure if this will conflict somehow if another module declares the same formatter (since this is not the "owner" module).. @zah?

arnetheduck (approved):

eth/p2p/discoveryv5/enr.nim
```diff
@@ -489,6 +489,9 @@ proc `$`(f: Field): string =
   of kList:
     "(Raw RLP list) " & "0x" & f.listRaw.toHex
 
+func `$`*(fp: FieldPair): string =
+  fp[0] & ": " & $fp[1]
```
arnetheduck, line 493: I'd maybe drop the space here (to keep the pair together when copy-pasting)

arnetheduck (approved):

eth/p2p/discoveryv5/protocol.nim
kdeme: I have the same issue. I wanted to put the `FieldPair` one in the ENR module, but I don't like it that I have to import chronicles just for that.

But I think you are talking rather about the ones for `Port` and `ValidIpAddress`, as they are more commonly used. At least it are the `Option` versions, which is less common I guess.

arnetheduck: >  have to import chronicles just for that.

I think this might be the least bad option, because that's the obvious singleton module for it - the alternative I see is an "xxx_log" module for every `xxx` which is .. annoying. It would probably make sense to have a `chronicles/format` import that _only_ imports the stuff needed to create formatters for types, and not the rest - also, chronicles should declare its own `LogJson` writer and not infect the Json conversion

arnetheduck: and yeah, `Port` etc are "far" from discv5 in terms of ownership which makes them more likely to conflict

PR #554, opened 2022-11-10
Fix RLP deserialzation for Enum with holes

kdeme:
RLP Enum deserialization would currently not check if "hole values" were attempted to be converted to the enum type. Now use checkedEnumAssign and fail with RlpTypeMismatch on invalid values.

There is at least one occurance of an enum with holes in rlpx p2p: DisconnectionReason. For this enum the issue could occur.

Also:
- Added enum RLP tests and rlpx p2p disconnect message tests to test the DisconnectionReason with enum hole value.
- Fixed worse custom DisconnectionReason decoding occurance in rlpx in waitSingleMsg proc where this issue could occur.

### DIFF
--- a/tests/p2p/test_rlpx_thunk.json
+++ b/tests/p2p/test_rlpx_thunk.json
@@ -59,15 +59,25 @@
     "error": "RlpTypeMismatch",
     "description": "listElem to assert on not a single entry list"
   },
-  "Listing single element list with entry off enum range": {
-    "payload": "01c116",
+  "DisconnectReason: single element list with entry out off enum range": {
+    "payload": "01c111",
     "error": "RlpTypeMismatch",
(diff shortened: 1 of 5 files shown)


tests/p2p/test_rlpx_thunk.json
```diff
@@ -59,15 +59,25 @@
     "error": "RlpTypeMismatch",
     "description": "listElem to assert on not a single entry list"
   },
-  "Listing single element list with entry off enum range": {
-    "payload": "01c116",
+  "DisconnectReason: single element list with entry out off enum range": {
+    "payload": "01c111",
```
kdeme, line 63: Note: I'm only changing these test values to test edge cases better

mjfh (approved):

PR #561, opened 2022-11-22
Added various procs that return Options.

AdamSpitz:
The idea is to make it possible to have tries that don't necessarily include all the nodes (so that we can have a stateless client that fetches only the necessary ones).

### DIFF
--- a/eth/trie/db.nim
+++ b/eth/trie/db.nim
@@ -1,7 +1,7 @@
 {.push raises: [Defect].}
 
 import
-  std/[tables, hashes, sets],
+  std/[tables, hashes, sets, options],
   "."/[trie_defs, db_tracing]
 
 type
@@ -233,6 +233,15 @@ proc contains*(db: TrieDatabaseRef, key: openArray[byte]): bool =
 
   if db.containsProc != nil:
     result = db.containsProc(db.obj, key)
+
+proc maybeGet*(db: TrieDatabaseRef, key: openArray[byte]): Option[seq[byte]] =
+  # FIXME-Adam: Could duplicate the structure of get, but for now let's just
+  # do this (I still don't know whether this overall approach makes any sense
+  # at all.)
+  if db.contains(key):
+    some(db.get(key))
+  else:
+    none[seq[byte]]()
 
 # TransactionID imitate subset of JournalDB behaviour
 # but there is no need to rollback or dispose
--- a/eth/trie/hexary.nim
+++ b/eth/trie/hexary.nim
@@ -1,4 +1,5 @@
 import
+  std/options,
   std/tables,
   nimcrypto/[keccak, hash],
   ../rlp,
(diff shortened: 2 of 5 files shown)


eth/trie/db.nim
```diff
@@ -1,7 +1,7 @@
 {.push raises: [Defect].}
 
 import
-  std/[tables, hashes, sets],
```
arnetheduck, line 4: fyi, in general we tend to prefer `results.Opt` over `options` since it's more flexible - specially for new code

eth/trie/hexary.nim
```diff
@@ -688,3 +689,101 @@ proc isValidBranch*(branch: seq[seq[byte]], rootHash: KeccakHash, key, value: se
...
+
+
+
+# The code below has a lot of duplication with the code above; I needed
+# versions of get/put/del that don't just assume that all the nodes exist.
+# Maybe there's some way to eliminate the duplication without screwing
+# up performance? But for now I don't want to meddle with the existing
+# code, for fear of breaking it. --Adam, Nov. 2022
+
+template maybeKeyToLocalBytes(db: DB, k: TrieNodeKey): Option[seq[byte]] =
```
zah, line 701: Since templates work on the AST level, their arguments will be inserted verbatim in the template body (to be evaluated multiple times potentially). This meant that if the input expression has side-effects, these side-effects will be repeated multiple times (if you use the template parameter multiple times in the body).

There are two ways to avoid this:

1) Use a regular `proc` or `func` instead of a template when you need to refer to some of the input parameters multiple times (this is usually the preferred approach).

2) Bind the input parameter to a local variable at the start of the template. You can then reference the local variable as many times as necessary.

zah:
How come there are no unit tests for the new functionality?

eth/trie/hexary.nim
AdamSpitz: Yes, that confused me too. I wrote `maybeKeyToLocalBytes` that way because `keyToLocalBytes` (earlier in the file) was written that way.

    template keyToLocalBytes(db: DB, k: TrieNodeKey): seq[byte] =
      if k.len < 32: k.getLocalBytes
      else: db.get(k.asDbKey)

Shall I change them both to procs?

AdamSpitz: I'm looking at the generated C code, trying to understand why keyToLocalBytes was written as a template in the first place, but I'm a bit confused.

I thought maybe the idea was that keyToLocalBytes was written as a template because maybe writing it as a proc would mean that its result (the `seq[byte]`) would be a whole big array of bytes that would need to be copied into a local variable (or whatever) at the call site. But when I try changing keyToLocalBytes to a proc and then look at the generated C code, it doesn't look like that's what's happening; I *think* the keyToLocalBytes proc is just returning a single-word pointer (of type `tySequence__6H5Oh5UUvVCLiakt9aTwtUQ*`). So I have no idea why it was a template in the first place.

So maybe it's safe (performance-wise) to change keyToLocalBytes and maybeKeyToLocalBytes into procs? (I'd certainly *prefer* them to be procs, for the sake of code cleanliness/simplicity/correctness; I just thought there might have been some performance-related reason why it was originally written as a template.)

arnetheduck: > So maybe it's safe (performance-wise) to change keyToLocalBytes and maybeKeyToLocalBytes into procs? 

in general, we prefer proc's even when there's a small performance penalty - said penalty is often theoretic because in actual reality, we compile things with LTO and the compiler is usually better at deciding what should be inlined or not.

We use templates:

* because it's the most convenient syntax sometimes in terms of getting access to local symbols etc
* because in rare exceptional cases, the underlying compiler has been proven to generate bad code
* because someone prematurely thought it would be a performance issue without actually verifying it

The last point happens more often than necessary, and you should feel free to get rid of any templates that look like that - Nim already does many other things under the hood for proc performance and `proc` (or even better, `func`) offer superior behavior / language semantics for understanding the code in almost all cases.

eth/trie/hexary.nim
```diff
@@ -32,11 +33,35 @@ proc expectHash(r: Rlp): seq[byte] =
     raise newException(RlpTypeMismatch,
       "RLP expected to be a Keccak hash value, but has an incorrect length")
 
+type MissingNodeError* = ref object of Defect
```
arnetheduck, line 36: you should not inherit from Defect if the exception is meant to be caught - defect is only for logic errors that lead to undefined behavior, where the application crashes after reporting it.

jangko:
Where is the code to build the incomplete tree?
Don't you need it to test `maybeGet`?

PR #565, opened 2022-11-29
Make discv5 `protocol-id` configurable at compile time

LNSD:
Make `protocol-id` and `version` selectable at compile time allowing applications to span a separate discv5 network.

These changes are required for the Waku discv5 protocol. They were made several months ago, but they never made it to the `master` branch.


kdeme (approved):
I'm fine with adding this, especially as Waku is using it already for months.

Ideally there is one discv5 DHT, but it is simply not feasible right now.

arnetheduck:
ideally this would be a runtime option so that you can instantiate two discovery networks - ie imo for waku it would make sense to run both discoveries at the same time, with the waku-specific one as a "crutch"

PR #567, opened 2022-12-01
normalise nimble, ci

arnetheduck:
* drop obsolete dep download

### DIFF
--- a/eth.nimble
+++ b/eth.nimble
@@ -1,3 +1,5 @@
+mode = ScriptMode.Verbose
+
 version       = "1.0.0"
 author        = "Status Research & Development GmbH"
 description   = "Ethereum Common library"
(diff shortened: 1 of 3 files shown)


eth.nimble
```diff
@@ -18,78 +20,70 @@ requires "nim >= 1.2.0",
...
 task test_rlp, "Run rlp tests":
   # workaround for github action CI
   # mysterious crash on windows-2019 64bit mode
   # cannot reproduce locally on windows-2019
   # running in virtualbox
-  let releaseMode = if existsEnv"PLATFORM":
-                      getEnv"PLATFORM" != "windows-amd64"
-                    else: true
+  let releaseMode =
+    if existsEnv"PLATFORM" and getEnv"PLATFORM" != "windows-amd64":
```
arnetheduck, line 63:
```suggestion
    if existsEnv"PLATFORM" and getEnv"PLATFORM" == "windows-amd64":
```

jangko:
Get this log message before the test crash on Windows CI-nim1.6

image

kdeme:
> Get this log message before the test crash on Windows CI-nim1.6
> 

I can have a look at this, been going through quite some uTP code anyhow :)

jangko:
Thanks. I will focus on nim-websock library .

kdeme:
fyi, while I couldn't immediately reproduce this locally, the issue seems to have been due to the vast amount of logging occuring. This was happening due to the `skipParentCfg` causing the build to ignore https://github.com/status-im/nim-eth/blob/master/tests/nim.cfg#L6-L8 which caused it not to set it to the runtime loglevel selected here: https://github.com/status-im/nim-eth/blob/master/tests/stubloglevel.nim#L6

@arnetheduck If the idea is that all repos should build with `skipParentCfg`, I can add seperate `nim.cfg` / `config.nims` files per test group folder, or perhaps just use a task that builds & runs all (but the individual test tasks are rather useful when developing).

PR #568, opened 2022-12-05
Add `eth` topic to all the chronicles log topics

LNSD:
In nwaku, we are working on enabling structured logging (https://github.com/waku-org/nwaku/issues/1195). 

Since `nim-eth` does not support the JSON serialization of certain types (e.g., `PublicKey`), we would like to disable all chronicles logs generated by the library at compile time via `chronicles_disabled_topics` option. 

This PR adds to all the chronicle's `logScope` topics a shared topic: `eth`.

### DIFF
--- a/eth/net/nat.nim
+++ b/eth/net/nat.nim
@@ -45,7 +45,7 @@ var
   externalUdpPort: Port
 
 logScope:
-  topics = "nat"
+  topics = "eth net nat"
 
 ## Also does threadvar initialisation.
 ## Must be called before redirectPorts() in each thread.
--- a/eth/p2p.nim
+++ b/eth/p2p.nim
@@ -16,6 +16,10 @@ import
 export
   p2p_types, rlpx, enode, kademlia
 
+logScope:
+  topics = "eth p2p"
+
+
 proc addCapability*(node: var EthereumNode,
                     p: ProtocolInfo,
                     networkState: RootRef = nil) =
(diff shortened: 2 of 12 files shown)


arnetheduck:
cc @zah what's the ETA of distinct / log-specific json? ie it would be reasonable to add "logging" json support for these types but it must absolutely not leak outside of that domain and into other json outputs.

kdeme (commented):
Aside from nim-libp2p, I don't think we are currently doing this type of "namespace" logScope (not sure how else to call it). Granted, there are probably only few of our libraries using chronicles (nim-libp2p, nim-eth, nim-json-rpc, nim-presto, ...?).

I'm not necessarily against it, but we should then decide to do this for all libraries that require logging.
For the top level applications I think it makes sense to stick with "local" topics.

Regarding the failed JSON serialization, I wasn't aware of this but it isn't a surprise as nim-eth is not compiled/tested in CI with this enabled. This should get enabled in CI and fixed.

Part of the code that is used in nimbus-eth2 does work and is tested there. The rest doesn't, which must mean that it also isn't enabled in nimbus-eth1 CI. Also there, this should be added.

Regarding the actual failures, I've had a look at it and appears to be just in discv4 (discovery and kademlia), one in p2p and rest in uTP. They are all to do with Node and IP addresses, similar fixes have been done for discv5.

kdeme (approved):

PR #570, opened 2022-12-13
extend `BlockHeader` for EIP-4844

etan-status:
Extends `BlockHeader` with `excessDataGas` according to EIP-4844 (used by Nimbus-CL in empty block prod fallback).

### DIFF
--- a/eth/common/eth_types.nim
+++ b/eth/common/eth_types.nim
@@ -121,6 +121,7 @@ type
     # `baseFee` is the get/set of `fee`
     fee*:             Option[UInt256]   # EIP-1559
     withdrawalsRoot*: Option[Hash256]   # EIP-4895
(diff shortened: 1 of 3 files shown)


zah (approved):

arnetheduck:
unrelated, but where would the earned fee go, so that one can compare blinded blocks with regular ones?

kdeme (approved):

etan-status:
into a state proof that shows that post balance of feeRecipient is higher than before applying the block.

anything else is just trust based. the regular fee is not the one you get from MEV relay, they can charge a commission as well.

etan-status:
example: https://beaconcha.in/slot/5343093

MEV recipient on top being different from fee recipient in bottom.
not sure how it interacts with tx that decrease recipient's balances

etan-status:
Nim devel failed checks are unrelated.

PR #571, opened 2022-12-13
fix `excessDataGas` type (`GasInt` --> `UInt256`)

etan-status:
Used incorrect type in original PR. Fixing.


etan-status:
Nim devel once more unrelated.

kdeme (approved):

PR #575, opened 2022-12-28
Attempt to connect to many peers concurrently, not in sequence.

AdamSpitz:
The old code, which did it in sequence, made it take a very long time to accumulate peers in the peer pool, particularly because occasionally a connection-attempt would hang for a couple of minutes before timing out.

### DIFF
--- a/eth/p2p/peer_pool.nim
+++ b/eth/p2p/peer_pool.nim
@@ -23,6 +23,8 @@ logScope:
 const
   lookupInterval = 5
   connectLoopSleep = chronos.milliseconds(2000)
+  maxConcurrentConnectionRequests = 40
+  sleepBeforeTryingARandomBootnode = chronos.milliseconds(3000)
 
 proc newPeerPool*(
     network: EthereumNode, networkId: NetworkId, keyPair: KeyPair,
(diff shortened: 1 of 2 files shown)


AdamSpitz:
This is very bare-bones, but it does seem to help quite a lot on my machine - I'm consistently getting several peers in my peer pool pretty quickly now. (Most connection attempts are still failing, but now that they're happening in parallel I do tend to get a few successes.)

eth/p2p/peer_pool.nim
```diff
@@ -162,8 +174,24 @@ proc maybeConnectToMorePeers(p: PeerPool) {.async.} =
...
-    await p.connectToNodes(p.nodesToConnect())
-
+    for n in p.nodesToConnect():
+      await p.connQueue.addLast(n)
+    
+    # The old version of the code (which did all the connection
+    # attempts in serial, not parallel) actually *awaited* all
+    # the connection attempts before reaching the code at the
+    # end of this proc that tries a random bootnode. Should
+    # that still be what happens? I don't think so; one of the
```
kdeme, line 184: Yeah, that code can be removed probably (but doesn't need to in this PR necessarily).

I'm not the author of that code, but I'm guessing the idea here is something as follows: There are much less testnet nodes than mainnet nodes in the common DHT. So it is pretty difficult to find relevant nodes. To still find relevant nodes, it just tries here to connect to one of the provided bootstrap nodes. As the bootstrap nodes are per network specific provided, it should be a relevant node. 
However, the bootstrap nodes are also more frequently connected to and thus probably more likely to be already "full" of connected peers.

kdeme (approved):

PR #590, opened 2023-02-21
ci: cancel stale builds

arnetheduck:


arnetheduck:
oops, there's https://github.com/status-im/nim-eth/pull/567 already

PR #598, opened 2023-04-18
Fix block body encoding

mjfh:


eth/common/eth_types_rlp.nim
```diff
@@ -358,10 +358,11 @@ proc rlpHash*[T](v: T): Hash256 =
 func blockHash*(h: BlockHeader): KeccakHash {.inline.} = rlpHash(h)
 
 proc append*(w: var RlpWriter, b: BlockBody) =
+  w.startList 2 + b.withdrawals.isSome.ord
```
kdeme, line 361: Strange that this worked before, as `BlockBody` has always been a list.

kdeme: Oh, never mind me. 
It used to use the general RLP code List creation from an object, but was adjusted more recently here https://github.com/status-im/nim-eth/commit/5b189ce5e22a785d1a3a3ea6a4d34387378df2b4

mjfh: Without the extra field it was the default behaviour, I suppose. The handler seems to be relatively new.

PR #600, opened 2023-04-20
remove ecdhRaw usage and replace it with ecdh+custom hash function

### LINKED ISSUE #222: Use custom hash function for eth1 ecdh
We currently extend upstream libsecp256k1 to add a `secp256k1_ecdh_raw` function - since this was implemented, upstream has added the ability do do custom hash functions that could be used to implement this ethereum-specific way of deriving keys - as such, we can stop patching upstream and move the logic into `nim-eth`.

Upstream discussion:
https://github.com/bitcoin-core/secp256k1/issues/352

Implementation blueprint:
https://github.com/bitcoin-core/secp256k1/blob/master/src/modules/ecdh/tests_impl.h#L18-L25

jangko:
fix #222

### DIFF
--- a/eth/keys.nim
+++ b/eth/keys.nim
@@ -17,7 +17,7 @@
 import
   std/strformat,
   secp256k1, bearssl/hash as bhash, bearssl/rand,
-  stew/[byteutils, objects, results],
+  stew/[byteutils, objects, results, ptrops],
   ./common/eth_hash
 
 from nimcrypto/utils import burnMem
@@ -25,7 +25,9 @@ from nimcrypto/utils import burnMem
 export secp256k1, results, rand
 
 const
-  KeyLength* = SkEcdhRawSecretSize - 1
+  FullKeyLength* = 33
+
+  KeyLength* = FullKeyLength - 1
     ## Shared secret key length without format marker
   RawPublicKeySize* = SkRawPublicKeySize - 1
     ## Size of uncompressed public key without format marker (0x04)
@@ -43,7 +45,9 @@ type
   SignatureNR* = distinct SkSignature
     ## ...but ENR uses non-recoverable signatures!
 
-  SharedSecretFull* = SkEcdhRawSecret
+  SharedSecretFull* = object
+    data*: array[FullKeyLength, byte]
+
   SharedSecret* = object
     data*: array[KeyLength, byte]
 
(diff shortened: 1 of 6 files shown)


arnetheduck (approved):

eth/keys.nim
```diff
@@ -17,15 +17,17 @@
...
+  stew/[byteutils, objects, results, ptrops],
   ./common/eth_hash
 
 from nimcrypto/utils import burnMem
 
 export secp256k1, results, rand
 
 const
-  KeyLength* = SkEcdhRawSecretSize - 1
+  FullKeyLength* = 33
```
cheatfate, line 28: Where this number `33` come from?

eth/keys.nim
```diff
@@ -230,11 +234,28 @@ func verify*(sig: SignatureNR, msg: openArray[byte], key: PublicKey): bool =
...
-  copyMem(addr result.data[0], unsafeAddr(tmp.data[1]), sizeof(result))
-
-func ecdhRawFull*(seckey: PrivateKey, pubkey: PublicKey): SharedSecretFull =
-  SharedSecretFull(ecdhRaw(SkSecretKey(seckey), SkPublicKey(pubkey)))
+proc ecdhSecretHash(output: ptr byte, x32, y32: ptr byte, data: pointer): cint
+                    {.cdecl, raises: [].} =
+  copyMem(output, x32, 32)
+  return 1
+
+func ecdhSecret*(seckey: PrivateKey, pubkey: PublicKey): SharedSecret =
```
cheatfate, line 242: This is really bad name `ecdhSecret`, it should be `ecdhSharedSecret` at least.

eth/keys.nim
```diff
@@ -230,11 +234,28 @@ func verify*(sig: SignatureNR, msg: openArray[byte], key: PublicKey): bool =
...
-  let tmp = ecdhRaw(SkSecretKey(seckey), SkPublicKey(pubkey))
-
-  # Remove first byte!
-  copyMem(addr result.data[0], unsafeAddr(tmp.data[1]), sizeof(result))
-
-func ecdhRawFull*(seckey: PrivateKey, pubkey: PublicKey): SharedSecretFull =
-  SharedSecretFull(ecdhRaw(SkSecretKey(seckey), SkPublicKey(pubkey)))
+proc ecdhSecretHash(output: ptr byte, x32, y32: ptr byte, data: pointer): cint
+                    {.cdecl, raises: [].} =
+  copyMem(output, x32, 32)
```
cheatfate, line 239: Where this number `32` come from?

cheatfate (changes requested):

eth/keys.nim
jangko: fixed

eth/keys.nim
jangko: changed to `ecdhSharedSecret`

eth/keys.nim
jangko: replaced with `KeyLength` constant

cheatfate (approved):

PR #602, opened 2023-04-21
Some changes to make hexary.nim better able to handle incomplete DBs

AdamSpitz:


AdamSpitz:
The big change here to the way hexary.nim works is that we pass around both the full path (NibblesSeq) and the index (int), rather than doing a "slice" every time we recurse down one step. So on each recursive step we just add to the int, rather than creating a whole new NibblesSeq.

I'm not doing this for efficiency, I'm doing it because it means that we still have access to the full path if we reach a missing node (which is possible when running in stateless mode). We need to be able to throw an exception containing the full path, so that (higher up, in the nimbus-eth1 code) we can send off an async request asking for the node at that path.

PR #613, opened 2023-05-30
rlp codec support optional fields

jangko:
Instead of patching BlockHeader or BlockBody codec each time it get additional optional fields,
this PR make the rlp codec automatically handle
optional fields. Thus rlp codec overloading of
EthBlock, BlockHeader, and BlockBody can be removed.

### DIFF
--- a/eth/rlp/writer.nim
+++ b/eth/rlp/writer.nim
@@ -1,6 +1,6 @@
 import
   std/options,
-  stew/shims/macros,
+  stew/[shims/macros, results],
   ./object_serialization, ./priv/defs
 
 type
@@ -187,10 +187,10 @@ proc hasOptionalFields(T: type): bool =
 
   proc helper: bool =
     var dummy: T
-    template countFields(RT, n, x) =
-      when x is Option:
+    template detectOptionalField(RT, n, x) =
+      when x is Option or x is Opt:
         return true
-    enumerateRlpFields(dummy, countFields)
+    enumerateRlpFields(dummy, detectOptionalField)
     false
 
   const res = helper()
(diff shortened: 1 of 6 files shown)


eth/rlp/writer.nim
```diff
@@ -181,22 +182,138 @@ proc appendImpl[T](self: var RlpWriter, listOrBlob: openArray[T]) =
     for i in 0 ..< listOrBlob.len:
       self.append listOrBlob[i]
 
+proc hasOptionalFields(T: type): bool =
+  mixin enumerateRlpFields
+
+  proc helper: bool =
+    var dummy: T
+    template countFields(RT, n, x) =
```
zah, line 190: This copy/pasted name could be a bit confusing. You could use something generic such as `template forEachField` or something revealing the intent such as `template detectOptionalField`. 

Let's also add support for the `Opt[T]` type from `results` which we have started introducing in many places as a replacement for `Option[T]`.

eth/rlp/writer.nim
```diff
@@ -181,22 +182,138 @@ proc appendImpl[T](self: var RlpWriter, listOrBlob: openArray[T]) =
...
+  for i in countdown(fd.high, loidx):
+    let fieldName = fd[i].name
+    let prevFields = genPrevFields(obj, fd, i-1, loidx-1)
+    result.add quote do:
+      if `obj`.`fieldName`.isSome:
+        `prevFields`
+
+  # generate something like
+  when false:
+    if obj.excessDataGas.isSome:
```
zah, line 246: So, this PR implements an assumption that `Option` fields are always added at the end of the RLP object (this is something that can be checked at compile-time) and you can never set a field to `None` unless all following fields are also `None`. This special treatment of None should be described in the README.

eth/rlp/writer.nim
jangko: `Opt[T]` added, template name fixed.

eth/rlp/writer.nim
jangko: README updated

PR #614, opened 2023-05-31
use correct exception in `parseCmdArg`

etan-status:
`parseCmdArg` is expected to raise `ValueError` but for `enr.Record`, `Node`, `PrivateKey`, and `NatConfig`, we raise `ConfigurationError`. Change to `ValueError` instead.


etan-status:
Nim 1.6 error was already present before this PR:

```
fatal.nim(54)            sysFatal
asyncfutures2.nim(426)   futureContinue

    Unhandled defect: asyncfutures2.nim(304, 14) `future.cancelcb.isNil` futures returned from `{.async.}` functions must not use `cancelCallback` [AssertionDefect]
  [FAILED] Router should clear all resources when connection future is cancelled
```

This is due to recent `nim-chronos` changes that replaced todos with asserts, so the error existed before those changes and is now being highlighted.

kdeme (approved):

kdeme:
> This is due to recent `nim-chronos` changes that replaced todos with asserts, so the error existed before those changes and is now being highlighted.

Right, something I'll have to look into :)

tools/dcli.nim
```diff
@@ -1,3 +1,11 @@
+# Copyright (c) 2020-2023 Status Research & Development GmbH
+# Licensed under either of
+#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+# at your option.
+# This file may not be copied, modified, or distributed except according to
+# those terms.
+
```
arnetheduck, line 8: should probably have slapped `raises: []` on top here

PR #615, opened 2023-06-01
fix cancellation syntax

arnetheduck:
see https://github.com/status-im/nim-eth/pull/614#issuecomment-1570403046

### DIFF
--- a/eth/utp/utp_router.nim
+++ b/eth/utp/utp_router.nim
@@ -301,21 +301,21 @@ proc innerConnect[A](s: UtpSocket[A]): Future[ConnectionResult[A]] {.async.} =
       debug "Connection cancelled", dst = s.socketKey
       raise exc
 
-proc connect[A](s: UtpSocket[A]): Future[ConnectionResult[A]] =
+proc connect[A](s: UtpSocket[A]): Future[ConnectionResult[A]] {.async.} =
   debug "Initiating connection", dst = s.socketKey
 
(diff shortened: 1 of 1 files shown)


eth/utp/utp_router.nim
```diff
@@ -301,21 +301,21 @@ proc innerConnect[A](s: UtpSocket[A]): Future[ConnectionResult[A]] {.async.} =
       debug "Connection cancelled", dst = s.socketKey
       raise exc
 
-proc connect[A](s: UtpSocket[A]): Future[ConnectionResult[A]] =
+proc connect[A](s: UtpSocket[A]): Future[ConnectionResult[A]] {.async.} =
   debug "Initiating connection", dst = s.socketKey
```
cheatfate, line 305:
```nim
proc connect[A](s: UtpSocket[A]): Future[ConnectionResult[A]] =
  debug "Initiating connection", dst = s.socketKey
  s.innerConnect()
```

cheatfate: There is no need to do one more cancellation handler in the wrapper procedure. `innerConnect()` will perform all the cleanup (it calls `s.destroy()`).

arnetheduck: oh, good catch :) I didn't even read internalConnect :) 484e453fb71e2bca1ba4c5f27898a824e01b95f1

PR #616, opened 2023-06-03
add/change Deneb fields per https://github.com/ethereum/execution-apis/pull/417

tersec:


tersec:
Same `devel`-only CI failure as usual,
```
/home/runner/work/nim-eth/nim-eth/eth/p2p/discoveryv5/routing_table.nim(488, 14) Error: undeclared identifier: 'shallowCopy'
```

PR #617, opened 2023-06-06
typealias for allowing different RNG contexts

etan-status:
Instead of requiring the BearSSL based `HmacDrbgContext`, take a generic `SecureRngContext` instead. For now, it is aliased to `HmacDrbgContext`, but this change allows using different RNG where BearSSL is unavailable.

### DIFF
--- a/doc/discv5.md
+++ b/doc/discv5.md
@@ -15,7 +15,7 @@ and `rlp`.
 
 ```Nim
 let
-  rng = keys.newRng
+  rng = SecureRngContext.new()
   privKey = PrivateKey.random(rng[])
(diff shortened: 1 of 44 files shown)


etan-status:
@tersec mentioned `std/sysrand` as another alternative, but that one is Nim 1.6+ only, and still pending security review.

etan-status:
Splitting this into multiple PRs

arnetheduck:
`std/sysrand` remains a non-alternative - it's not viable due to it being underspecified, differing between platforms and running out of randomness, block etc, in particular on shared servers and other similar environments - as discussed elsewhere, `Hmac..` serves to take a small amount of random input and turn it into a large, practically inexhaustible pool with decent performance.

PR #628, opened 2023-08-01
specify exceptions of callback called by exception-free function

tersec:
`randomNodes` supposedly cannot raise an exception, but it takes a callback which might.

Also some `proc` to `func`


tersec:
Usual CI pattern: 1.6 passes, `devel` fails.

PR #638, opened 2023-09-22
fix build for `dcli` and instructions for new path

etan-status:
`dcli` got moved and the guide is no longer accurate. Update the paths. Also, relative imports were incorret inside `dcli.nim`, and error when using Nim 2.0 compiler.

### DIFF
--- a/doc/discv5.md
+++ b/doc/discv5.md
@@ -88,13 +88,13 @@ node, by providing its ENR.
 # Make sure you have the latest modules, do NOT trust nimble on this.
 nimble install
 # Build dcli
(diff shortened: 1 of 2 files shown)


etan-status:
Thanks @parithosh for catching this :)

etan-status:
`Nim devel` problem alreay present on `master`:

```
test api usage ............... (0.00s)
rlp encoding ... (0.00s)
eip 2718 transaction ........... (0.01s)
Segmentation fault (core dumped)
Error: execution of an external program failed: '/home/runner/work/nim-eth/nim-eth/tests/rlp/all_tests'
```

kdeme (commented):
Cool, thanks!

PR #645, opened 2023-10-29
fix XDeclaredButNotUsed warnings in tests/

tersec:
Potentially, these point to more things to test, but sometimes not, and even when so, it's easier to search for `_` or `discard` than it is some variable which turns out not to be used.


tersec:
Usual success/failure pattern

PR #650, opened 2023-12-21
Added additional tests to cover HexaryTrie MPT proofs.

web3-developer:


tests/trie/test_hexary_proof.nim
```diff
@@ -59,3 +60,88 @@ suite "MPT trie proof verification":
...
+    let db = newMemoryDB()
+    var trie = initHexaryTrie(db)
+
+    let 
+      proof = trie.getBranch("not-exist".toBytes)
+      res = verifyMptProof(proof, trie.rootHash, "not-exist".toBytes, "not-exist".toBytes)
+
+    check: 
+      trie.rootHash == keccakHash(emptyRlp)
+      proof.len() == 1 # Maybe this should return an empty list?
```
kdeme, line 77: I think it makes sense that it is `len() == 1` considering it drops in the empty rlp value?

The fact that the result is `InvalidProof` is however inconsistent for when you would add a "not-exist` into a not empty trie. The emptry trie thing is quite the edge case though.

tests/trie/test_hexary_proof.nim
```diff
@@ -59,3 +60,88 @@ suite "MPT trie proof verification":
...
       res.isMissing()
+
+  # proof test ref:
+  # - https://github.com/ethereum/go-ethereum/blob/master/trie/proof_test.go
+  # - https://github.com/ethereum/py-trie/blob/master/tests/test_proof.py
+  test "Validate proof for empty trie":
+    let db = newMemoryDB()
+    var trie = initHexaryTrie(db)
+
+    let
```
kdeme, line 71: I noticed quite some trailing whitespace in this and previous PR. You might want to setup up some marking and/or trimming of that.

web3-developer: Do you mean at the bottom of the file? Or spaces after the let?

tests/trie/test_hexary_proof.nim
web3-developer: Yeah it is an edge case so probably not worth spending too much time on. I raised it because I noticed that this library behaves differently than the Rust implementation which returns an empty proof when the trie is empty.

tests/trie/test_hexary_proof.nim
kdeme: spaces after the let and some other locations.

tests/trie/test_hexary_proof.nim
kdeme: I'd suggest to add a Note in comment that mentions this and perhaps also link to this rust library.

web3-developer: Sure, will do

tests/trie/test_hexary_proof.nim
web3-developer: Ok thanks

web3-developer:
@kdeme I've updated the code as you suggested. Ready for the final review. Thanks

kdeme (approved):

PR #653, opened 2024-01-12
Use cancelSoon instead of deprecated cancel

kdeme:


kdeme:
waiting with merging this, as I'm not sure if we should go ahead before chronos gets a new version (This would brake with latest tagged one)

kdeme:
Closing for https://github.com/status-im/nim-eth/pull/667

PR #661, opened 2024-01-16
Adjust test names and comments for `blobGasUsed` field

etan-status:
`blobGasUsed` was not mentioned in `test_rlp_codec` test names, despite being used. Further, update the idea of `genOptionalFieldsValidation` to also check for `blobGasUsed`, and fix the check to catch unsupported combinations of optional fields.


etan-status:
- https://github.com/status-im/nimbus-eth2/pull/5763

PR #667, opened 2024-01-19
Fix chronos related deprecated warnings in discv5 and uTP

kdeme:


eth/p2p/discoveryv5/protocol.nim
```diff
@@ -1041,11 +1041,11 @@ proc close*(d: Protocol) =
 
   debug "Closing discovery node", node = d.localNode
   if not d.revalidateLoop.isNil:
-    d.revalidateLoop.cancel()
+    d.revalidateLoop.cancelSoon()
```
arnetheduck, line 1044: generally, it's might be reasonable to `await noCancel .. cancelAndWait()` to simplify reasoning about order of shutdown

kdeme: Oh, I was not yet aware of the `noCancel`, thanks for pointing that out. I'll add it.

Do you mean to also better split the shutdown of the loops and the actual transport, to avoid sends in those loops on closed transport?

Something like this?
```
proc closeWait*(d: Protocol) {.async.} =
  doAssert(not d.transp.closed)

  debug "Closing discovery node", node = d.localNode
  var futures: seq[Future[void]]
  if not d.revalidateLoop.isNil:
    futures.add(d.revalidateLoop.cancelAndWait())
  if not d.refreshLoop.isNil:
    futures.add(d.refreshLoop.cancelAndWait())
  if not d.ipMajorityLoop.isNil:
    futures.add(d.ipMajorityLoop.cancelAndWait())

  await noCancel(allFutures(futures))

  await noCancel(d.transp.closeWait())
```

And a `proc close(..` then that just `AsyncSpawn`s the `closeWait()` (and perhaps deprecate that not awaited one).

arnetheduck: > Do you mean to also better split the shutdown of the loops and the actual transport, to avoid sends in those loops on closed transport?

I haven't looked at the shutdown flow in detail but this sounds reasonable .. usually when we shut down nimbus there's a bunch of discovery errors logged from what I suspect is this misordering of shutdown events. The point of `cancelAndWait` (vs `cancelSoon`) is to give some determinism to that order so that we don't get these odd effects.

Oh, and in general a "close" function is async, uses `cancelAndWait`, doesn't raise any errors and stops cancellation (via the `noCancel` blocker), we tend to call it `proc closeWait() {.async: (raises: []).}` which is a nice and simple pattern for controlled closing.

PR #675, opened 2024-02-13
Use asyncraises in p2p

jangko:


eth/p2p/private/p2p_types.nim
```diff
@@ -155,7 +155,7 @@ type
   MessageHandlerDecorator* = proc(msgId: int, n: NimNode): NimNode
 
   ThunkProc* = proc(x: Peer, msgId: int, data: Rlp): Future[void]
-    {.gcsafe, raises: [RlpError].}
+    {.gcsafe, async: (raises: [RlpError, CatchableError]).}
```
arnetheduck, line 158: there's no point adding `CatchableError` to raises - that's already the default at which point it's generally better to use just plain `async`

eth/p2p/rlpx.nim
```diff
@@ -176,7 +176,7 @@ proc messagePrinter[MsgType](msg: pointer): string {.gcsafe.} =
   # result = $(cast[ptr MsgType](msg)[])
 
 proc disconnect*(peer: Peer, reason: DisconnectionReason,
-  notifyOtherPeer = false) {.gcsafe, async.}
+  notifyOtherPeer = false) {.gcsafe, async: (raises:[CatchableError]).}
```
arnetheduck, line 179: no need for `gcsafe` when `async` is specified

eth/p2p/private/p2p_types.nim
jangko: okay, thanks

arnetheduck: this would be because you need to match usages to the declaration - ie either all use `{.async.}` or all use `{.async: [errors].}` - this is related to the guarantees that you want to uphold, ie either the thunk proc is responsible for handling errors _or_ the thing calling it is and one needs to decide.

broadly though, this declaration is non-sensical: `RlpError` is a subclass of `CatchableError` so if this declaration is needed, something went wrong "somewhere else" and it would be interesting perhaps to investigate where that was - ie likely, there is some deeper change that will either remove `RlpError` _or_ remove `CatcahbleError` from this declaration.

jangko: Yup, the rlpx needs overhaul.

PR #681, opened 2024-04-13
Enable chronos dualstack with new constructor.

cheatfate:


eth/p2p/discoveryv5/protocol.nim
```diff
@@ -1018,17 +1072,30 @@ proc newProtocol*(
...
+proc open*(d: Protocol) {.raises: [TransportOsError].} =
   info "Starting discovery node", node = d.localNode,
     bindAddress = d.bindAddress
 
   # TODO allow binding to specific IP / IPv6 / etc
-  let ta = initTAddress(d.bindAddress.ip, d.bindAddress.port)
-  d.transp = newDatagramTransport(processClient, udata = d, local = ta)
-
+  d.transp =
+    newDatagramTransport(processClient, udata = d,
```
kdeme, line 1096: Do I understand correctly that with the changes in https://github.com/status-im/nim-chronos/commit/0d050d582306e8c521c3a4a6f6dcb3c83c93a90f, the behaviour here is:
- If no `bindIp` is provided it will try the IPv6 AnyAdress and next IPv4 AnyAddress as explained here: https://github.com/status-im/nim-chronos/commit/0d050d582306e8c521c3a4a6f6dcb3c83c93a90f#diff-137db3be35f90e323d096fbe9b76bf20e43a92bbb161cac69229dc530acd29f2R927-R931
- if `bindIp` is given it will try to bind to given IPv4/IPv6 address.

cheatfate: 1) YES, but it also will unset `IPV6_ONLY` flag on socket. In such way it enables dual-socket support.
2) YES.

cheatfate: On Unix such sockets usually get marked with `*`, so if you will list sockets opened in your system it should be listed as `*:<port>`, but in worst case it will be listed as `0.0.0.0:<port>` (when IPv6 is disabled).

PR #682, opened 2024-05-06
Introduce wrapper type for EIP-4844 transactions

etan-status:
EIP-4844 blob sidecars are a concept that only exists in the mempool. After inclusion of a transaction into an execution block, only the versioned hash within the transaction remains. To improve type safety, replace the `Transaction.networkPayload` member with a wrapper type `PooledTransaction` that is used in contexts where blob sidecars exist.


jangko:
I run nimbus-eth1/feat/eip-6493 branch with hive and it looks good. It actually pass two more test case. Exactly where we have problem with blob transactions.

jangko (approved):

PR #683, opened 2024-05-23
rlp: refresh code

arnetheduck:
A first step in cleaning up RLP, which has lots of interesting issues - the next step would be to clean up the exception handling as well (Resultify?)

* remove `RlpNode` (unused)
* single-pass parsing for most functionality via RlpItem
* stricter conformance to spec
  * remove float support
  * warn about signed integers
  * raise on invalid RLP earlier
* avoid several pointless allocations, in particular in `listLen`, `listElem` etc
* include spec docs


jangko:
Looks like you hit the heisenbug in the rlp decoder. I've been chasing it but have not get to it. Perhaps we need a new approach to write the decoder.

arnetheduck:
> Perhaps we need a new approach to write the decoder.

yeah, this was already quite a significant rewrite .. the most "unusual" thing I see in this code is the returning of `var self` from the `items` iterator - other than that, it's hard to imagine what could be causing it.. I wanted to deprecate that iterator, but it's used all over the place.

One thought I had was to make `RlpItem` public and return that from the iterator - it would act as a "soft" reference that can be used to extract payloads from the rlp instance, though of course if the rlp mutates between getting the item and using it, it would break.

It's not a perfectly safe approach, but maybe less .. unusual.

arnetheduck:
Nim bug (in orc): https://github.com/nim-lang/Nim/issues/23645

PR #688, opened 2024-06-04
faster `hash` for `Address`

arnetheduck:
https://github.com/nim-lang/Nim/issues/23678

### DIFF
--- a/eth/common/eth_types.nim
+++ b/eth/common/eth_types.nim
@@ -10,7 +10,7 @@
 ## from many places
 
 import
-  std/[options, strutils],
+  std/[options, hashes, strutils],
   stew/[byteutils, endians2], stint,
   ./eth_hash, ./eth_times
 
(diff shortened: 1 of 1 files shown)


eth/common/eth_types.nim
```diff
@@ -319,3 +319,20 @@ func `==`*(a, b: NetworkId): bool =
...
+
+# TODO https://github.com/nim-lang/Nim/issues/23678
+func hash*(a: EthAddress): Hash {.inline.} =
+  static: doAssert sizeof(a) == 20
+  var a0{.noinit.}, a1 {.noinit.}: uint64
+  var a2{.noinit.}: uint32
+
+  # Addresses are more or less random so we should not need a fancy mixing
+  # function
+  copyMem(addr a0, unsafeAddr a[0], sizeof(a0))
```
tersec, line 334: These will be more efficient if `EthAddress` also is required to be aligned, which `nim-eth` does not require and `array[20, byte]` might not automatically trigger since its base type has `sizeof() == 1`. But that'd have to be changed in another repo, and it's cleaner to keep the changes separate. Both can help, but this works and is useful on its own.

tersec (approved):

jangko:
nimbus-eth1 also using a lot of Hash256 as lookup key, maybe add one hash function for Hash256 too?

arnetheduck:
> nimbus-eth1 also using a lot of Hash256 as lookup key, maybe add one hash function for Hash256 too?

One problem here would be that we use the same alias of `MDigest` in eth2 as well:
https://github.com/status-im/nimbus-eth2/blob/d2a07514541ffe6ee02a2ec7272ce7a315131e04/beacon_chain/spec/digest.nim#L134 - there, we've already fixed this silly problem - with an even more simple implementation.

One thing I wasn't sure of was whether to use all bytes of `EthAddress` - on the one hand, it should be random, on the other, I'm not sure it has uniform distribution - for hash on the other hand, it already has uniform distribution so any bytes will do.

cc @cheatfate wdyt of adding the hash from beacon chain to nimcrypto?

cheatfate:
Such trick should not be used for SHA256 because you significantly lowering its chance for uniform distribution while taking only first 8 bytes. Such trick could work with 3rd generation of digests like KECCAK/BLAKE2 (this functions has such function), but not with second generation like SHA256.

arnetheduck:
For the purpose of `Hash`, does that matter though? Ie I'm guessing you're referring to the lack of proof that it is sufficiently random in a security context, but for the purpose of selecting a bucket in Table, it seems to me like the bar is much lower?

That said, I don't think it's the end of the world if we use this `xor` variant on `MDigest` - it'll have to read a bit more memory but that's likely not going to change anything significant since it's likely going to be sitting in the same cache line.

The question is more whether we can add it to nimcrypto?

arnetheduck:
https://github.com/cheatfate/nimcrypto/pull/79 adds some infrastructure so we can unroll an xor implementation for `MDigest` that does all bits

PR #690, opened 2024-06-07
Add async raises annotations for discv5

kdeme:


eth/p2p/discoveryv5/protocol.nim
```diff
@@ -686,8 +691,12 @@ proc lookup*(d: Protocol, target: NodeId): Future[seq[Node]] {.async.} =
...
       error "Resulting query should have been in the pending queries"
 
-    let nodes = query.read
-    # TODO: Remove node on timed-out query?
+    let nodes =
+      try:
+        query.read
+      except CatchableError:
+        raiseAssert("query should have been successfully completed")
+
```
kdeme, line 699: I don't like it much that I need to try/accept here for `CatchableError`, but I don't know how to do this else with usage of `one()` higher up.

When doing an `await d.lookupWorker(n, target)` it would not be needed as `lookupWorker` does not async raise (aside from `CancelledError`).

arnetheduck: `var pendingQueries = newSeqOfCap[Future[seq[Node]].Raising([CancelledError])](alpha)` should fix this - ie you need to preserve the types in seqs..

Something like
`var pendingQueries = newSeqOfCap[typeof(d.lookupWorker(default(X), default(Y))](alpha)` might also work

arnetheduck: it's a short path through (mostly :), re-`await` recommended for most cases for now - might come back to that decision at some point, but ...

eth/p2p/discoveryv5/protocol.nim
```diff
@@ -686,8 +691,9 @@ proc lookup*(d: Protocol, target: NodeId): Future[seq[Node]] {.async.} =
     else:
       error "Resulting query should have been in the pending queries"
 
-    let nodes = query.read
-    # TODO: Remove node on timed-out query?
+    # Note: When using query.read as it gives:
+    # Error: read(query) can raise an unlisted exception: FuturePendingError
```
arnetheduck, line 695: this is the documented / intended behavior of read - not sure it deserves special mention here ;)

kdeme: Ah, I had just force-pushed a different comment

arnetheduck (approved):

PR #692, opened 2024-06-07
Add async raises annotations for uTP code

kdeme:


kdeme:
Hitting some strange compiler errors here on some of the Nim 2.0 and devel runs.

I'm not hitting these issues locally with Nim 2.0.4 and devel (but only tested from nimbus-build-system env so far).

kdeme:
This compile failure is not occurring when building locally from nimbus-eth1/vendor/nim-eth repository with `env.sh` loaded. Also not after bumping to latest chronos.

kdeme:
Minimized the compiler error to this code snippet:
```
import chronos

type
  Socket* = object
    a*: int

  ErrorType* = enum
    ErrorA, ErrorB

  Error* = object
    case kind*: ErrorType
    of ErrorA:
      discard
    of ErrorB:
      discard

proc connect(s: Socket): Future[Error] {.async: (raises: [CancelledError]).} =
  discard

let socket = Socket(a: 123)
let res = waitFor socket.connect()
```

Not using the `Future[Error]` but instead `Future[ErrorType]` avoids the compiler error. So it seems to be similar problem as in https://github.com/nim-lang/Nim/issues/20653, but with the difference that it doesn't hit for Nim 1.6.x but it does for 2.0.x and devel.

In the uTP code we can workaround this easily as the object variant is not even needed/used.

ringabout:
> Not using the Future[Error] but instead Future[ErrorType] avoids the compiler error. So it seems to be similar problem as in https://github.com/nim-lang/Nim/issues/20653, but with the difference that it doesn't hit for Nim 1.6.x but it does for 2.0.x and devel.

fixed in https://github.com/nim-lang/Nim/pull/23706

kdeme:
> > Not using the Future[Error] but instead Future[ErrorType] avoids the compiler error. So it seems to be similar problem as in [nim-lang/Nim#20653](https://github.com/nim-lang/Nim/issues/20653), but with the difference that it doesn't hit for Nim 1.6.x but it does for 2.0.x and devel.
> 
> fixed in [nim-lang/Nim#23706](https://github.com/nim-lang/Nim/pull/23706)

Great, thanks!

PR #699, opened 2024-06-12
Reduce compiler warnings related to rlp encode int

jangko:


eth/p2p/rlpx.nim
```diff
@@ -1000,7 +1000,7 @@ proc p2pProtocolBackendImpl*(protocol: P2PProtocol): Backend =
       var `rlpWriter` = `initRlpWriter`()
       const `perProtocolMsgIdVar` {.used.} = `msgId`
       let `perPeerMsgIdVar` = `perPeerMsgIdValue`
-      `append`(`rlpWriter`, `perPeerMsgIdVar`)
+      `append`(`rlpWriter`, `perPeerMsgIdVar`.uint)
```
arnetheduck, line 1003: this one we should change in the rplx layer - this conversion here is invalid

arnetheduck: in fact, we should make it a hard error when it's signed

jangko: rlpx need cleanup and major refactoring. will open an issue for the time being.

arnetheduck: ok, let's remove this from here for now then - it's better to have the warning in the code until it is fixed - it's a dangerous conversion that can panic, hidden deep in macro code => not godd.

PR #705, opened 2024-06-18
Use results.Opt instead of option in discv5, utp and nat

kdeme:
+ some other minor cleanups


kdeme:
First CI run failed due to nim-nat-traversal miniupnp bump, but this should fix that: https://github.com/status-im/nim-nat-traversal/pull/25

PR #707, opened 2024-06-21
Overhaul of ENR implementation - part I

kdeme:
- Rework adding and updating of fields by having an insert call that gets used everywhere. Avoiding also duplicate keys. One side-effect of this is that ENR sequence number will always get updated on an update call, even if nothing changes.
- Deprecate initRecord as it is only used in tests and is flawed
- Assert when predefined keys go into the extra custom pairs. Any of the predefined keys are only to be passed now via specific parameters to make sure that the correct types are stored in ENR.
- Clearify the Opt.none behaviour for Record.update
- When setting ipv6, allow for tcp/udp port fields to be used default
- General clean-up
- Rework/clean-up completely the ENR tests.

### DIFF
--- a/eth/p2p/discoveryv5/enr.nim
+++ b/eth/p2p/discoveryv5/enr.nim
@@ -11,7 +11,7 @@
 {.push raises: [].}
 
 import
-  std/[strutils, macros, algorithm, net],
+  std/[strutils, sequtils, macros, algorithm, net],
   nimcrypto/[keccak, utils],
   stew/base64,
   results,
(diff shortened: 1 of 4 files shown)


eth/p2p/discoveryv5/enr.nim
```diff
@@ -270,80 +309,46 @@ func get*(r: Record, T: type PublicKey): Opt[T] =
...
+    ip: Opt[IpAddress] = Opt.none(IpAddress),
+    tcpPort: Opt[Port] = Opt.none(Port),
+    udpPort: Opt[Port] = Opt.none(Port),
+    extraFields: openArray[FieldPair] = []):
+    EnrResult[void] =
+  ## Update a `Record` with given ip address, tcp port, udp port and optional
+  ## custom k:v pairs.
+  ##
+  ## If none of the k:v pairs are changed, the sequence number of the `Record`
+  ## will still be incremented and a new signature will be applied.
```
kdeme, line 324: I've checked and allthe usages through discv5 `updateRecord` call in nimbus-eth2 will only call this if the provided extraField actually changes, avoiding a seqNum update on same update with same data. Nevertheless, perhaps this is still something to re-add...

eth/p2p/discoveryv5/enr.nim
```diff
@@ -94,8 +100,45 @@ func `==`(a, b: Field): bool =
...
+  if index.isSome():
+    pairs[index.get()] = item
+  else:
+    pairs.insert(item, pairs.lowerBound(item, cmp))
+
+func insert(pairs: var seq[FieldPair], b: openArray[FieldPair]) =
+  ## Insert all items in key:value pairs.
+  for item in b:
+    pairs.insert(item)
+
```
kdeme, line 141: Could use some other structure than a sequence  (e.g. btree , table, etc.) but that seemed (?) kind of overkill considering an ENR will hold typically hold 8 k:v pairs in beacon, or perhaps 11 when we add ipv6, and can be max. 300 bytes.

PR #720, opened 2024-08-14
Treat putting empty data in hexary trie as deleting data

etan-status:
Other implementations of MPT delete entries when attempting to put empty value, because empty value cannot exist in RLP. We should match the behaviour.

- https://github.com/ethereum/py-trie/pull/109

Also cross-checked with Geth and Ethereumjs implementations.


etan-status:
This fixes nimbus-eth2 block hash computation when there are transactions with 0 bytes, aligning it with other ELs.

tersec (approved):

PR #733, opened 2024-09-20
Align core types with execution spec

arnetheduck:
Since these types were written, we've gained an executable spec:

https://github.com/ethereum/execution-specs

This PR aligns some of the types we use with this spec to simplify comparisons and cross-referencing.

Using a `distinct` type is a tradeoff between nim ergonomics, type safety and the ability to work around nim quirks and stdlib weaknesses.

In particular, it allows us to overload common functions such as `hash` with correct and performant versions as well as maintain control over string conversions etc at the cost of a little bit of ceremony when instantiating them.

Apart from distinct byte types, `Hash32`, is introduced in lieu of the existing `Hash256`, again aligning this commonly used type with the spec which picks bytes rather than bits in the name.


kdeme (commented):
I'm all in for these changes that align more with the exec. spec., and have consistency with nimbus-eth2.

eth/common/eth_address.nim
```diff
@@ -0,0 +1,85 @@
...
+
+template to*(v: array[20, byte], _: type Address): Address =
+  Address(v)
+
+template data*(v: Address): array[20, byte] =
+  distinctBase(v)
+
+template default*(_: type Address): Address =
+  # Avoid bad codegen where fixed bytes are zeroed byte-by-byte at call site
+  zeroAddress
```
tersec, line 30: This also is often possible by wrapping things in `static(default(Foo))`

arnetheduck: yeah, this trick allows the caller to not have to type that out .. afair, this gets better in orc so we don't really want to uglify call sites like that.

eth/common/base.nim
```diff
@@ -0,0 +1,166 @@
...
+  ## filling any missing bytes with zero.
+  ##
+  ## This is a lenient function in that `v` may contain both fewer and more
+  ## bytes than N and start might be out of bounds.
+  if v.len > 0:
+    assign(result.data, v.toOpenArray(min(start, v.len), min(start + N, v.len()) - 1))
+
+template default*[N](T: type FixedBytes[N]): T =
+  # Avoid bad codegen where fixed bytes are zeroed byte-by-byte at call site
+  const def = system.default(T)
```
tersec, line 71: noticed `anonConst` in `stew/objects`, useful pattern to enshrine in general

eth/common/base.nim
```diff
@@ -0,0 +1,166 @@
...
+  KzgCommitment* = Bytes48
+  KzgProof* = Bytes48
+
+  ForkID* = tuple[crc: uint32, nextFork: uint64] ## EIP 2364/2124
+
+func toBlockNonce*(n: uint64): BlockNonce =
+  BlockNonce(n.toBytesBE())
+
+func `==`*(a, b: NetworkId): bool =
+  a.uint == b.uint
```
tersec, line 163: Using `distinctBase` is kind of nice for these because if `NetworkId` acquires an explicit bit width/size (which it probably should, to avoid strange quirks compiling on 32-bit and 64-bit machines -- `nimbus-eth1` is 64-bit only practically but this is a library 32-bit code can reasonably use too), then this kind of construction then automatically follows. Or, this can also probably be a `{.borrow.}`

otherwise one gets silent truncation. not crashes but incorrect behavior specifically on 32-bit builds with network IDs > 2^32

eth/common/base.nim
```diff
@@ -0,0 +1,166 @@
...
+
+func `$`*(x: NetworkId): string =
+  `$`(uint(x))
```
tersec, line 166: see also re network ID/`uint64`/`distinctBase`

eth/common/keys.nim
```diff
@@ -0,0 +1,249 @@
...
+func `$`*(pubkey: PublicKey): string =
+  ## Convert public key to hexadecimal string representation.
+  toHex(pubkey.toRaw())
+
+func `$`*(sig: Signature): string =
+  ## Convert signature to hexadecimal string representation.
+  toHex(sig.toRaw())
+
+func `$`*(seckey: PrivateKey): string =
+  ## Convert private key to hexadecimal string representation
```
tersec, line 167: some of these comments have trailing `.` and some don't

arnetheduck: yeah, this is just moving the file - there's lots of stuff like this built up over the years

eth/common/base.nim
arnetheduck: hm, yeah, didn't know about that one - one thing though is that it generates a new, separate const on every instantiation even if it's the same value/type which I'm not too happy about - ie each `default(Bytes32)` would result in a separately declared constant in the generated code which is not great

eth/common/base.nim
```diff
@@ -0,0 +1,166 @@
...
+    ## This type is specialized to `Bytes4`, `Bytes8` etc below.
+    # A distinct array is used to avoid copying on trivial type conversions
+
+  ChainId* = distinct uint64
+    ## Chain identifier used for transaction signing to guard against replay
+    ## attacks between networks
+    ##
+    ## https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md
+
+  NetworkId* = distinct uint
```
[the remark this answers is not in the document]
arnetheduck: yeah, this is weird .. updated to `uint64` - ie this should always be the same as ChainId as far as types go

eth/common/base.nim
arnetheduck: fixed

eth/common/base.nim
tersec: In theory especially with LTO these could be folded into a single constant. My understanding is that this is a typical linker optimization

arnetheduck: depends - for that to happen, it must determine that the address of the variable is not used for uniqueness / comparisons which doesn't necessarily happen and/or requires special flags. 

https://stackoverflow.com/a/53077229

PR #748, opened 2024-10-11
Add ENR Record.init and update call for dual stack

kdeme:
Attempt to add dual stack support in ENR in line with the current API.
In draft however as I don't really like this API, it feels rather clumsy, both in implementation and in usage.

I'll probably implement dualstack in Fluffy to see how it would typically be used, and continue/adjust from there.

Might also go for some `EnrBuilder` approach instead to get to something in the trend of:

`let enr = EnrBuilder.init().tcp4(Port(9000)).ip4(ip4).ip6(ip6).build(privatekey)`

and

`let enrUpdated = EnrBuilder.init(enr).tcp4(Port(9000)).ip4(ip4).build(privatekey)

Ideas and/or use cases are welcome.

### DIFF
--- a/eth/p2p/discoveryv5/enr.nim
+++ b/eth/p2p/discoveryv5/enr.nim
@@ -243,7 +243,9 @@ func init*(
     extraFields: openArray[FieldPair] = []):
     EnrResult[T] =
   ## Initialize a `Record` with given sequence number, private key, optional
-  ## ip address, tcp port, udp port, and optional custom k:v pairs.
+  ## IP address, TCP port, UDP port, and optional custom k:v pairs.
+  ##
+  ## The IP address can be an IPv4 or IPv6 address.
   ##
   ## Can fail in case the record exceeds the `maxEnrSize`.
   doAssert(not hasPredefinedKey(extraFields), "Predefined key in custom pairs")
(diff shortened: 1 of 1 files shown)


eth/p2p/discoveryv5/enr.nim
```diff
@@ -254,6 +256,51 @@ func init*(
   fields.insert extraFields
   makeEnrAux(seqNum, "v4", pk, fields)
 
+func init*(
+    T: type Record,
+    seqNum: uint64, pk: PrivateKey,
+    ip4: array[4, byte],
+    ip6: array[16, byte],
```
kdeme, line 263: There is no `Opt` here, as in dual stack case I'm assuming both IPs or no IP is known (perhaps this is not correct?).

Also not using `IpAddress` as from that type I cannot be sure whether it is IPv4 or IPv6 (unless I check for it of course).

In the other use cases, the `init` call with only 1 address can be used.

eth/p2p/discoveryv5/enr.nim
```diff
@@ -254,6 +256,51 @@ func init*(
...
+  fields.insert(("ip6", ip6.toField))
+
+  if tcp4Port.isSome():
+    fields.insert(("tcp", tcp4Port.value().uint16.toField))
+  if udp4Port.isSome():
+    fields.insert(("udp", udp4Port.value().uint16.toField))
+
+  # From the ENR specification:
+  # "Declaring the same port number in both tcp, tcp6 or udp, udp6 should be avoided
+  # but doesn't render the record invalid."
```
kdeme, line 293: Perhaps I don't need to do this check and can let the user decide what they want to do in case of same ports.

eth/p2p/discoveryv5/enr.nim
```diff
@@ -357,6 +404,73 @@ func update*(
...
+  ##
+  ## This function is to be used when running in IPv4 and IPv6 dual stack mode.
+  ## The tcp6 and udp6 fields will only be set if they differ from the tcp and udp
+  ## fields.
+  ##
+  ## If none of the k:v pairs are changed, the sequence number of the `Record`
+  ## will still be incremented and a new signature will be applied.
+  ##
+  ## Providing an `Opt.none` for `tcp4Port`/`udp4Port`/`tcp6Port`/`udp6Port`/
+  ## will leave the corresponding field untouched.
```
kdeme, line 429: Perhaps it makes more sense to unset the field in that case

eth/p2p/discoveryv5/enr.nim
```diff
@@ -357,6 +404,73 @@ func update*(
...
+    pk: PrivateKey,
+    ip4: array[4, byte],
+    ip6: array[16, byte],
+    tcp4Port: Opt[Port] = Opt.none(Port),
+    udp4Port: Opt[Port] = Opt.none(Port),
+    tcp6Port: Opt[Port] = Opt.none(Port),
+    udp6Port: Opt[Port] = Opt.none(Port),
+    extraFields: openArray[FieldPair] = []):
+    EnrResult[void] =
+  ## Update a `Record` with given IPv4 and IPv6 address, optional TCP port and
```
kdeme, line 418: The fact that I have to write such a lengthy documentation for this call probably is not a sign of good API.

NagyZoltanPeter:
> Attempt to add dual stack support in ENR in line with the current API. In draft however as I don't really like this API, it feels rather clumsy, both in implementation and in usage.
> 
> I'll probably implement dualstack in Fluffy to see how it would typically be used, and continue/adjust from there.
> 
> Might also go for some `EnrBuilder` approach instead to get to something in the trend of:
> 
> `let enr = EnrBuilder.init().tcp4(Port(9000)).ip4(ip4).ip6(ip6).build(privatekey)`
> 
> and
> 
> `let enrUpdated = EnrBuilder.init(enr).tcp4(Port(9000)).ip4(ip4).build(privatekey)`
> 
> Ideas and/or use cases are welcome.

@kdeme 
Thanks for the effort! From Waku point of view I have gone through it. 
From our POV I would really vote for the builder pattern than anything else. In Waku we have something similar implemented over enr.Record.  But the approach you crafter here, above it might be even better.
I think it has greater advantages controlling what can get into the ENR or filtering duplicates, fine graining updates (with Opt.none for clearing a field).

For the draft solution, I have the feeling that making IP4 and IP6 mandatory we will still end up in incompatibilities/regression. Do you see dual stack to be forced?

arnetheduck:
>  Do you see dual stack to be forced?

re ipv6, the idea _in general_ is to push towards a world where dual stack is enabled by default, ie if the caller or user does "nothing special", they should end up in a dual stack world that serves both ipv4 and ipv6.

PR #754, opened 2024-10-23
discv5: fix potential infinite loop in randomNodes

cskiraly:
randomNodes could do an infinite loop if used with a filter expression

- added fix
- added test (goes in infinite loop without the fix)


kdeme (approved):

kdeme:
Fine with merging this as it fixes your issue.

I forgot however that this `randomNodes` call existed still in the first place, I thought I removed that a long time ago as I didn't see a proper use case for it.

How are you using that exactly or in which use case? Because for getting random nodes from the DHT is is rather bad as it only selects from the routing table.

cskiraly:
just found it reading through the code (the codex version, actually).
I just don't like infinite loops being in the code, which is easy to achieve with a filter. As you say, the whole function could also be reimplemented, or redefined and then reimplemented.

PR #757, opened 2024-10-31
probable nim bug due to compare before decrementing

chirag-parmar:
Check source branch CI for passing tests


chirag-parmar:
Was a blunder. Closing the issue

PR #765, opened 2024-12-11
Optimize RLP serialization

chirag-parmar:
## Description

This PR improves upon the memory and time complexity of the RLP serialisation algorithm(`RlpWriter`). The existing implementation of RLP uses the `RlpWriter` that extends a buffer and moves memory whenever required. This wastes a lot of time for large payloads. This PR, firstly, implements a `RlpLengthTracker` that dry runs the serialisation to calculate the full length of the serialised data and the prefixes of lists. Secondly, the `RlpTwoPassWriter` uses the pre-calculated values to assign memory in one shot and avoid moving memory to make space for prefixes.

Additionally, the PR also implements a `RlpHashWriter` that directly updates the sponge construction of keccak256 hash to avoid allocating any memory (few exceptions which are negligible).

## Rationale

Naive implementations of RLP serialisation require moving memory because prefixes of lists can only be calculated after serialising the items within the list. Since complex types such as objects are represented as lists the serialisation of any data structure in Ethereum requires moving the encoded/serialised list items to make space for the prefix. In cases where the encoded buffer is hashed, the encoded list items cannot be *absorbed* until the prefix is calculated.

However, the encoded lengths of base types can be calculated without actually serialising them (or allocating memory). This is exploited to split the computation into two. First to calculate lengths of the base types, and then to actually serialise the data. 

By calculating the lengths of the base types we can find the length of encoded list items and thereby also the length of the prefix (without actually serialising the prefix). Additionally, we also derive the total length of the serialised data from this. This logic is implemented as the `RlpLengthTracker` which records `listLengths`, `prefixLengths` and the `totalLength`.

In the second part of the computation, we actually serialise the data (and the prefixes) without skipping over memory or moving memory positions. This is implemented as the `RlpTwoPassWriter`, "two-pass" because we iterate through the data twice. Because the data is iterated through twice the `RlpTwoPassWriter` can only be used within the `encode` function. In cases where users manually call `append`, the `RlpTwoPaasWriter` would require users to manually run the `RlpLengthTracker` for the first pass and then the second pass. 

`RlpHashWriter` also uses two passes but for the second pass it absorbs the data directly into the sponge construction of keccak instead of allocating memory thereby saving memory (amortized constant usage). And for similar reasons as stated above, a new api method, `encodeHash` is exposed.

## Changelog

* introduces a new method `encodeHash` for constant memory RLP encoding and hashing.
* updates the `encode` method to serialise data in "two passes".
* isolates all writer implementations into individual files with `writer.nim` handling only the dependency injection architecture of the `append` method
* introduces two new tests, `test_hashes.nim` and `test_rlp_profiler.nim`. The first, does a sanity check of the hash writer. The second one, time profiles the new writer implementation to the default implementation.

## Benchmarks

using `--mm=refc` (refc GC)
```
CPU Time [Transaction serialization using two pass writer] 0.000001402s
CPU Time [Block Sequence serialization using two pass writer] 0.000002592s
CPU Time [Block serialization using two pass writer] 0.000007493s
.CPU Time [Transaction serialization using default writer] 0.000001559s
CPU Time [Block Sequence serailization using default writer] 0.000003281s
CPU Time [Block serialization using default writer] 0.000009589s
.CPU Time [Transaction hashing using hash writer] 0.000001810s
CPU Time [Block Sequence hashing using hash writer] 0.000006301s
CPU Time [Block hashing using hash writer] 0.000028100s
.CPU Time [Transaction hashing using default writer and then hash] 0.000002291s
CPU Time [Block Sequence hashing using default writer and then hash] 0.000007532s
CPU Time [Block hashing using defualt writer and the hash] 0.000007491s
```

using `--mm=orc` (orc GC)
```
CPU Time [Transaction serialization using two pass writer] 0.000000901s
CPU Time [Block Sequence serialization using two pass writer] 0.000002089s
CPU Time [Block serialization using two pass writer] 0.000005357s
.CPU Time [Transaction serialization using default writer] 0.000000973s
CPU Time [Block Sequence serailization using default writer] 0.000002513s
CPU Time [Block serialization using default writer] 0.000006726s
.CPU Time [Transaction hashing using hash writer] 0.000001609s
CPU Time [Block Sequence hashing using hash writer] 0.000006161s
CPU Time [Block hashing using hash writer] 0.000026231s
.CPU Time [Transaction hashing using default writer and then hash] 0.000001879s
CPU Time [Block Sequence hashing using default writer and then hash] 0.000006502s
CPU Time [Block hashing using defualt writer and the hash] 0.000006492s
```


eth/rlp/two_pass_writer.nim
```diff
@@ -0,0 +1,97 @@
...
+      let listLenBytes = prefixLen - 1
+      self.output[self.fillLevel] = LEN_PREFIXED_LIST_MARKER + byte(listLenBytes)
+      self.fillLevel += prefixLen
+      self.output.writeBigEndian(uint64(listLen), self.fillLevel - 1,
+        listLenBytes)
+
+func initTwoPassWriter*(length: int): RlpTwoPassWriter =
+  result.fillLevel = 0
+  result.listCount = 0
+  result.output = newSeqOfCap[byte](length)
```
jangko, line 82: ditto

eth/rlp/writer.nim
```diff
@@ -270,30 +178,52 @@ template append*[T](w: var RlpWriter; data: T) =
...
 proc encode*[T](v: T): seq[byte] =
   mixin append
 
-  var writer = initRlpWriter()
+  var tracker = initLengthTracker()
+  
+  tracker.append(v)
+
+  var writer = initTwoPassWriter(tracker.totalLength)
+  writer.prefixLengths = tracker.prefixLengths
```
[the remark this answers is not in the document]
jangko: And why not using ordinary array if `LIST_LENGTH` is only 50.

eth/rlp/two_pass_writer.nim
```diff
@@ -0,0 +1,97 @@
+import
+  std/options,
+  pkg/results,
+  stew/[arraybuf, assign2, shims/macros],
+  ./priv/defs,
+  utils
+
+type
+  RlpTwoPassWriter* = object
+    pendingLists*: seq[tuple[remainingItems, startPos, prefixLen: int]]
```
jangko, line 10: Should the `pendingLists` a seq? Isn't it related to object recursion? Can we use array here?

If you are worried about recursion depth, we can put compile time/static validation in `proc encode` and `proc encodeHash` and tell the user to adjust the pending list len if necessary. we can use static param with default value to avoid too much bloat code generated by generic code.

eth/rlp/hash_writer.nim
```diff
@@ -0,0 +1,89 @@
...
+
+func writeCount(writer: var RlpHashWriter, count: int, baseMarker: byte) =
+  if count < THRESHOLD_LIST_LEN:
+    writer.update(baseMarker + byte(count))
+  else:
+    let lenPrefixBytes = uint64(count).bytesNeeded
+
+    writer.update baseMarker + (THRESHOLD_LIST_LEN - 1) + byte(lenPrefixBytes)
+
+    var buf = newSeqOfCap[byte](lenPrefixBytes)
```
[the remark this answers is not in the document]
chirag-parmar: yes we could do this. Thanks for the hint.

eth/rlp/default_writer.nim
```diff
@@ -0,0 +1,106 @@
...
+    self.maybeClosePendingLists()
+  else:
+    self.pendingLists.add((listSize, self.output.len))
+
+template finish*(self: RlpDefaultWriter): seq[byte] =
+  doAssert self.pendingLists.len == 0, 
+    "Insufficient number of elements written to a started list"
+  self.output
+
+func clear*(w: var RlpDefaultWriter) =
```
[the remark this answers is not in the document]
chirag-parmar: Because `clear` clears the memory without returning the output it was added in the API for users manually adding to the buffer (without using `encode`). But this is just my guess. It isn't used anywhere AFAIK. I let it be to keep the API consistent with previous versions.

eth/rlp/hash_writer.nim
```diff
@@ -0,0 +1,89 @@
...
+      buf.setLen(listLenBytes)
+      buf.writeBigEndian(uint64(listLen), buf.len - 1, listLenBytes)
+      self.update(buf)
+
+template finish*(self: var RlpHashWriter): MDigest[self.keccak.bits] =
+  self.listLengths.setLen(0)
+  self.prefixLengths.setLen(0)
+  self.keccak.finish()
+
+func clear*(w: var RlpHashWriter) =
```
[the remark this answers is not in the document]
chirag-parmar: will maintain this discussion on the above comment. Resolving this one.

eth/rlp/writer.nim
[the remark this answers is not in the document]
chirag-parmar: > They have the same length no?

Yes, we could convert it into a sequence of tuples. Will do that.

> And why not using ordinary array if LIST_LENGTH is only 50.

It is extended if the number of lists (not only nested) required for an object increases above 50. I did this to avoid extending the sequence in increments of one using `add`. For some reason, `add` seems to be the bottleneck of the writer. Removing usage of `add` and `delete` from sequences has given me a 50% improvement over previous versions of the code.

eth/rlp/writer.nim
```diff
@@ -270,30 +178,52 @@ template append*[T](w: var RlpWriter; data: T) =
...
 
+proc encodeHash*[T](v: T): Hash32 =
+  mixin append
+
+  var tracker = initLengthTracker()
+  var writer: RlpHashWriter
+
+  tracker.append(v)
+
+  writer.prefixLengths = tracker.prefixLengths
```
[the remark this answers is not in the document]
chirag-parmar: > We can avoid copying seq using move or system.move.

Oh dang. I'll change this

> Why not embed RlpLengthTracker inside RlpHashWriter and RlpTwoPassWriter to avoid copying altogether

Thought about doing this but it didn't seem elegant. `RlpLengthTracker`, `RlpHashWriter` and `RlpTwoPassWriter` all qualify as `RlpWriter` type. As is, the entire structure of the `append` is confusing. If we use `move` not merging the types wouldn't have any downsides to it.

eth/rlp/two_pass_writer.nim
chirag-parmar: This is a great idea. You are right, we could calculate the maximum number of nested lists at compile time and just define an array. Let me implement this and get back to you.

eth/rlp/hash_writer.nim
```diff
@@ -0,0 +1,89 @@
...
+    else:
+      let listLenBytes = prefixLen - 1
+      self.update(LEN_PREFIXED_LIST_MARKER + byte(listLenBytes))
+
+      var buf = newSeqOfCap[byte](listLenBytes)
+      buf.setLen(listLenBytes)
+      buf.writeBigEndian(uint64(listLen), buf.len - 1, listLenBytes)
+      self.update(buf)
+
+template finish*(self: var RlpHashWriter): MDigest[self.keccak.bits] =
```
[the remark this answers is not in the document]
chirag-parmar: wanted to avoid cyclic dependencies and the `MDigest` to `Hash32` converter was already defined in `hashes.nim` so didn't want to make the code redundant.

eth/rlp/two_pass_writer.nim
jangko: If you managed to calculate maximum number of nested lists at compile time, it should not be too different for `listLengths` and `prefixLenghts`, their maximum len can also be calculated at compile time.
A very crude algorithm would only need to calculate the amount of fields + their amount of nested fields.

eth/rlp/hash_writer.nim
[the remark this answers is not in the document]
jangko: yes, if you import `hashes.nim` it will be no risk of cyclic dependency.

eth/rlp/writer.nim
```diff
@@ -270,30 +200,62 @@ template append*[T](w: var RlpWriter; data: T) =
...
-
 proc encode*[T](v: T): seq[byte] =
   mixin append
+  
+  const nestedListsDepth = countNestedListsDepth(type v) 
+
+  when nestedListsDepth > 0:
+    var tracker = StaticRlpLengthTracker[nestedListsDepth]()
+  elif nestedListsDepth == 0:
+    var tracker = DynamicRlpLengthTracker()
```
jangko, line 215: What kind of types need `DynamicLengthTracker`? a string/int? Can we use `StaticRlpLengthTracker[1]()`?

chirag-parmar: Types that are recursive and only fully defined during runtime. By a full definition I mean, these types assume the full structure of the object they hold only at runtime. At compile time they are empty. For example, `JsonNode`

chirag-parmar: references too

jangko: I see.

jangko:
I'm curious why block hashing using hash writer is slower than default writer + hash. And can you add block header hashing measurement too? Because in nimbus-eth1 we only use transaction hashing and block header hashing only. Almost no need to hash a full block.

chirag-parmar:
> I'm curious why block hashing using hash writer is slower than default writer + hash. 
I have been wondering the same thing for almost two weeks now. I made a lot of measurements using the time profiler on MacOS Instruments but couldn't corner it down. Then yesterday, I switched the tests, i.e. I ran the default writer + hash benchmark first and then the hash writer. This time hash writer was faster in the same order of magnitude. I think it has something to do with caching or garbage collection, I'm not sure though and but I'm trying to corner this down. 

what is weird is that, the benchmark is calculated by collecting 100 measurements and then averaging them out. So caching shouldn't be an issue. Additionally, the encoding benchmarks before hashing benchmarks use the same data which means it should already be cached

> And can you add block header hashing measurement too? Because in nimbus-eth1 we only use transaction hashing and block header hashing only. Almost no need to hash a full block.
Yes will add this

jangko:
Nim generic instantiation cache(in the compiler) is known to have a bug. And by looking into the bug tracker, I believe it still there.
https://github.com/nim-lang/Nim/issues?q=is%3Aissue+is%3Aopen+generic+cache
This is only a hunch. The exact problem is yet to be found.

chirag-parmar:
refc GC
```
CPU Time [Transaction serialization using two pass writer] 0.000001261s
CPU Time [Block Sequence serialization using two pass writer] 0.000002928s
CPU Time [Block serialization using two pass writer] 0.000006690s
CPU Time [Block header serialization using two pass writer] 0.000000858s

CPU Time [Transaction serialization using default writer] 0.000001452s
CPU Time [Block Sequence serailization using default writer] 0.000003362s
CPU Time [Block serialization using default writer] 0.000009260s
CPU Time [Block header serialization using default writer] 0.000001547s

CPU Time [Transaction hashing using hash writer] 0.000001419s
CPU Time [Block Sequence hashing using hash writer] 0.000006039s
CPU Time [Block hashing using hash writer] 0.000027683s
CPU Time [Block header hashing using hash writer] 0.000004401s

CPU Time [Transaction hashing using default writer and then hash] 0.000002267s
CPU Time [Block Sequence hashing using default writer and then hash] 0.000007470s
CPU Time [Block hashing using defualt writer and the hash] 0.000007329s
CPU Time [Block header hashing using hash writer] 0.000005271s
```

orc GC
```
CPU Time [Transaction serialization using two pass writer] 0.000000780s
CPU Time [Block Sequence serialization using two pass writer] 0.000001807s
CPU Time [Block serialization using two pass writer] 0.000005291s
CPU Time [Block header serialization using two pass writer] 0.000000830s

CPU Time [Transaction serialization using default writer] 0.000000930s
CPU Time [Block Sequence serailization using default writer] 0.000002420s
CPU Time [Block serialization using default writer] 0.000006576s
CPU Time [Block header serialization using default writer] 0.000001354s

CPU Time [Transaction hashing using hash writer] 0.000001478s
CPU Time [Block Sequence hashing using hash writer] 0.000005939s
CPU Time [Block hashing using hash writer] 0.000026274s
CPU Time [Block header hashing using hash writer] 0.000004191s

CPU Time [Transaction hashing using default writer and then hash] 0.000001850s
CPU Time [Block Sequence hashing using default writer and then hash] 0.000006511s
CPU Time [Block hashing using defualt writer and the hash] 0.000006459s
CPU Time [Block header hashing using hash writer] 0.000004640s
```

chirag-parmar:
So the benchmarks collected before are biased because of the CPU cache (I verified this). So I collected new benchmarks and ran each benchmark as a different process to avoid cache hits causing skewed/biased data. The code for the benchmarks is attached after the benchmarks.

Mem Usage
CPU Time

```

import 
  eth/[rlp, common],
  times,
  std/[os, strutils],
  stew/io2,
  results

type
  Timeval {.importc: "timeval", header:"<sys/time.h>", bycopy.} = object
  
  Rusage* {.importc: "struct rusage", header:"<sys/resource.h>", bycopy.} = object
    ru_utime {.importc.}: Timeval
    ru_stime {.importc.}: Timeval
    ru_maxrss* {.importc.}: int32  # Maximum resident set size
    # ...
    ru_minflt* {.importc.}: int32  # page reclaims (soft page faults)
  
  RusageWho* {.size: sizeof(cint).} = enum
    RusageChildren = -1
    RusageSelf = 0
    RusageThread = 1

when defined(debug):
  var H_RUSAGE_SELF{.importc, header:"<sys/resource.h".}: cint
  var H_RUSAGE_CHILDREN{.importc, header:"<sys/resource.h".}: cint
  var H_RUSAGE_THREAD{.importc, header:"<sys/resource.h".}: cint
  assert H_RUSAGE_SELF == ord(RusageSelf)
  assert H_RUSAGE_CHILDREN = ord(RusageChildren)
  assert H_RUSAGE_THREAD = ord(RusageThread)

proc getrusage*(who: RusageWho, usage: var Rusage) {.importc, header: "sys/resource.h".}

const 
  accesses  = @[AccessPair(
    address: address"0x0000000000000000000000000000000000000001", 
    storageKeys: @[default(Bytes32)]
  )]

  tx = Transaction( 
    txType: TxLegacy, 
    chainId: 7.ChainId, 
    nonce: 11253.AccountNonce, 
    gasPrice: 9.GasInt, 
    maxPriorityFeePerGas: 0.GasInt, 
    maxFeePerGas: 0.GasInt, 
    gasLimit: 88920.GasInt, 
    payload: @[99, 0, 0, 0, 25, 96, 1, 1, 56, 3, 128, 99, 0, 0, 0, 25, 96, 1, 1, 96, 0, 57, 96, 0, 243, 91, 78, 30, 176, 85, 200, 234, 14, 45, 97, 73, 65, 149, 199, 11, 118, 202, 83, 30, 211, 109, 119, 168, 184, 89, 6, 38, 132, 53, 2, 237, 54, 131, 30, 141, 225, 155, 174, 92, 96, 211, 133, 53, 218, 245, 132, 17, 173, 79, 95, 241, 197, 214, 244, 196, 37, 88, 27, 34, 51, 69, 116, 64, 170, 77, 95, 191, 152, 7, 214, 85, 249, 244, 167, 67, 76, 137, 136, 37, 169, 40, 20, 131, 165, 153, 120, 158, 20, 26, 114, 99, 129, 254, 172, 229, 99, 18, 178, 251, 40, 126, 210, 155, 108, 238, 127, 2, 156, 67, 61, 199, 191, 71, 215, 72, 23, 173, 131, 213, 35, 87, 54, 248, 41, 221, 119, 31, 223, 144], 
    accessList: @[], 
    maxFeePerBlobGas: 0.u256, 
    versionedHashes: @[], 
    authorizationList: @[], 
    V: uint64(49), 
    R: 2812423844.u256, 
    S: 532553168.u256
  )

  header = Header(
    parentHash: Hash32.fromHex("0x4d24f4dddde8ab09ea900bc87492f74e44e1c3982042d0fd01673a98e72c14a6"), 
    ommersHash: Hash32.fromHex("0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347"), 
    coinbase: Address.fromHex("0x0000000000000000000000000000000000000000"), 
    stateRoot: Root.fromHex("0xa06c17a5ebaeacef3b760293e8f2092e775448c038aa7361350cde9f27e6b218"), 
    transactionsRoot: Root.fromHex("0x773faa6ccc82788a4392df241b7c6e7175fd77648fed2aaad583a528293a80bc"), 
    receiptsRoot: Root.fromHex("0x8ccc9b260110dad8161d01e72913cd27baab66809aa106a4c28a7c7a3a418b65"), 
    difficulty: 131072.u256, 
    number: 1024.BlockNumber, 
    gasLimit: 3141592.GasInt, 
    gasUsed: 891237.GasInt, 
    timestamp: EthTime(15748660), 
    extraData: @[], 
    mixHash: Bytes32.fromHex("0x378c37bd4d8692035f9d03a619d7c44a579d3b1c1c52db3096bc92a71b74aeb5"), 
    nonce: Bytes8.fromHex("0x4ea813c0c15b4215"), 
    baseFeePerGas: Opt.some(9.u256), 
  )

  blk = EthBlock(
    header: header,
    transactions: @[
      tx, tx, tx, tx,
      tx, tx, tx, tx
    ],
    uncles: @[],
  )
  
  blk80 = EthBlock(
    header: header, 
    transactions: @[
      tx, tx, tx, tx, 
      # total of 80 txs
      tx, tx, tx, tx], 
    uncles: @[], 
  )

  blk320 = EthBlock(
    header: header, 
    transactions: @[
      tx, tx, tx, tx, tx, tx, tx, tx, 
      #.. total of 320 txs
      tx, tx, tx, tx, tx, tx, tx, tx], 
    uncles: @[], 
  )
  
  blk640 = EthBlock(
    header: header, 
    transactions: @[
      tx, tx, tx, tx, tx, tx, tx, tx, 
      #... total of 640 txs
      tx, tx, tx, tx, tx, tx, tx, tx], 
    uncles: @[], 
  )
  
  blk1280 = EthBlock(
    header: header, 
    transactions: @[
      tx, tx, tx, tx, tx, tx, tx, tx, 
      #... total of 1280 txs
      tx, tx, tx, tx, tx, tx, tx, tx], 
    uncles: @[], 
  )

proc encodeOnePass[T](v: T): seq[byte] =
  var writer = initRlpWriter()

  writer.append(v)
  move(writer.finish)

proc encodeAndHash[T](v: T): Hash32 =
  keccak256(encodeOnePass(v))

# source: https://forum.nim-lang.org/t/7238
{.push checks: off.}
template benchmark(msg: string, code: untyped) =
  when not defined(windows):
    var ru: Rusage
    getrusage(RusageSelf, ru)
    var
      rss = ru.ru_maxrss
      flt = ru.ru_minflt

  when not defined(windows):
    let start = cpuTime()

  code

  when not defined(windows):
    let stop = cpuTime()

  when not defined(windows):
    getrusage(RusageSelf, ru)
    rss = ru.ru_maxrss - rss
    flt = ru.ru_minflt - flt

  echo "Benchmark: " & msg
  when not defined(windows):
    echo "Time(s)              ", stop - start
    echo "Runtime RSS (KB):     ", rss
    echo "# of page faults:     ", flt
{.pop.}

when defined(opt)
  when defined(blk):
    benchmark "encodeHash with 8 transactions":
      let bytes2 = rlp.encodeHash(blk)
  elif defined(blk80):
    benchmark "encodeHash with 80 transactions":
      let bytes2 = rlp.encodeHash(blk80)
  elif defined(blk320):
    benchmark "encodeHash with 320 transactions":
      let bytes2 = rlp.encodeHash(blk320)
  elif defined(blk640):
    benchmark "encodeHash with 640 transactions":
      let bytes2 = rlp.encodeHash(blk640)
  elif defined(blk1280):
    benchmark "encodeHash with 1280 transactions":
      let bytes2 = rlp.encodeHash(blk1280)
else:
  when defined(blk):
    benchmark "encodeHash with 8 transactions":
      let bytes2 = encodeAndHash(blk)
  elif defined(blk80):
    benchmark "encodeHash with 80 transactions":
      let bytes2 = encodeAndHash(blk80)
  elif defined(blk320):
    benchmark "encodeHash with 320 transactions":
      let bytes2 = encodeAndHash(blk320)
  elif defined(blk640):
    benchmark "encodeHash with 640 transactions":
      let bytes2 = encodeAndHash(blk640)
  elif defined(blk1280):
    benchmark "encodeHash with 1280 transactions":
      let bytes2 = encodeAndHash(blk1280)
```

```bash
nim compile --mm:orc -d:blk -d:release --out:bench_blk test.nim
nim compile --mm:orc -d:blk80 -d:release --out:bench_blk80 test.nim
nim compile --mm:orc -d:blk320 -d:release --out:bench_blk320 test.nim
nim compile --mm:orc -d:blk640 -d:release --out:bench_blk640 test.nim
nim compile --mm:orc -d:blk1280 -d:release --out:bench_blk1280 test.nim
nim compile --mm:orc -d:blk -d:opt -d:release --out:opt_bench_blk test.nim
nim compile --mm:orc -d:blk80 -d:opt -d:release --out:opt_bench_blk80 test.nim
nim compile --mm:orc -d:blk320 -d:opt -d:release --out:opt_bench_blk320 test.nim
nim compile --mm:orc -d:blk640 -d:opt -d:release --out:opt_bench_blk640 test.nim
nim compile --mm:orc -d:blk1280 -d:opt -d:release --out:opt_bench_blk1280 test.nim
```

chirag-parmar:
Also want to point out that the timing improvements for encoding blocks (not hashing) using the new implementation is way more significant than for hashing with new implementations. This is because the keccak implementation adds a huge bias. The next step to improving the rlp hash algo should be optimizing the keccak implementation before optimizing the writers again.

tests/rlp/test_rlp_profiler.nim
```diff
@@ -0,0 +1,125 @@
...
+      let bytes11 = rlp.encodeHash(nonEmptyBlock)
+    benchmark "Block header hashing using hash writer":
+      let bytes12 = rlp.encodeHash(h)
+
+  test "encoding and hashin using default writer":
+    benchmark "Transaction hashing using default writer and then hash":
+      let bytes13 = encodeAndHash(myTx)
+    benchmark "Block Sequence hashing using default writer and then hash":
+      let bytes14 = encodeAndHash(blkSeq)
+    benchmark "Block hashing using defualt writer and the hash":
```
jangko, line 122: defualt -> default

tests/rlp/test_rlp_profiler.nim
```diff
@@ -0,0 +1,125 @@
...
+      let bytes15 = encodeAndHash(blkSeq)
+    benchmark "Block header hashing using hash writer":
```
jangko, line 124: Block header hashing using hash writer -> Block header hashing using default writer and the hash

tersec:
As a general note, `refc` performance is significantly more important to Nimbus and to IFT in general than ORC for the foreseeable future.

It's not to benchmark both, but the operative benchmarks in a short/medium-term way are refc ones.

chirag-parmar:
here are refc benchmarks:

| type of benchmark        | Time taken by default implementation | Time taken by new implementation |
|------------------------|---------------------------------------|------------------------------------|
| Tx Encoding (single)     | 17us                                                          | 14us                                                    |
| Tx Hashing (single)       | 30us                                                         | 40us                                                   |
| Header Encoding          | 22us                                                         | 15us                                                    |
| Header Hashing            | 34us                                                         | 36us                                                   |
| Block Encoding             | 45us                                                         | 25us                                                   |
| Block Hashing               | 72us                                                         |  63us                                                  |

chirag-parmar:
mem usage
cpu time

tersec:
No copyright linter will enforce this in this repository, but still good to check all the modified files in the PR and update their copyright years to 2025.

arnetheduck:
something to note about benchmarking garbage-collected languages: the cost of allocation and deallocation is often not charged on the spot, but happens "randomly" - specially in a real-world application where you can't really micro-benchmark it in isolation (due to fragmentation etc).

Instead of benchmarking, the other way to reason about a change such as this is to measure/reason about the number of allocations that happen during an operation because this will ultimately determine how RLP contributes to the bigger picture in an application - the aim of the optimization is to reduce this number while not compromising on other things - for the _ideal_ case where you're hashing an `object` (ie a call to `rlpHash`), _all_ allocations can be done on the stack / precomputed - if this PR hits this target, it is already a success (dynamic usage of RLP is a separate topic) - the _remaining_ unnecessary overhead will be found in zero:ing the (now) stack-allocated memory, but that's a separate concern as well.

Another way to look at it is that the objective of this PR can be reduced to "`rlpHash(myobject)` must not perform any heap memory alllocations" - once it fulfills this objective, it is done and we can merge / move on.

chirag-parmar:
> for the _ideal_ case where you're hashing an `object` (ie a call to `rlpHash`), _all_ allocations can be done on the stack / precomputed - if this PR hits this target, it is already a success (dynamic usage of RLP is a separate topic) - the _remaining_ unnecessary overhead will be found in zero:ing the (now) stack-allocated memory, but that's a separate concern as well.

Noted

> Another way to look at it is that the objective of this PR can be reduced to "`rlpHash(myobject)` must not perform any heap memory alllocations" - once it fulfills this objective, it is done and we can merge / move on.

Makes sense, there are two edge cases where dynamic allocation is unavoidable. First, are lists(static or dynamic) within dynamic sized lists. In the RLP world this would include sequences/arrays within dynamic sized sequences and objects within dynamic sized sequences. An example is the transactions field within the block object.

The second is where the type of an object doesn't fully assume the shape of the object at compile time. Examples are recurring objects for binary trees and parsed data types like JSON.

Apart from this all other cases achieve static memory allocation

chirag-parmar:
https://github.com/status-im/nimbus-eth1/actions/runs/12814215402
https://github.com/status-im/nimbus-eth1/actions/runs/12814215403
https://github.com/status-im/nimbus-eth1/actions/runs/12814215405

tested against nimbus-eth1

jangko:
For `rlpHash` we can remove all the allocations. But then we need to recalculate length prefix for each nested object/list. This approach is good for an innermost object or  an object without nested object or list. But for complex object like a `Block` or a `Transaction`, recalculating length prefix will undo the optimization. It will perform worse.

Second approach, and this has been tested is replacing `seq` with `c_malloc/c_free` when allocating length prefix buffer. It will improve performance. But that can be done in separate PR.

I prefer to merge this first, and use it in nimbus-eth1 because we have achieve improvements compared to older implementation.

If no more objections, I will merge this soon.

eth/rlp/length_writer.nim
```diff
@@ -0,0 +1,110 @@
...
+    if self.pendingLists[self.listTop - 1].remainingItems == 0:
+      let 
+        listIdx = self.pendingLists[self.listTop - 1].idx
+        startLen = self.pendingLists[self.listTop - 1].startLen
+        listLen = self.totalLength - startLen
+        prefixLen = if listLen < int(THRESHOLD_LIST_LEN): 1
+                      else: int(uint64(listLen).bytesNeeded) + 1
+
+      # save the list lengths and prefix lengths
+      self.lengths[listIdx] = (listLen, prefixLen)
```
arnetheduck, line 49: since `prefixLen` is trivial to compute from `listLen`, it seems unnecessary to put it in lengths, no?

eth/rlp/length_writer.nim
```diff
@@ -0,0 +1,110 @@
...
+    self.totalLength += 1
+    self.maybeClosePendingLists()
+  else:
+    # open a list
+    when self is DynamicRlpLengthTracker:
+      self.pendingLists.setLen(self.listTop + 1)
+    self.pendingLists[self.listTop] = (self.listCount, listSize, self.totalLength)
+    self.listTop += 1
+    self.listCount += 1
+    if self.listCount == self.lengths.len:
```
arnetheduck, line 75: why this odd growth strategy? what's wrong with `newSeqOfCap`?

eth/rlp/default_writer.nim
```diff
@@ -0,0 +1,113 @@
...
+# Copyright (c) 2019-2025 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+import
+  std/options,
+  pkg/results,
+  stew/[arraybuf, assign2, shims/macros],
```
arnetheduck, line 11: lots of these imports are not actually used

eth/rlp/length_writer.nim
chirag-parmar: Wanted to skip on the branching condition for recalculating the `prefixLen`.

eth/rlp/length_writer.nim
chirag-parmar: while benchmarking the implementation I had found that the `add` and `setLen` operations consumed way too many cycles when the memory was uninitialised. So I initialized the sequence with zeroes using `newSeq`. I decided to grow the sequence in bursts because 50 `setLen(len + 1)` increments is slower than `setLen(len + 50)`

eth/rlp/default_writer.nim
chirag-parmar: you are right, apologies. Will remove them ASAP with a new PR.

eth/rlp/length_writer.nim
chirag-parmar: https://github.com/status-im/nim-eth/blob/rlp-opt-2/eth/rlp/vla.nim this is @jangko 's implementation

arnetheduck: > So I initialized the sequence with zeroes using newSeq

what you're looking for is `newSeqOfCap(128)` _usually_ or possibly `newSeqUninitialized` to skip the zeroing (though when adding items one by one, the zeroing doesn't matter, this is only for large seqs of bytes) - `seq` grows itself in almost exactly the same way you've implemented it here - ie if you were to look at the generated assembly, they would be very similar.

`listCount` for example is the exact same field as `len` in the seq, if you think about it, and the seq capacity (the total space available before you have to reallocate) is what in your code is represented by len.

The cost is code complexity mainly, because at the end of the day, what you're doing manually here almost exactly matches what the seq itself is doing internally if you give it an initial capacity.

> [vla.nim](https://github.com/status-im/nim-eth/blob/rlp-opt-2/eth/rlp/vla.nim)

this is probably a bit over the top for the complexity that it brings in having to manually manage memory - what _usually_ works well in situations like this is a "small seq" optimization where the first 16:ish items are held in a stack-based array so that all the little cases don't require any allocation at all, ie https://live.boost.org/doc/libs/1_85_0/doc/html/boost/container/small_vector.html is the classic implementation of this technique. I think once we move to orc we might add a smallseq implementation like that to stew and then it can be used here, without the risk of mem leaks etc.

eth/rlp/length_writer.nim
arnetheduck: I'd be impressed if you were able to measure that ;) it does make the code a bit more complex however, which in and of itself can outweigh whatever benefits you might find - it's a good thought, but it doesn't always practically add up.

eth/rlp/default_writer.nim
arnetheduck: while you're at it, I'd recommend enabling "strip whitespace at end of line" in your editor (or just use `nph`)

eth/rlp/length_writer.nim
chirag-parmar: makes sense

eth/rlp/length_writer.nim
chirag-parmar: Ok so I tried running the benchmarks again but couldn't reproduce them. So I ran new benchmarks and you are right. I am not sure why at the time, `add` and `setLen` were taking longer. I'll change this, thank you for the review.

PR #767, opened 2025-01-06
add discv5_network_byte metric in discoveryv5 protocol

Ivansete-status:
This PR adds a simple metric that helps measuring the in/out traffic
See, for example [this](https://www.notion.so/Measure-DiscV5-bandwidth-with-Waku-discovery-1698f96fb65c80659fa1fbfdac49b1ef?d=16f8f96fb65c8067adb2001c2a329eca#16a8f96fb65c800ca96fe41f60052693)

### DIFF
--- a/eth/p2p/discoveryv5/protocol.nim
+++ b/eth/p2p/discoveryv5/protocol.nim
@@ -171,6 +171,10 @@ type
 
   DiscResult*[T] = Result[T, cstring]
 
+  Direction {.pure.} = enum
+    In = "in"
+    Out = "out"
+
 const
   defaultDiscoveryConfig* = DiscoveryConfig(
     tableIpLimits: DefaultTableIpLimits,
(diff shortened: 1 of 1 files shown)


eth/p2p/discoveryv5/protocol.nim
```diff
@@ -276,6 +277,8 @@ proc sendTo(d: Protocol, a: Address, data: seq[byte]): Future[void] {.async.} =
     # because of ping failures due to own network connection failure.
     warn "Discovery send failed", msg = e.msg, address = $ta
 
+  discv5_network_bytes.inc(data.len.int64, labelValues = ["out"])
```
arnetheduck, line 280: ideally we'd use an enum here (`In = "in"` etc) which should avoid some allocations in the metrics library

arnetheduck (approved):

PR #768, opened 2025-01-20
Discv5 Routing Table: Add support for banning nodes

bhartnett:
This PR add support for banning nodes from the Discv5 routing table. This feature will be used by both Discv5 and the portal subnetworks in Fluffy. See the related task here: https://github.com/status-im/nimbus-eth1/issues/2809

- Nodes are banned by id so they can be banned even if not yet present in the routing table. 
- Once banned, nodes cannot be added to the routing table. 
- When banned, nodes are removed from the routing table and replaced using the replacement cache.

The updates to the Discv5 protocol to implement banning of peers will be included in a separate PR.

### DIFF
--- a/eth/p2p/discoveryv5/routing_table.nim
+++ b/eth/p2p/discoveryv5/routing_table.nim
@@ -197,37 +197,49 @@ func ipLimitDec(r: var RoutingTable, b: KBucket, n: Node) =
 func getNode*(r: RoutingTable, id: NodeId): Opt[Node]
 proc replaceNode*(r: var RoutingTable, n: Node)
 
+proc cleanupExpiredBans(r: var RoutingTable) =
+  let currentTime = now(chronos.Moment)
+
+  var expiredIds = newSeq[NodeId]()
+  for id, timeout in r.bannedNodes:
+    if currentTime >= timeout:
+      expiredIds.add(id)
+
+  for id in expiredIds:
+    r.bannedNodes.del(id)
+
 proc banNode*(r: var RoutingTable, nodeId: NodeId, period: chronos.Duration) =
   ## Ban a node from the routing table for the given period. The node is removed
   ## from the routing table and replaced using a node from the replacement cache.
-
   let banTimeout = now(chronos.Moment) + period
 
   if r.bannedNodes.contains(nodeId):
     let existingTimeout = r.bannedNodes.getOrDefault(nodeId)
     if existingTimeout < banTimeout:
       r.bannedNodes[nodeId] = banTimeout
-  else:
-    r.bannedNodes[nodeId] = banTimeout
+    return # node is already banned so we don't need to try replacing it because
+    # it should have already been replaced when it was initially banned
+
+  # NodeId doesn't yet exist in the banned nodes table
+  r.bannedNodes[nodeId] = banTimeout
 
   # Remove the node from the routing table
   let node = r.getNode(nodeId)
   if node.isSome():
     r.replaceNode(node.get())
 
-proc isBanned*(r: var RoutingTable, nodeId: NodeId): bool =
+  # Remove all expired bans from the banned nodes table
+  r.cleanupExpiredBans()
+
+proc isBanned*(r: RoutingTable, nodeId: NodeId): bool =
   if not r.bannedNodes.contains(nodeId):
     return false
 
   let
     currentTime = now(chronos.Moment)
     banTimeout = r.bannedNodes.getOrDefault(nodeId)
-  if currentTime < banTimeout:
-    return true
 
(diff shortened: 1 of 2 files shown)


bhartnett:
@kdeme FYI, I didn't implement any filter in the `neighbors` function because we are now removing banned nodes from the routing table so they shouldn't need to be filtered out.

eth/p2p/discoveryv5/routing_table.nim
```diff
@@ -189,6 +194,41 @@ func ipLimitDec(r: var RoutingTable, b: KBucket, n: Node) =
...
+    let existingTimeout = r.bannedNodes.getOrDefault(nodeId)
+    if existingTimeout < banTimeout:
+      r.bannedNodes[nodeId] = banTimeout
+  else:
+    r.bannedNodes[nodeId] = banTimeout
+
+  # Remove the node from the routing table
+  let node = r.getNode(nodeId)
+  if node.isSome():
+    r.replaceNode(node.get())
```
bhartnett, line 216: Should we replace the node as I've done here or simply remove it without using the replacement cache?

kdeme: Replace is the way to go :+1: 

In general we never really use remove AFAIK (maybe in testing?), always use the replacement cache as that cache is not actively used to search for nodes.

tests/p2p/test_routing_table.nim
```diff
@@ -561,3 +562,71 @@ suite "Routing Table Tests":
...
+      node1 = generateNode(PrivateKey.random(rng[]))
+      node2 = generateNode(PrivateKey.random(rng[]))
+
+    var table = RoutingTable.init(localNode, 1, DefaultTableIpLimits, rng = rng)
+
+    check table.addNode(node1) == Added
+    table.banNode(node1.id, 1.nanoseconds)
+    table.banNode(node2.id, 1.nanoseconds)
+
+    sleep(100)
```
kdeme, line 618: This is 100 ms right? can be just value of 1 then?

eth/p2p/discoveryv5/routing_table.nim
```diff
@@ -189,6 +194,41 @@ func ipLimitDec(r: var RoutingTable, b: KBucket, n: Node) =
...
+    return false
+
+  let
+    currentTime = now(chronos.Moment)
+    banTimeout = r.bannedNodes.getOrDefault(nodeId)
+  if currentTime < banTimeout:
+    return true
+
+  # Peer is in bannedNodes table but the time period has expired
+  r.bannedNodes.del(nodeId)
```
kdeme, line 229: I wonder if this is sufficient for banned nodes clean-up. 

A node could be started, send 1 invalid message and then disappear from the network? Repeat this continuously and there is a small but increasing mem leak?

eth/p2p/discoveryv5/routing_table.nim
bhartnett: Ok great, I'll leave it as it is then.

tests/p2p/test_routing_table.nim
bhartnett: Yes true, I'll update. I guess it will speed up the test.

bhartnett:
@kdeme I've updated this PR based on your suggestions. Ready for another review.

eth/p2p/discoveryv5/routing_table.nim
```diff
@@ -189,6 +194,53 @@ func ipLimitDec(r: var RoutingTable, b: KBucket, n: Node) =
...
+  # NodeId doesn't yet exist in the banned nodes table
+  r.bannedNodes[nodeId] = banTimeout
+
+  # Remove the node from the routing table
+  let node = r.getNode(nodeId)
+  if node.isSome():
+    r.replaceNode(node.get())
+
+  # Remove all expired bans from the banned nodes table
+  r.cleanupExpiredBans()
```
kdeme, line 232: I think having the cleanup here is potentially worse than the actual mem leak.

Allowing this to run in the `banNode` call effectively puts it in an attacker's power to run this cleanup. While at the same time the attacker can make it more expensive by "adding bans". 

I think that this type of clean-up call should always be run independently of any peer node actions.

bhartnett: I see so the concern in this case is the computational overhead of looping over the table rather than the list growing?

Did you have a suggestion for where you think the cleanup should happen?

bhartnett: I've now removed the call to `cleanupExpiredBans` from the `banPeer` function and made it public so it can be called at a high layer.

kdeme (approved):

PR #776, opened 2025-02-24
Remove unused `deadlineVar` from p2p dsl

etan-status:
The `deadlineVar` is not used anywhere and can be removed.


jangko:
included in #777

PR #784, opened 2025-03-26
NibblesSeq -> NibblesBuf

arnetheduck:
Switch nibbles implementation to the array-based impl from nimbus-eth1 which overall is faster (though it only supports up to 64 nibbles, unlike the current one) - a collateral benefit is that this speeds up the hexary trie by quite a bit.


arnetheduck:
Build fails probably related to https://github.com/nim-lang/Nim/pull/24815

ringabout:
I think it works by chance (i.e. a bug)

```nim
{.emit:"""
void foo(unsigned long long* x)
{
}
""".}


proc foo(x: var culonglong) {.importc: "foo", nodecl.}

proc main(x: var uint64) =
  # var s: culonglong = u
  var m = uint64(12)
  foo(culonglong(m))

var u = uint64(12)
main(u)
```

Conversions from `var uint64` to `culonglong` were kept, but not `uint64` to `culonglong` because source types didn't ignore `tyVar`, which is irrelevant to backends. Now I make it picky about caliases in this PR => https://github.com/nim-lang/Nim/pull/24818

arnetheduck:
Error is actually in stint, whose important_packages test is too shallow to catch it

Issue #724, opened 2024-08-30
Implement incremental hasher for computing txroot/etc without constructing database


arnetheduck:
When computing MPT roots, the current version of the code works by constructing a database of the full value and then hashing up to the root - this is extremely inefficient due to the many allocations involved as well as the cost of maintaining the full data in memory - instead, it's possible to build an incremental hasher that:

* only stores hashes, not values
* every time a branch is full (ie has 16 children) collapses those into a single hash

Examples of this can be found:

* https://github.com/status-im/nim-ssz-serialization/blame/b5aa78cae197cc97277be050ba69ef9c5d33aad1/ssz_serialization/merkleization.nim#L42
* https://github.com/alloy-rs/trie/tree/main/src/hash_builder

Also related is https://github.com/status-im/nim-eth/issues/698

arnetheduck:
Implemented partially, but needs to be updated to take advantage of the new direct rlp hasher: https://github.com/status-im/nim-eth/blob/89d7310bb756ae60c3a448de23ab1156bad5df47/eth/trie/ordered_trie.nim#L157

PR #787, opened 2025-04-10
Use hash writer for transaction signing and ordered tries

chirag-parmar:
* closes [724](https://github.com/status-im/nim-eth/issues/724)
* addresses [this](https://github.com/status-im/nimbus-eth1/issues/3164#issuecomment-2771998449)


eth/common/transactions_rlp.nim
```diff
@@ -128,117 +202,34 @@ proc append*(w: var RlpWriter, tx: PooledTransaction) =
...
-  w.append(tx.maxPriorityFeePerGas)
-  w.append(tx.maxFeePerGas)
-  w.append(tx.gasLimit)
-  w.append(tx.to)
-  w.append(tx.value)
-  w.append(tx.payload)
-  w.append(tx.accessList)
-  w.append(tx.authorizationList)
-  w.finish()
+func toUnsignedTransaction(tx: Transaction, eip155: bool): UnsignedTransaction =
```
jangko, line 205: Isn't make a copy of `UnsignedTransaction` is more expensive than using existing `encodeForSigning`?

eth/rlp/hash_writer.nim
```diff
@@ -10,7 +10,9 @@ import nimcrypto/keccak, ./priv/defs, utils, ../common/hashes, length_writer
 type RlpHashWriter* = object
   keccak: keccak.keccak256
   lengths*: seq[int]
+  wrapLengths*: seq[int]
```
jangko, line 13: Is the perf gain of using `wrapLengths` outperform the previous algorithm?

chirag-parmar: Until we don't assign memory to hashing RLP encodings we should be good, right?

chirag-parmar: I can make measurements though. Let me get back to you shortly

chirag-parmar: @jangko i made a few measurements initially with the `refc` GC and i wasn't getting any reliable results with the garbage collector off `-mm:none` this implementation actually performs worse

chirag-parmar:
```

proc append*(w: var RlpWriter, txs: MyType) =
  for t in txs.txs:
    w.append(rlp.encode(t))

proc append*(w: var RlpWriter, txs: MySecondType) =
  for t in txs.txs:
    w.wrapEncoding(1)
    w.append(t)

let values2 = block:
  var tmp: MySecondType
  tmp.txs = newSeqOfCap[Transaction](110)
  for i in 0 .. 10:
    tmp.txs.add(tx0(i))
    tmp.txs.add(tx1(i))
    tmp.txs.add(tx2(i))
    tmp.txs.add(tx3(i))
    tmp.txs.add(tx4(i))
    tmp.txs.add(tx5(i))
    tmp.txs.add(tx6(i))
    tmp.txs.add(tx7(i))
    tmp.txs.add(tx8(i))
    tmp.txs.add(txEip7702(i))
    tmp.txs.add(txEip7873(i))
  tmp

let values1 = block:
  var tmp: MyType
  tmp.txs = newSeqOfCap[Transaction](110)
  for i in 0 .. 10:
    tmp.txs.add(tx0(i))
    tmp.txs.add(tx1(i))
    tmp.txs.add(tx2(i))
    tmp.txs.add(tx3(i))
    tmp.txs.add(tx4(i))
    tmp.txs.add(tx5(i))
    tmp.txs.add(tx6(i))
    tmp.txs.add(tx7(i))
    tmp.txs.add(tx8(i))
    tmp.txs.add(txEip7702(i))
    tmp.txs.add(txEip7873(i))
  tmp

# this is done to load the data into the RAM and not account for it in the benchmark
let hash2 = rlp.computeRlpHash(values2)
let hash1 = rlp.computeRlpHash(values1)

benchmark("Measure wrapEncoding implementation"):
  let hash3 = rlp.computeRlpHash(values2)

benchmark("Measure manual nested encoding"):
  let hash5 = rlp.computeRlpHash(values1)

```

chirag-parmar: @jangko the implementation is now better performing for `orderedTrieRoot` too

eth/rlp/hash_writer.nim
```diff
@@ -82,12 +82,17 @@ func initHashWriter*(tracker: var RlpLengthTracker): RlpHashWriter =
...
+func reInit*(self: var RlpHashWriter, tracker: var RlpLengthTracker) =
+  self.lengths = move(tracker.lengths)
+  self.wrapLengths = move(tracker.wrapLengths)
+
 template finish*(self: var RlpHashWriter): Hash32 =
-  self.lengths.setLen(0)
-  self.wrapLengths.setLen(0)
   self.keccak.finish.to(Hash32)
 
 func clear*(self: var RlpHashWriter) =
```
jangko, line 92: Please measure using compiler builtin `reset`. e.g. `self.reset()`

eth/rlp/two_pass_writer.nim
```diff
@@ -85,13 +85,19 @@ func initTwoPassWriter*(tracker: var RlpLengthTracker): RlpTwoPassWriter =
...
+  self.output.setLen(tracker.totalLength)
+  self.lengths = move(tracker.lengths)
+  self.wrapLengths = move(tracker.wrapLengths)
+
 template finish*(self: RlpTwoPassWriter): seq[byte] =
-  self.lengths.setLen(0)
-  self.wrapLengths.setLen(0)
   self.output
 
 func clear*(self: var RlpTwoPassWriter) =
```
jangko, line 96: ditto reset

eth/common/transactions_rlp.nim
arnetheduck: possible inheritance could be used, ue `type SignedTransaction = object of UnsignedTransaction...`

chirag-parmar: I tried using inheritance before but decided against it. Particularly because for `EIP155` handling the passed transaction will have to mutable. The only benefit of using inheritance would be type casting, I believe.

eth/rlp/hash_writer.nim
chirag-parmar: no difference between using `self.reset` or manually resetting each buffer

eth/rlp/two_pass_writer.nim
chirag-parmar: see comment above

eth/rlp/hash_writer.nim
jangko: probably the compiler expand the `reset` to something similar to manual reset instead of resetting them in one go.

jangko (approved):

eth/rlp/hash_writer.nim
arnetheduck: `reset` releases the payload memory of each seq in the writer - `setLen(0)` keeps it around for reuse (ie it can grow to its previous capacity).

In general, there's no point creating manually written `clear` functions that just call `reset` (except for backwards compat) - if instead they do setLen(0) they can help avoid some memory allocations if the same instance is reused.

chirag-parmar: noted. rolledback to `setLen(0)`

chirag-parmar:
@jangko shall we merge this once the pipeline is green?

jangko:
I have no problem with that

Issue #262, opened 2020-06-30
Better handling of stale nodes


kdeme:
After some tests in https://github.com/status-im/nim-eth/pull/261, it was decided to still replace nodes on first failure.

One scenario where this is troublesome is in case the own network connection drops, the routing table will get cleared.
Other is possibly when the own network connection is just very slow and/or flakey. 
And as last, perhaps for usage with much less lookups, one might want a better control of the "stale" nodes.

One simple way to handle this is to flag a node on first failure not to be used for lookups/revalidation any more. And to do in the background (e.g. by some callback after first failure) some exponential back off pinging or just random pinging. (e.g. 5 times).

Should be tested properly though with looking at metrics, e.g. in bad network conditions.

PR #791, opened 2025-05-06
p2p RoutingTable: Fix replaceNode so that the node will not be removed if the replacement cache is empty.

bhartnett:


eth/p2p/discoveryv5/routing_table.nim
```diff
@@ -458,20 +460,35 @@ proc removeNode*(r: var RoutingTable, n: Node) =
...
-  # TODO: Kademlia paper recommends here to not remove nodes if there are no
+  ## no entries in the replacement cache, node `n` will simply be marked as
+  ## not seen. Returns bool indictating whether or not the node was successfully
+  ## replaced.
+
+  # Note: Kademlia paper recommends here to not remove nodes if there are no
   # replacements. However, that would require a bit more complexity in the
   # revalidation as you don't want to try pinging that node all the time.
+
+  var replaced = false
```
kdeme, line 473: nit: I find it slightly cleaner to just return the resulting bool directly in each if/else clause

tests/p2p/test_routing_table.nim
```diff
@@ -162,8 +162,8 @@ suite "Routing Table Tests":
       check table.addNode(n) == Added
 
     table.replaceNode(table.nodeToRevalidate())
-    # This node should still be removed
-    check (table.getNode(bucketNodes[bucketNodes.high].id)).isNone()
+    # This node should not be removed
+    check (table.getNode(bucketNodes[bucketNodes.high].id)).isSome()
```
kdeme, line 166: could add a check on the n.seen = false here.

tests/p2p/test_routing_table.nim
```diff
@@ -191,7 +191,7 @@ suite "Routing Table Tests":
     check:
       res.isSome()
       res.get() == doubleNode
-      table.len == 1
+      table.len == 16
```
kdeme, line 194: There was a part of this test that checks if order of revalidations is preserved by checking at the end if the latest added node (doubleNode) is still there (the one not revalidated and thus replaced).

As now all nodes remain, this `getNode` will for sure always pass.

I think same test can be achieved by checking the `seen` value for each node.

tests/p2p/test_routing_table.nim
```diff
@@ -237,7 +237,7 @@ suite "Routing Table Tests":
 
     for n in bucketNodes:
       table.replaceNode(table.nodeToRevalidate())
-      check (table.getNode(n.id)).isNone()
+      check (table.getNode(n.id)).isSome()
```
kdeme, line 240: similar item as above, the `getNode` is no longer a good test. I think we need to check `seen` value also.

kdeme (commented):
There is longstanding issue related to this: https://github.com/status-im/nim-eth/issues/262

That issue also refers to some testing done on early testnets: https://github.com/status-im/nim-eth/pull/261#issuecomment-651653287

Basically the gist of it is/was that the re-pinging of stale nodes was causing less quick enabling of new (unseen / unverified) nodes. 
This for sure had to some part to due with the state of those early networks.

It is possibly the case now that the quick removal of nodes on failure might work morecounter productive due to having to re-add / ping eventually the same or other nodes. This does depend somewhat on how well filled the replacement caches are typically.

Now with this PR, stale nodes that are in close to the local node buckets, are unlikely to be ever removed (yet will be re-pinged). This is going to be a very small subset of nodes of the routing table however, so it probably does not outweigh the benefit? Difficult to really know though without some metrics.

bhartnett:
@kdeme This is the test that is failing which I'm not sure how is best to fix it:
 /home/runner/work/nim-eth/nim-eth/tests/p2p/test_discoveryv5.nim
 /home/runner/work/nim-eth/nim-eth/build/p2p/all_tests 'Discovery v5.1 Tests::Resolve target'

See failure in CI here: https://github.com/status-im/nim-eth/pull/791/checks#step:11:555

eth/p2p/discoveryv5/routing_table.nim
bhartnett: Sure, will update.

tests/p2p/test_routing_table.nim
bhartnett: Good point, will do.

bhartnett:
> There is longstanding issue related to this: #262
> 
> That issue also refers to some testing done on early testnets: [#261 (comment)](https://github.com/status-im/nim-eth/pull/261#issuecomment-651653287)

Thanks. Good to have the history linked here.

> Basically the gist of it is/was that the re-pinging of stale nodes was causing less quick enabling of new (unseen / unverified) nodes. This for sure had to some part to due with the state of those early networks.
> 
> It is possibly the case now that the quick removal of nodes on failure might work morecounter productive due to having to re-add / ping eventually the same or other nodes. This does depend somewhat on how well filled the replacement caches are typically.

I would think at the Discv5 level the network is very large so the replacement caches would likely be very full so perhaps this change wouldn't have much impact on Discv5 network. In portal the replacement caches are likely smaller.

> Now with this PR, stale nodes that are in close to the local node buckets, are unlikely to be ever removed (yet will be re-pinged). This is going to be a very small subset of nodes of the routing table however, so it probably does not outweigh the benefit? Difficult to really know though without some metrics.

Well its a small number of nodes and the ping process picks a bucket at random so its likely a small overhead. I would say in the short term this is probably acceptable overhead but in the longer term we should use peer scoring in combination with node banning to remove misbehaving nodes from the routing table which would mitigate this issue you describe.

bhartnett:
> @kdeme This is the test that is failing which I'm not sure how is best to fix it: /home/runner/work/nim-eth/nim-eth/tests/p2p/test_discoveryv5.nim /home/runner/work/nim-eth/nim-eth/build/p2p/all_tests 'Discovery v5.1 Tests::Resolve target'
> 
> See failure in CI here: https://github.com/status-im/nim-eth/pull/791/checks#step:11:555

I have now fixed this test in my last commit.

PR #792, opened 2025-05-12
Add a predicate to filter nodes in routing table neighbours function

bhartnett:
This PR adds a new filter predicate to the `neighbours` function in the routing table. 

This is useful for applying additional filtering criteria while still getting the correct number of k nodes in the result. The new parameter is optional and if not set, all nodes are returned as before.


kdeme (commented):
I assume you have a specific use case in mind for this?

kdeme (approved):

PR #794, opened 2025-05-13
Drop support for logsBloom

advaita-saha:


eth/common/receipts_rlp.nim
```diff
@@ -13,21 +13,26 @@ from stew/objects import checkedEnumAssign
...
-proc append*(w: var RlpWriter, rec: Receipt) =
+proc append*(w: var RlpWriter, rec: SomeReceipt) =
   if rec.receiptType in {Eip2930Receipt, Eip1559Receipt, Eip4844Receipt, Eip7702Receipt, Eip7873Receipt}:
     w.append(rec.receiptType.uint)
 
-  w.startList(4)
+  when rec.type is StoredReceipt:
+    w.startList(3)
+  else:
+    w.startList(4)
```
tersec, line 23: Currently there are exactly two possible types of `SomeReceipt`, but (in theory) this might change, and even if not, locally in the code it's not obvious, so one approach is to use something like:
```nim
  when rec.type is StoredReceipt:
    w.startList(3)
  elif rec.type is Receipt:
    w.startList(4)
  else:
    static: doAssert false
```

Other things besides `doAssert false` work -- using an `{.error.}` pragma, using `raiseAssert`, etc. The important part is that it fail at compile time if the type isn't what it's expecting.

eth/common/receipts.nim
```diff
@@ -56,3 +68,30 @@ func hasStateRoot*(rec: Receipt): bool {.inline.} =
 func stateRoot*(rec: Receipt): Hash32 {.inline.} =
   doAssert(rec.hasStateRoot)
   rec.hash
+
+func logsBloom(logs: openArray[Log]): BloomFilter =
+  for log in logs:
+    result.incl log.address
```
tersec, line 74: https://status-im.github.io/nim-style-guide/language.result.html still applies -- the usual approach here is something like
```nim
  var res: BloomFilter
  ...
    res.incl log.address
  ...
  res
```
Sometimes this truly isn't very efficient because of Nim (N)RVO weaknesses, but we tend to regard those as bugs and try to get those fixed, and for `Uint2048` it shouldn't be that problematic.

eth/common/receipts.nim
```diff
@@ -56,3 +68,30 @@ func hasStateRoot*(rec: Receipt): bool {.inline.} =
...
+  StoredReceipt(
+    receiptType       : rec.receiptType,
+    isHash            : rec.isHash,
+    status            : rec.status,
+    hash              : rec.hash,
+    cumulativeGasUsed : rec.cumulativeGasUsed,
+    logs              : rec.logs
+  )
+
+func to*(rec: StoredReceipt, _: Receipt): Receipt =
```
tersec, line 88: Roundtrip tests for these two `to` functions?

jangko:
I think `StoredReceipt` should use the same encoding for `eth/69`.

> * (eth/69): receipt = [tx-type, post-state-or-status, cumulative-gas, logs]      # <------ this encoding

This improves reusability instead of eth/69 and storage use their own encoding. Also simplify or avoid conversion between types used in eth/69 and storage.

spec: https://eips.ethereum.org/EIPS/eip-7642#receipts-message-changes
example: https://github.com/status-im/nimbus-eth1/blob/3099fd6cea2ffa3970ea806426bbffe78884f69a/execution_chain/sync/wire_protocol/receipt69.nim

advaita-saha:
> This improves reusability instead of eth/69 and storage use their own encoding. Also simplify or avoid conversion between types used in eth/69 and storage.

This was my intention all together, seems like I just missed the format difference in rlp. Thanks for pointing it out @jangko

eth/common/receipts_rlp.nim
advaita-saha: Not relevant anymore, encoding format changed, thus needing separate functions

jangko:
My local test with eth/69 looks good. We only need to wait for https://github.com/status-im/nimbus-eth1/pull/3283 become green and then this PR can be merged.

advaita-saha:
> become green and then this PR can be merged.

Seems good to merge now, as CI is green

PR #795, opened 2025-05-14
Revert PR #787 "Use hash writer for transaction signing and ordered tries"

bhartnett:
Revert "Use hash writer for transaction signing and ordered tries"

This reverts PR https://github.com/status-im/nim-eth/pull/787 from commit 6002a29bfef28dfc9dea66f7b7c14e4374d1c038.

This change breaks both Fluffy and Nimbus-Eth1 tests. This change should be tested against Nimbus-Eth1 and perhaps even Nimbus-Eth2 before putting the change back in.


bhartnett:
FYI @chirag-parmar

PR #803, opened 2025-06-19
dcli: Store all discovered peers, not just eth2 peers

kdeme:
- Store all nodes
- add EL eth field and CL syncnets
- General cleanup and small improvements
- adjust bootstrap-node cli flag to be consistent with nimbus_portal_client and nimbus_beacon_node


kdeme:
Letting the crawler run for a bit, here are some plotted results of nodes with announced `fork_hash` (EL):

fork_hash_plot

PR #806, opened 2025-07-22
common: Change authList v field to yParity in Transaction type

kdeme:
EIP-7702 defines it as y_parity. This has no actual effect on rlp encoding/decoding. But the type is also used for JSON encoding in execution API. There it is specifically defined as yParity and thus for JSON-RPC requests it would not encode/decode it properly (= always 0 v value without noticable error message)

See execution API specs: https://github.com/ethereum/execution-apis/blob/d41fdf10fabbb73c4d126fb41809785d830acace/src/schemas/transaction.yaml#L75

And also in the EIP it is called as named: https://eips.ethereum.org/EIPS/eip-7702#set-code-transaction

Note that also the v value of the Transaction is called there as such, and for the JSON-RPC values it sets apparently both values for backwards compatibility: https://github.com/ethereum/execution-apis/blob/d41fdf10fabbb73c4d126fb41809785d830acace/src/schemas/transaction.yaml#L343-L350

I did not verify here if this is all correct on nimbus EL.

### DIFF
--- a/eth/common/transactions.nim
+++ b/eth/common/transactions.nim
@@ -24,7 +24,7 @@ type
     chainId*: ChainId
     address*: Address
     nonce*: AccountNonce
-    yParity*: uint64
+    yParity*: uint8
     r*: UInt256
     s*: UInt256
 


arnetheduck:
Sounds a bit like https://github.com/status-im/nim-web3/pull/209?

jangko:
it first implemented as yParity https://github.com/status-im/nim-eth/pull/749 :)
so no objection

kdeme:
> it first implemented as yParity #749 :) so no objection

ok thanks. Note that I've also changed it now to be a uint8: https://github.com/status-im/nim-eth/pull/806/commits/293731ddf9ceccdd62c1780120e3b4c5430d30c4

This appears to work according to CI in nimbus-eth1: https://github.com/status-im/nimbus-eth1/pull/3497
It does still trigger some issue in nimbus-eth2: https://github.com/status-im/nimbus-eth2/pull/7316

kdeme:
> edit: The fail on nimbus-eth2 has just to do with the `JrpcConv.decode` of `TransactionObject`. This is fixed by adding `readValue` for `uint8` in web3, like there is for `uint64`. Or we could just fully apply [status-im/nim-web3#209](https://github.com/status-im/nim-web3/pull/209) and add the `U8Quantity`.
> 
> Not sure if the latter is really still that useful when currently you need to explicitly define the JrpcConv.decode. Does this comment still apply to this date? https://github.com/status-im/nim-web3/blob/master/web3/primitives.nim#L42-L44

@jangko any preference on this?

jangko:
This is a good reason to push https://github.com/status-im/nim-json-serialization/issues/121

PR #809, opened 2025-07-31
update error handling

gerceboss:
- [x] update error handling for receive function
- [x] add constructor for common transport

Question:
Should I add open, close functions for shared port ?

### DIFF
--- a/eth/p2p/discoveryv5/protocol.nim
+++ b/eth/p2p/discoveryv5/protocol.nim
@@ -467,83 +467,91 @@ proc banNode*(d: Protocol, n: Node, banPeriod: chronos.Duration) =
 proc isBanned*(d: Protocol, nodeId: NodeId): bool =
   d.banNodes and d.routingTable.isBanned(nodeId)
 
-proc receive*(d: Protocol, a: Address, packet: openArray[byte]) =
+proc receive*(d: Protocol, a: Address, packet: openArray[byte]) : DiscResult[void] =
   discv5_network_bytes.inc(packet.len.int64, labelValues = [$Direction.In])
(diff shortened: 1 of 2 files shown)


kdeme (commented):
> 
> Question: Should I add open, close functions for shared port ? @kdeme

Yes, that would be required. If you look at `closeWait` implementation, it will actually assert if the transport is already closed.

kdeme:
So the issue I have with this version is that you can still use the open/close functions no matter which new call you used. The API relies heavily  on the naming for the users to use it correctly. I prefer not do rely on naming.

gerceboss:
@kdeme can you please suggest some solution ? Either we can alias it with a different typename or something else.

gerceboss:
@kdeme I tried updating the submodules at nim-eth1 on my branch , I don't know why it hangs at cloning eth-tests submodule updates . I updated gitmodules nim-eth url and branch in my branch and was trying to update it with the new `.gitsubmodules` file

jangko:
Why not use inheritance?

```Nim
type
     DiscoveryV5Base = ref object of RootRef
          ...
          common fields
     
    DiscoveryV5Ref = ref object of DiscoveryV5Base
         fields
             
    DiscoveryV5WithTransportRef = ref object of DiscoveryV5Base
        fields

# all shared procs will use `DiscoveryV5Base` as their first param

# Specific to DiscoveryV5Ref
proc open(p: DiscoveryV5Ref) = 
    ....
    
```

gerceboss:
I think the flag might be a better option @kdeme @jangko , provides customisation as well and helps maintain code

PR #822, opened 2025-10-17
EIP 7919: Pureth Meta

RazorClient:
This Pr introduces all the types required + the helper functions for the same to implement 

[EIP 7919- Pureth Meta](https://eips.ethereum.org/EIPS/eip-7919)

The files inside eth/ssz have the type definitions and the recast functions to be per the 
- https://eips.ethereum.org/EIPS/eip-7807
- https://eips.ethereum.org/EIPS/eip-6465
- https://eips.ethereum.org/EIPS/eip-6466
- https://eips.ethereum.org/EIPS/eip-6404 

The open question was 
- How to introduce the new fields in the receipts as per

Field | Description
-- | --
from | The transaction sender’s address.
gas_used | How much gas this individual transaction used. Replaces cumulative_gas_used.
contract_address | For transactions deploying a contract, the new contract address.
logs | Logs emitted during transaction execution.
status | EIP-658 transaction status code.
authorities | For transactions with an authorization list, the list of EIP-7702 authority addresses. Non-successful authorizations are represented with an all-zero address.

The answer that came up was to change stored receipts to make room for the new fields and pass them on to the ```makereceipts``` and have them produce the new stored receipt variant with eip7807  and then recast them into seq[ssz_receipts] and then calc the root 

So we need to save the new variant in the db too, as the stored receipt is what is present in the db and do tests to make sure its correct


eth/common/receipts_rlp.nim
```diff
@@ -15,7 +15,7 @@ export addresses_rlp, base_rlp, hashes_rlp, receipts, rlp
 
 # RLP encoding for Receipt (eth/68)
 proc append*(w: var RlpWriter, rec: Receipt) =
-  if rec.receiptType in {Eip2930Receipt, Eip1559Receipt, Eip4844Receipt, Eip7702Receipt}:
+  if rec.receiptType in {Eip2930Receipt, Eip1559Receipt, Eip4844Receipt, Eip7702Receipt, Eip7807Receipt}:
```
advaita-saha: since maybe this EIP will alter receipts heavily
Maybe we can cleanup `Receipt` and `StoredReceipt`
Still debateable, thoughts @tersec

tersec:
CI won't run here until the `eth/common/headers.nim` merge conflict is resolved.

PR #835, opened 2025-12-09
feat: add QUIC port

richard-ramos:
Adds an optional quic port to ENRs

### DIFF
--- a/eth/enr/enr.nim
+++ b/eth/enr/enr.nim
@@ -75,6 +75,8 @@ type
     udp*: Opt[int]
     tcp6*: Opt[int]
     udp6*: Opt[int]
+    quic*: Opt[int]
+    quic6*: Opt[int]
 
   EnrResult*[T] = Result[T, cstring]
 
@@ -218,7 +220,7 @@ macro initRecord*(
 func insertAddress(
     fields: var seq[FieldPair],
     ip: Opt[IpAddress],
-    tcpPort, udpPort: Opt[Port]) =
+    tcpPort, udpPort, quicPort: Opt[Port]) =
   ## Insert address data.
   ## Incomplete address information is allowed (example: Port but not IP) as
   ## that information might be already in the ENR or added later.
@@ -233,6 +235,8 @@ func insertAddress(
     fields.insert(("tcp", tcpPort.get().uint16.toField))
   if udpPort.isSome():
     fields.insert(("udp", udpPort.get().uint16.toField))
+  if quicPort.isSome():
+    fields.insert(("quic", quicPort.get().uint16.toField))
 
 func init*(
     T: type Record,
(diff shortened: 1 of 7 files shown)


eth/enr/enr.nim
```diff
@@ -326,16 +331,17 @@ func update*(
     ip: Opt[IpAddress] = Opt.none(IpAddress),
     tcpPort: Opt[Port] = Opt.none(Port),
     udpPort: Opt[Port] = Opt.none(Port),
+    quicPort: Opt[Port] = Opt.none(Port),
```
kdeme: idem ^

arnetheduck:
> What is that status of the quic usage / addition of CL clients?

we're among the last to implement it

arnetheduck:
https://github.com/ethereum/consensus-specs/pull/3644
https://github.com/ethereum/consensus-specs/pull/3874

It's out there and used by the majority of clients, so it's maybe a crawler issue?

kdeme:
> 
> It's out there and used by the majority of clients, so it's maybe a crawler issue?

Could be, I added it rather quickly and test ran it. I'll verify later.

eth/p2p/discoveryv5/protocol.nim
```diff
@@ -1059,7 +1059,7 @@ func init*(
 proc newProtocol*(
     privKey: PrivateKey,
     enrIp: Opt[IpAddress],
-    enrTcpPort, enrUdpPort: Opt[Port],
+    enrTcpPort, enrUdpPort, enrQuicPort: Opt[Port],
```
kdeme: We could also do without altering this one (afaik it is not needed for nimbus-eth2 and I'd like to deprecate it).

eth/p2p/discoveryv5/protocol.nim
```diff
@@ -1124,6 +1124,26 @@ proc newProtocol*(
...
+  newProtocol(
+    privKey, enrIp, enrTcpPort, enrUdpPort, Opt.none(Port), localEnrFields,
+    bootstrapRecords, previousRecord, bindPort, bindIp, enrAutoUpdate,
+    banNodes, config, rng
+  )
+
+proc newProtocol*(
+    privKey: PrivateKey,
+    enrIp: Opt[IpAddress],
+    enrTcpPort, enrUdpPort, enrQuicPort: Opt[Port],
```
kdeme: Idem for this one

eth/p2p/discoveryv5/protocol.nim
```diff
@@ -1124,6 +1124,26 @@ proc newProtocol*(
...
+    config = defaultDiscoveryConfig,
+    rng = newRng()):
+    Protocol =
+  newProtocol(
+    privKey, enrIp, enrTcpPort, enrUdpPort, Opt.none(Port), localEnrFields,
+    bootstrapRecords, previousRecord, bindPort, bindIp, enrAutoUpdate,
+    banNodes, config, rng
+  )
+
+proc newProtocol*(
```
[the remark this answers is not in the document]
kdeme: The difference is the `bindIp` parameter.

One accepts an `IpAddress`, the other (newer) one an `Opt[IpAddress]`. 
Functionally mostly the same, but with the important (and rather hidden) difference that passing a `None` there has the effect that this also gets passed down to the `newDatagramTransport` which underneath looks up for the appropriate "AnyAddress" (`::` when dual stack, `0.0.0.0` when IPv4 only).

I want to deprecate the original one, and keep just the newer one with proper comment explaining this.

PR #836, opened 2025-12-10
Update eth/common/times to support compiling to bare metal

bhartnett:


kdeme (approved):
I think a cleaner way would be to pull this `now` outside of eth/common/times.nim and have some `EthTime.from....` defined in nimbus-eth1. 

Might still have to do the `when not defined` trick there though as `std/times` would probably still leak.

PR #842, opened 2025-12-18
Remove block access list from block body

bhartnett:
EIP-7928 has been updated and the block body will no longer include the block access list. See here: https://github.com/ethereum/EIPs/pull/10937

### DIFF
--- a/eth/enr/enr.nim
+++ b/eth/enr/enr.nim
@@ -76,6 +76,8 @@ type
     udp*: Opt[int]
     tcp6*: Opt[int]
     udp6*: Opt[int]
+    quic*: Opt[int]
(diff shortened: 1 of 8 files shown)


kdeme (commented):
Do I understand correctly that this is purely an EL split up of block body and BAL?

As in, on CL side it remains in the blocks (ExecutionPayload) and gets transferred as a whole?

Is the split up on EL side then rather about storage and the flexibility in storage it gives? Like to be able to store only recent BALs but older blocks?

This does have the effect that when syncing on EL side, one would not be able to default use any performance improvement (parallelization) that BALs provide, right? (Unless getting data from the CL, like we do for the last part. Or from CL era files.)

PR #850, opened 2026-04-10
Add RlpArrayBufWriter - uses ArrayBuf to avoid heap allocations

bhartnett:
This new rlp writer significantly improves the performance of the stateroot computation in the Nimbus EL. 

It's much faster to pre-allocated the buffer on the stack using the ArrayBuf type.


bhartnett:
CI failures are unrelated to this change.

PR #851, opened 2026-04-14
remove the `init` block

narimiran:
After
https://github.com/nim-lang/Nim/commit/2501e23d8170c56b61d126a9e349ae6d8e4b6267, the Nim compiler is stricter about the scopes.
This fixes the failing test on both Nim devel and 2.2.x.


kdeme (commented):
This does change behaviour, specifically for llvm fuzzers: https://github.com/status-im/nim-testutils/blob/master/testutils/fuzzing.nim#L56

But I guess this way of doing that just wont work anymore then.

PR #865, opened 2026-07-19
fix(db): bound SQLite prefix searches

catwith1hat:
SqKeyspaceRef.find implements prefix matching with a lexicographic range:

    key >= prefix AND key < nextPrefix(prefix)

nextPrefix must therefore produce the smallest byte string greater than every key beginning with prefix. This is done by scanning from the end, incrementing the rightmost byte below 0xff, and dropping any trailing 0xff bytes. For example:

    [0]        -> [1]
    [0, 7]     -> [0, 8]
    [0, 255]   -> [1]
    [255, 255] -> no upper bound

The old implementation looked for a byte equal to 0xff instead of the first byte not equal to 0xff. A normal prefix such as [0] consequently returned no upper bound, so find fell back to `key >= prefix` and included every later key. It also returned no bound for [0, 255], even though [1] is the correct bound.

Additionally, the old code used its backward scan counter as a forward slice position. For example, [0, 255, 1] produced [1] rather than [0, 255, 2], making the query range include keys that did not share the requested prefix. Only a prefix consisting entirely of 0xff bytes should lack an upper bound.

nextPrefix is private to kvstore_sqlite3 and is called directly only by the SQLite find implementation. The bug is nevertheless visible to external library users through the exported SqKeyspaceRef.find and KvStoreRef.find APIs when the latter wraps a SQLite keyspace. There are no production callers in this repository; the in-tree callers are database tests. MemStoreRef.find uses a separate implementation and is unaffected.

Add regressions for ordinary and trailing-0xff prefixes, with later nonmatching keys present so an accidentally unbounded or overly broad query is detected.


catwith1hat:
This was caught by Codex & Claude and I believe the analysis to be correct.

catwith1hat:
Ping?

arnetheduck:
thanks for the patience, this one fell between the cracks :tada:

Issue #706, opened 2024-06-20
Rework ENR code to solve several shortcomings 


kdeme:
There are several issues in the current ENR code, to name some:
- Currently it is possible to have FieldPairs with duplicate keys or even full duplicate FieldPairs in the ENR due to the custom fields in the API
- Currently we cannot remove FIeldPairs in some update ENR function (see also https://github.com/status-im/nim-eth/issues/639)
- Currently when using the API with Opt IpAddres or Port, providing an Opt.none will not actually remove the field
- Some API that is more restrictive than the other: `Record.init` versus `initRecord`. The latter is not even used anywhere except for testing.

Additionally, there is currently no easy API to set the `ip6`, `tcp6` and `udp6` reserved fields. It is only possible through the custom `extraFields`, which doesn't limit these reserved keys to the correct value type.

kdeme:
Part 1: https://github.com/status-im/nim-eth/pull/707

More parts to do:
- Rework also the decoding of an ENR (need to move id key check and possibly others?) and then make toTypedRecord no longer allowed to fail.
- Add dual stack support in ENR
- Cleaner kv access
- Adding an EnrBuilder is something that fits well for ENR's I think
- Removal of fields + better API for update and/or just set individual parts

PR #874, opened 2026-08-27
enr: reject records with unsorted or duplicate keys

pucedoteth:
## Problem

[EIP-778](https://github.com/ethereum/devp2p/blob/master/enr.md#rlp-encoding) requires the key:value pairs of a record to be sorted by key and to be unique. `fromBytesAux` appends them in the order they arrive and checks neither rule, so a correctly signed record that breaks either one decodes successfully.

That leaves the decoded record contradicting itself, because the two halves of the parse disagree on which occurrence of a repeated key wins:

- the decode loop reads `id` and `secp256k1` as it goes, so it keeps the **last** occurrence — and that is what selects the identity scheme, gets signature-verified, and becomes `record.publicKey`, and so the node id;
- `getField` scans from the front and returns the **first** — and that is what `get`, `tryGet`, `get(PublicKey)` and `TypedRecord.fromRecord` all use.

A record carrying `secp256k1` twice is therefore signature-verified against, and takes its node id from, one key while `get(PublicKey)` hands out a different one that was never verified.

It also breaks the invariant `Record.pairs` documents for itself — "must remain sorted and without duplicate keys. Use the insert func to ensure this." Records arriving off the wire never go through `insert`, so that no longer holds, and `insert` in turn relies on it for its `lowerBound` call: a later `update` of such a record inserts at the wrong position.

## Fix

Reject both cases while decoding. go-ethereum enforces the same two rules in the equivalent loop in [`p2p/enr/enr.go`](https://github.com/ethereum/go-ethereum/blob/master/p2p/enr/enr.go), returning `errNotSorted` and `errDuplicateKey`. Without this, nim-eth accepts — and, because it re-broadcasts records by their raw bytes, propagates — records that spec-conformant clients drop.

## Compatibility

Records built here are unaffected: `Record.init` reaches `pairs` only through `insert`, which keeps them sorted and unique. The EIP-778 test vector and the discv5 packet/ENR test vectors all still decode.

## Tests

Two tests that fail without the change. They build their input with `initRecord`, which is deprecated precisely because it neither sorts nor deduplicates — exactly what is needed to produce a correctly signed record with a non-conformant encoding.

Locally, with the change: `tests/test_enr` 27 OK, `tests/p2p/all_tests` 98 OK, `tests/rlp/all_tests` 104 OK, `tests/test_enode` 2 OK — 0 failed.

Related to the decoding rework mentioned in #706, though that issue tracks duplicates arriving through the `extraFields` construction API rather than off the wire.

### DIFF
--- a/eth/enr/enr.nim
+++ b/eth/enr/enr.nim
@@ -492,9 +492,23 @@ func fromBytesAux(T: type Record, s: openArray[byte]): EnrResult[T] =
     pairs = newSeqOfCap[FieldPair](numPairs)
     id: string = ""
     pkRaw = Opt.none(seq[byte])
+    prevKey = ""
 
   for i in 0 ..< numPairs:
     let k = rlpResult rlp.read(string)
+    # The key:value pairs must be sorted by key and must be unique. `pairs` is
+    # documented to hold them that way and `insert` relies on it, so a record
+    # that breaks the ordering cannot be stored as it arrived. Reading on would
+    # also leave the record disagreeing with itself: the identity scheme and
+    # the public key below are taken from the last occurrence of a repeated
+    # key, while every `getField` lookup returns the first.
+    if i > 0:
+      if k == prevKey:
+        return err("Duplicate key in the ENR")
+      if k < prevKey:
+        return err("Key:value pairs in the ENR are not sorted")
+    prevKey = k
+
     case k
     of "id":
       id = rlpResult rlp.read(string)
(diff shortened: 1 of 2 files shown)


kdeme:
Thanks for the report and fix. 

This issue has already been resolved in the meanwhile in https://github.com/status-im/nim-eth/pull/875 as for some reason your PR was suddenly no longer accessible (a 404 link for me, github issue or not?).

And before that I didn't merge as I didn't like the usage of the deprecated `initRecord` in the tests, and the comments are overly verbose, but I didn't have the time to write a review comment before it became inaccesible.

